PR #49, opened 2016-01-27
Finish FromWire implementation for everything.

abhinav:
Some weirdness had to be done because structs are special cased to always be
passed by reference, which means that typedefs for structs should do the same.
And the reference vs non-reference thing along with pointers affects how
`FromWire` should work. I've verified these changes by hand but we'll
obviously want tests for all the weird corner cases to get better confidence
in this.

To make fromWire work consistently between types, every custom types gets a
function similar to the map/list/set `_Read` function:

    func _$typeName_Read(v wire.Value) ($typeName, error) {
        // ...
    }

This gives us a consistent return type to work with in different `FromWire`
cases.

----

Also note that this commit is hitting a major milestone finally: the generated
code actually compiles (once you fix the package name). I was able to replace
the hand-written prototype code for the following two structs with generated
code. The benchmark we had written against the hand-written version was easily
converted to use the generated code and ran with similar performance
characteristics as the hand-written version.

    struct NestedMixed {
      1: required list<set<i32>> int_set_list
      2: required map<i32,set<string>> map_int_strset
      3: required list<map<i32,set<string>>> map_int_strset_list
    }

    struct A {
        1: required string name
        2: required i64 birthDay
        3: required string phone
        4: required i32 siblings
        5: required bool spouse
        6: required double money
    }

Generated code can be seen here: https://gist.github.com/abhinav/3b6230165cc1ccf34966


gen/container.go
```diff
@@ -95,21 +95,23 @@ func (g *Generator) listReader(spec *compile.ListSpec) (string, error) {
...
+					return nil, nil
 				}
 
 				<$o> := make(<$listType>, 0, <$l>.Size)
-				<$l>.Items.ForEach(func(<$x> <$wire>.Value) error {
-					var <$i> <typeReference .Spec.ValueSpec Required>
-					<fromWire .Spec.ValueSpec $i $x>
-					// TODO error handling
+				err := <$l>.Items.ForEach(func(<$x> <$wire>.Value) error {
+					var err error
```
abhinav: Yeah, that's not needed. It's a leftover of the code I had there before.

gen/generator.go
```diff
@@ -206,6 +209,15 @@ func (g *Generator) recordGenDeclNames(d *ast.GenDecl) error {
...
 // typeCode(TypeSpec): Gets the wire.Type for the given TypeSpec, importing
 // the wire module if necessary.
+//
+// isReferenceType(TypeSpec): Returns true if the given TypeSpec is for a
+// reference type.
+//
+// toWire(TypeSpec, v): Returns an expression of type Value that contains the
+// wire representation of the item "v" of type TypeSpec.
+//
+// fromWire(TypeSpec, v): Returns an expression of type (T, error) where T is
```
abhinav: :+1:

gen/typedef.go
```diff
@@ -27,23 +27,46 @@ func (g *Generator) typedef(spec *compile.TypedefSpec) error {
...
-			// TODO: Implement some sort of fromValue template function that
-			// will tell us what to put here.
-			return nil
+		<if isStructType .Spec>
+			func (<$v> <$typedefType>) FromWire(<$w> <$wire>.Value) error {
+				return (<typeReference .Spec.Target Required>)(<$v>).FromWire(<$w>)
+			}
+		<else>
+			func (<$v> *<$typedefType>) FromWire(<$w> <$wire>.Value) error {
+				var err error
```
abhinav: :+1:

gen/struct.go
```diff
@@ -61,27 +74,45 @@ func (g *Generator) structure(spec *compile.StructSpec) error {
...
 				case <.ID>:
 					if <$f>.Value.Type() == <typeCode .Type> {
-						<$t := printf "%s.%s" $v (goCase .Name)>
-						<fromWire .Type $t (printf "%s.Value" $f)>
-						// TODO read errors
+						<$valueErr := fromWire .Type (printf "%s.Value" $f)>
+						<if or .Required (or (isReferenceType .Type) (isStructType .Type))>
+							<$v>.<goCase .Name>, err = <$valueErr>
+						<else>
+							*<$v>.<goCase .Name>, err = <$valueErr>
```
abhinav: Derp. Will fix.

prashantv:
merge this and we can finish off generation on mater.

PR #53, opened 2016-02-10
Set up coveralls

abhinav:


abhinav:
@blampe @breerly

breerly:
:sheep:

Makefile
```diff
@@ -24,4 +23,25 @@ install:
 
 .PHONY: test
 test:
-	go test $(PACKAGES) $(TEST_ARGS)
+	go test $(PACKAGES) -v
+
+.PHONY: cover
+cover:
```
prashantv: do we need cover if we don't use it to generate the coverage files?

Makefile
```diff
@@ -24,4 +23,25 @@ install:
 
 .PHONY: test
 test:
-	go test $(PACKAGES) $(TEST_ARGS)
+	go test $(PACKAGES) -v
```
prashantv: I think `-race` got dropped here. does `goveralls` run with `-race`?

abhinav: goveralls doesn't support `-race`

Makefile
abhinav: It still mentions the percentages on the command line.

PR #54, opened 2016-02-17
Enum: Handle duplicates and unexpected values

abhinav:
Add tests for enum code generation.

This PR is against the from-wire branch. I plan on making a few more PRs against that branch rather than work directly on it because fixing some of the issues from it is a bit more involved.


abhinav:
@prashantv @breerly

breerly:
:sheep: :sheep: :sheep:

gen/enum.go
```diff
@@ -68,5 +56,43 @@ func (g *Generator) enum(spec *compile.EnumSpec) error {
...
+		`
+			<$fmt := import "fmt">
+			<$wire := import "github.com/thriftrw/thriftrw-go/wire">
+
+			<$enumName := defName .Spec>
+			<$v := newVar "v">
+			<$w := newVar "w">
+
+			func (<$v> *<$enumName>) FromWire(<$w> <$wire>.Value) error {
+				switch <$w>.GetI32() {
```
prashantv: Can't we just do
```
(<$enumName>)(value)
```

That will handle unknown enums + doesn't care if there's 2 names for the same enum (both will work when the user tries to compare it etc)

gen/enum_test.go
```diff
@@ -0,0 +1,117 @@
...
+		assert.Equal(t, tt.v, tt.e.ToWire())
+	}
+}
+
+func TestEnumWithDuplicateValuesFromWire(t *testing.T) {
+	var e testdata.EnumWithDuplicateValues
+	if assert.NoError(t, e.FromWire(wire.NewValueI32(0))) {
+		assert.True(
+			t,
+			e == testdata.EnumWithDuplicateValuesP || e == testdata.EnumWithDuplicateValuesR,
```
prashantv: once the change i suggested earlier is made, you'll want to replace this with 2 `assert.Equal` since the value should be equal to both.

gen/golden_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func hash(t *testing.T, name string) string {
+	f, err := os.Open(name)
+	defer f.Close()
```
prashantv: put `defer` after the `require.NoError` (otherwise it'll panic if the file isn't found)

gen/golden_test.go
```diff
@@ -0,0 +1,67 @@
...
+	require.NoError(t, err)
+
+	h := sha1.New()
+	_, err = io.Copy(h, f)
+	require.NoError(t, err)
+
+	return fmt.Sprintf("% x", h.Sum(nil))
```
prashantv: intentional space between the hex values? (i'm fine either way).

gen/golden_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+func TestCodeIsUpToDate(t *testing.T) {
+	// This test just verifies that the generated code in testdata/ is up to
+	// date. If this test failed, run 'make generate' in the testdata/ directory
+	// and commit the changes.
+
+	files, err := filepath.Glob("testdata/*.thrift")
+	require.NoError(t, err)
+
+	outputDir := filepath.Join(os.TempDir(), "testdata")
```
prashantv: prefer if we used `ioutil.TempDir` here, makes the test resilient to having multiple copies run at the same time
```go
outputDir := ioutil.TempDir("", "testdata")
```

gen/golden_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	for _, file := range files {
+		currentPath := file + ".go"
+		currentHash := hash(t, currentPath)
+
+		newPath := filepath.Join(outputDir, filepath.Base(file)+".go")
+		require.NoError(t, run("../thriftrw-go", "-o", newPath, file))
+
+		if hash(t, newPath) != currentHash {
+			_ = run("diff", "-u", currentPath, newPath)
```
prashantv: can drop the `_ =`, unless you wanted to be explicit that you were ignoring the error

gen/testdata/Makefile
```diff
@@ -0,0 +1,18 @@
+ROOT = ../..
+THRIFTRW = $(ROOT)/thriftrw-go
+GO_FILES = $(patsubst %, %.go, $(wildcard *.thrift))
+# We put the generated code in _test files so that they don't get built with the
```
prashantv: think comment is stale

gen/testdata/Makefile
```diff
@@ -0,0 +1,18 @@
...
+# library.
+
+.PHONY: generate
+generate: $(GO_FILES)
+
+.PHONY: clean
+clean:
+	make -C $(ROOT) clean
```
prashantv: should clean also delete `$(GO_FILES)`?

gen/enum.go
abhinav: Yeah, that handles unknown better. Done

gen/enum_test.go
abhinav: true

gen/golden_test.go
abhinav: :+1:

gen/golden_test.go
abhinav: :+1:

gen/golden_test.go
abhinav: :+1:

gen/testdata/Makefile
abhinav: deleted

gen/testdata/Makefile
abhinav: I would rather not have `clean` delete anything that is supposed to be checked into the repo.

abhinav:
Updated and squashed commits. Ignore Travis failure. Coveralls is down.

prashantv:
lgtm

PR #55, opened 2016-02-17
containers: Fix "use of untyped nil"

abhinav:
When parsing containers with primitives, the `fromWire` expression is,
`value.Get$Type(), nil` (where `$Type` is the primitive type). While reading
an item from Value, this becomes,

    foo, err := value.GetI32(), nil

Assigning nil to a variable with no type information will result in "use of
untyped nil". To fix this, we need to add, `var err error` on the line before
it.

    var err error
    foo, err := value.GetI32(), nil

This changes the code generator for lists, sets, and maps to do this.

Also added tests for containers containing primitive types.


abhinav:
@breerly @akshayjshah @prashantv

gen/container_test.go
```diff
@@ -0,0 +1,211 @@
...
+						{Key: wire.NewValueString("foo"), Value: wire.NewValueBool(true)},
+						{Key: wire.NewValueString("bar"), Value: wire.NewValueBool(false)},
+						{Key: wire.NewValueString("baz"), Value: wire.NewValueBool(true)},
+					}),
+				}),
+			}}}),
+		},
+	}
+
+	for _, tt := range tests {
```
breerly: that test table though :)

breerly:
:sheep: :it:

prashantv:
lgtm

PR #56, opened 2016-02-18
Minor refactor of code generation logic

abhinav:
It was really annoying me that there were all these methods on the Generator types spread across N files. I swapped that out in favor of a Generator interface and moved most of those methods into functions on the interface.

The gen/container.go change is noisy because it also splits maps, sets, and lists into their own files. No logic was changed besides adding a map, set, and list generator type which maintains the state for the container types that was previously in the Generator struct.

I also added a TemplateOption thing which is not being used right now but I plan to use it to reduce the number of globally available template functions in favor of explicitly requested groups.


abhinav:
@prashantv

prashantv:
lgtm

PR #57, opened 2016-02-19
structs: fix optional primitives

abhinav:
Previously the generated code for optional primitive fields was:

    *s.SomeField, err := ...

Which is invalid because SomeField is probably `nil` and dereferencing that is
invalid. The new logic code for this is,

    x, err := ...
    s.SomeField = &x


abhinav:
@prashantv @breerly @akshayjshah

gen/struct_test.go
```diff
@@ -0,0 +1,344 @@
...
+				{ID: 4, Value: wire.NewValueI32(3)},
+				{ID: 5, Value: wire.NewValueI64(4)},
+				{ID: 6, Value: wire.NewValueDouble(5.0)},
+				{ID: 7, Value: wire.NewValueString("foo")},
+				{ID: 8, Value: wire.NewValueBinary([]byte("bar"))},
+			}}),
+		},
+		{
+			&testdata.PrimitiveOptionalStruct{BoolField: boolp(true)},
+			wire.NewValueStruct(wire.Struct{Fields: []wire.Field{
```
prashantv: optional: you might want a helper to reduce the line noise in the tests
```go
singleFieldStruct := func(id int, value wire.Value) wire.Value {
  return wire.NewValueStruct(wire.Struct{Fields: []wire.Field{
    {ID: id, Value: value},
  }})
}
```

gen/struct_test.go
```diff
@@ -0,0 +1,344 @@
...
+						}),
+					}),
+				},
+				{
+					ID: 3,
+					Value: wire.NewValueMap(wire.Map{
+						KeyType:   wire.TI64,
+						ValueType: wire.TDouble,
+						Size:      2,
+						Items: wire.MapItemListFromSlice([]wire.MapItem{
```
prashantv: does serialization of the map lead to a consistent ordering of map items? Or does it use range to iterate over the map, which could lead to different orders in the resulting wire value.

abhinav: Uses range. Not consistent. `ValuesAreEqual` makes unordered comparison.

gen/list.go
```diff
@@ -123,7 +123,6 @@ func (l listGenerator) Reader(g Generator, spec *compile.ListSpec) (string, erro
 
 				<$o> := make(<$listType>, 0, <$l>.Size)
 				err := <$l>.Items.ForEach(func(<$x> <$wire>.Value) error {
-					var err error
```
prashantv: Shouldn't this change have caused `structs.thrift.go` golden test to fail?

prashantv: nvm, diff was suppressed, the file did change as expected.

prashantv:
lgtm

gen/struct_test.go
abhinav: :+1:

PR #63, opened 2016-03-18
Fix coverage

abhinav:
Until goveralls is fixed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: go
 
 go:
     - 1.5
+    - 1.6
 
 env:
     global:
--- a/scripts/cover.sh
+++ b/scripts/cover.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+set -e
+
+COVER=cover
+ROOT_PKG=github.com/thriftrw/thriftrw-go/
+
+if [[ -d "$COVER" ]]; then
+	rm -rf "$COVER"
+fi
+mkdir -p "$COVER"
+
+i=0
+for pkg in "$@"; do
+	i=$((i + 1))
+
+	coverpkg=$(go list -json "$pkg" | jq -r '
+		.Deps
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -3,6 +3,7 @@ language: go
 
 go:
     - 1.5
+    - 1.6
```
breerly: yay :)

breerly:
wow, surprised that this is needed. :shipit: 

I guess we'll be needing this for yarpc-go as well.

abhinav:
Oh, forgot to add the package being tested to `-coverpkg`.

scripts/cover.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -e
+mkdir -p cover
```
prashantv: you probably want to delete all files in the `cover` directory if it already exists

scripts/cover.sh
```diff
@@ -0,0 +1,22 @@
...
+PACKAGES=$(go list $(glide nv))
+QUERY='.Deps | .[] | select(startswith("github.com/thriftrw/thriftrw-go/"))'
+
+i=0
+for pkg in ${PACKAGES[@]}; do
+	i=$((i + 1))
+	dependencies=$(go list -json "$pkg" \
+		| jq -r "$QUERY" \
+		| tr '\n' ',' \
+		| sed 's/,$//')  # need to drop trailing ,
```
prashantv: you can do all this in jq too,
```
jq -r '.Deps | map(select(startswith("$PKG"))) | join(",")'
```

scripts/cover.sh
abhinav: true

scripts/cover.sh
abhinav: :+1:

abhinav:
Updated

prashantv:
lgtm

PR #65, opened 2016-03-22
Tests for union types

abhinav:
@prashantv @kriskowal

### DIFF
--- a/gen/struct_test.go
+++ b/gen/struct_test.go
@@ -400,3 +400,231 @@ func TestBasicException(t *testing.T) {
 		assert.Equal(t, "DoesNotExistException{Key: foo}", err.Error())
 	}
 }
+
+func TestUnionSimple(t *testing.T) {
+	tests := []struct {
+		s testdata.Document
+		v wire.Value
+	}{
+		{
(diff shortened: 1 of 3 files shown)


gen/struct_test.go
```diff
@@ -318,3 +318,133 @@ func TestNestedStructsOptional(t *testing.T) {
...
+			wire.NewValueStruct(wire.Struct{Fields: []wire.Field{
+				{ID: 1, Value: wire.NewValueBinary([]byte{1, 2, 3})},
+			}}),
+		},
+		{
+			testdata.Document{PlainText: stringp("hello")},
+			wire.NewValueStruct(wire.Struct{Fields: []wire.Field{
+				{ID: 2, Value: wire.NewValueString("hello")},
+			}}),
+		},
```
prashantv: what happens when you get an `ID` that's not recognized?

prashantv:
Let's add tests for:
 - no IDs received
 - single unrecognized ID
 - multiple IDs (either both recognized, or only one recognized, or both recognized)

These should probably cause errors?

abhinav:
:+1:

abhinav:
Updated

prashantv:
lgtm

PR #66, opened 2016-03-22
Generate String() method for all struct-like types

abhinav:


gen/testdata/test.thrift.go
```diff
@@ -365,6 +396,17 @@ func _Event_Read(w wire.Value) (*Event, error) {
 	err := v.FromWire(w)
 	return &v, err
 }
+func (v *Event) String() string {
+	var fs [2]string
+	i := 0
+	if v.Time != nil {
+		fs[i] = fmt.Sprintf("Time: %v", *(v.Time))
```
prashantv: nit: do you need to dereference here? since the `String` method is already on the pointer, it should work without right?

prashantv:
Required fields won't cause nil panics if they're not set right? Want to make sure `String` works even if a required field is `nil` (for debugging purposes)

gen/testdata/test.thrift.go
abhinav: `Timestamp` is a `type Timestamp int64` which doesn't have a `String()` method yet.

abhinav:
Yeah, they won't cause panics since we're not dereferencing anything. `fmt.Sprint` handles nil values just fine. Example:

    Edge{End: <nil>, Start: Point{X: 1, Y: 2}}

Where `End`, a required field was not specified. Adding tests for this just in case.

abhinav:
Added tests for `String()` output

gen/struct_test.go
```diff
@@ -316,5 +330,45 @@ func TestNestedStructsOptional(t *testing.T) {
 		if assert.NoError(t, s.FromWire(tt.v)) {
 			assert.Equal(t, tt.s, s)
 		}
+
+		assert.Equal(t, tt.o, tt.s.String())
+	}
+}
+
+func TestStringDoesNotPanicOnMissingRequiredFields(t *testing.T) {
+	// must not panic on String() even if required fields are missing.
```
prashantv: can rename test + remove comment, since if hte tests pass, it's not causing panics

prashantv:
lgtm

gen/struct_test.go
abhinav: Oh, true. Missed that.

PR #67, opened 2016-03-22
Track coverage of generated test code

abhinav:


scripts/cover.sh
```diff
@@ -14,12 +14,22 @@ i=0
 for pkg in "$@"; do
 	i=$((i + 1))
 
+	extracoverpkg=""
+	if [[ -f "$GOPATH/src/$pkg/.extra-coverpkg" ]]; then
+		extracoverpkg=$( \
+			perl -pe "s{^}{$pkg/}" < "$GOPATH/src/$pkg/.extra-coverpkg" \
```
prashantv: nit: use `sed` instead of `perl`

prashantv:
lgtm

scripts/cover.sh
prashantv:
```
> echo "testdata" | sed -e 's|^|github.com/uber/tchannel/|g'
```

PR #68, opened 2016-03-22
exceptions: Generate Error() method

abhinav:
Depends on #66.

### DIFF
--- a/gen/struct.go
+++ b/gen/struct.go
@@ -127,6 +127,12 @@ func structure(g Generator, spec *compile.StructSpec) error {
 				<$strings>.Join(<$fields>[:<$i>], ", "),
 			)
 		}
+
+		<if .Spec.IsExceptionType>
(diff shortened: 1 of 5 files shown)


gen/struct.go
```diff
@@ -127,6 +127,17 @@ func structure(g Generator, spec *compile.StructSpec) error {
 				<$strings>.Join(<$fields>[:<$i>], ", "),
 			)
 		}
+
+		<if .Spec.IsExceptionType>
+			func (<$v> <$structRef>) Error() string {
+				return <$v>.String()
+			}
+
+			func Is<$structName>(err error) bool {
```
prashantv: don't need this  since it's just a cast. what we wanted was `Is<Method>Error`

abhinav: :+1:

abhinav:
#66 is merged. Rebased on master.

prashantv:
lgtm

PR #69, opened 2016-03-22
Retry submission to coveralls.

abhinav:
Something causes coveralls to randomly fail with

    Bad response status from coveralls: 422 - {"message":"Couldn't find a repository matching this job.","error":true}

`travis_retry` will retry the command three times if it exits with non-zero
exit codes.


prashantv:
lgtm

PR #80, opened 2016-03-23
Code generation cleanup

abhinav:
This includes a bunch of refactoring for code generation logic that makes it easier to read. Especially the complex if expressions -- I got rid of them.

The following are the two actual code generation changes:

1.  In struct FromWire, for optional primitive fields, we'll do:

        x, err := f.Value.GetBool(), error(nil)
        v.BoolValue = &x

    We'll do,

        var x bool
        x, err = f.Value.GetBool(), error(nil)
        v.BoolValue = &x

    That is, the err variable will be picked up from the surrounding scope rather
    than being declared anew. This makes the contract for this snippet similar to
    that for the other `FromWire` logic:

        v.Contact, err = _ContactInfo_Read(f.Value)
        v.EmailAddress, err = f.Value.GetString(), error(nil)

2.  We won't include optional fields that are unset in the `String()` output.

### DIFF
--- a/gen/constant.go
+++ b/gen/constant.go
@@ -26,7 +26,7 @@ import "github.com/thriftrw/thriftrw-go/compile"
 func Constant(g Generator, c *compile.Constant) error {
 	err := g.DeclareFromTemplate(
 		`
-		const <goCase .Name> <typeReference .Type Required> = nil // TODO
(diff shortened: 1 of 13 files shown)


abhinav:
Ugh, `%v` formatting on map via `fmt` is non-deterministic. Going to drop that test.

prashantv:
lgtm

PR #82, opened 2016-03-25
compile/TypeSpec: Track original file names

abhinav:
This changes TypeSpec to track and expose the name of each file in which a
type was defined.

We need this to be able to import the corresponding generated modules correctly.

### DIFF
--- a/compile/enum.go
+++ b/compile/enum.go
@@ -28,6 +28,7 @@ import (
 // EnumSpec represents an enum defined in the Thrift file.
 type EnumSpec struct {
 	Name  string
+	File  string
 	Items []EnumItem
 }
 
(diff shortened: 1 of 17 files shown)


abhinav:
@prashantv @akshayjshah

compile/enum.go
```diff
@@ -88,6 +89,11 @@ func (e *EnumSpec) ThriftName() string {
 	return e.Name
 }
 
+// ThriftFile for EnumSpec
+func (e *EnumSpec) ThriftFile() string {
+	return e.File
+}
+
```
akshayjshah: Is this required to satisfy some interface?

akshayjshah: Yes, I see that it is.

Damn GitHub and their lack of staged comments.

akshayjshah:
lgtm, though I don't have much context into what the surrounding code is doing.

PR #83, opened 2016-03-25
Make one package per Thrift file; support includes

### LINKED ISSUE #74: Code generation for imported Thrift files
We'll have to add imports, etc. as well

abhinav:
Generated code for each Thrift file now goes into its own package. If a type
defined in a Thrift file is imported and referenced in another Thrift file, the
generated code for that file will include an import for the original Thrift
file's generated package. If a file is imported but nothing from it is
referenced, an import is *not* added, but code for it is still generated.

The generated package structure is calculated based on the locations of the
Thrift files relative to a 'Thrift root' directory. By default, that's just
the deepest common ancestor of all Thrift files in a dependency tree but it
may be specified explicitly with the `-thrift-root` option. So, if the Thrift
root is foo/ and the Thrift files are foo/a.thrift and foo/common/b.thrift,
the generated packages are foo/a and foo/common/b.

The import paths of these packages are calculated based on the location of the
output directory relative to `$GOPATH/src`. This may be provided explicitly
with the `-pkg-prefix` option. This option MUST be provided if the output
directory is not in the `$GOPATH`.

The following other command line options were added:

- `-out DIR`: The directory into which all the generated code will be written.
  Defaults to the current directory.
- `-no-recurse`: Used to disable code generation for Thrift files `include`d
  by the main Thrift file. Note that this is the opposite of Apache Thrift's
  behavior, which defaults to not generating code for included Thrift files.

Resolves #74

Depends on #85.


abhinav:
@prashantv

gen/container_test.go
```diff
@@ -32,13 +34,13 @@ import (
 func TestCollectionsOfPrimitives(t *testing.T) {
 	tests := []struct {
 		desc string
-		p    testdata.PrimitiveContainers
+		p    containers.PrimitiveContainers
```
prashantv: is it worth trying to reduce the stutter here? Maybe by importing the testdata subpackages as like "tc" "te" "ts"

gen/enum.go
```diff
@@ -22,17 +22,49 @@ package gen
 
 import "github.com/thriftrw/thriftrw-go/compile"
 
+// enumGenerator generates code to serialize and deserialize enums.
+type enumGenerator struct {
+	hasReaders
+}
+
+func (e *enumGenerator) Reader(g Generator, spec *compile.EnumSpec) (string, error) {
```
prashantv: can you explain this change? should it be moved out to a separate PR?

gen/generate.go
```diff
@@ -21,60 +21,165 @@
...
+		files["types.go"] = buff
 	}
 
+	// TODO(abg): Services
+
+	for relPath, contents := range files {
+		fullPath := filepath.Join(packageDir, relPath)
+		directory := filepath.Dir(fullPath)
+
+		if err := os.MkdirAll(directory, 01766); err != nil {
```
prashantv: why `01766`? usually directories have `755`

gen/generate.go
```diff
@@ -21,60 +21,165 @@
...
+			return fmt.Errorf("could not create directory %q: %v", packageDir, err)
+		}
+
+		f, err := os.Create(fullPath)
```
prashantv: can simplify using `ioutil.WriteFile` since you have `[]byte` in the `bytes.Buffer` anyway

gen/generate_test.go
```diff
@@ -0,0 +1,62 @@
...
+			OutputDir:     "testdata",
+			PackagePrefix: "github.com/thriftrw/thriftrw-go/gen",
+			ThriftRoot:    thriftRoot,
+		},
+	}
+
+	for _, opt := range opts {
+		assert.Error(t, Generate(module, opt),
+			"expected code generation with %v to fail", opt)
+	}
```
prashantv: look for `must be an absolute path` in the error message

gen/generator.go
```diff
@@ -79,27 +92,59 @@ type generator struct {
...
+	if t.ThriftFile() == g.ThriftFile {
+		return goCase(t.ThriftName()), nil
+	}
+
+	pkg, err := filepath.Rel(g.thriftRoot, t.ThriftFile())
+	if err != nil {
+		return "", err
+	}
+	pkg = g.Import(filepath.Join(g.importPrefix, strings.TrimSuffix(pkg, ".thrift")))
+	// TODO(abg): Refactor. Generator shouldn't know any of this crap.
```
prashantv: +1, this won't work once you add support for `namespace go` either

gen/helpers_test.go
```diff
@@ -37,3 +46,65 @@ func int32p(x int32) *int32      { return &x }
...
+		// We only care about the path relative to the directory being
+		// hashed.
+		path, err = filepath.Rel(dir, path)
+		if err != nil {
+			return err
+		}
+		fileHashes[path] = fileHash
+		return nil
+	})
+
```
prashantv: nit: delete blank. optionally can inline the if into the `err := `

gen/helpers_test.go
```diff
@@ -37,3 +46,65 @@ func int32p(x int32) *int32      { return &x }
...
+	fileNames := make([]string, 0, len(fileHashes))
+	for name := range fileHashes {
+		fileNames = append(fileNames, name)
+	}
+	sort.Strings(fileNames)
+
+	h := sha1.New()
+	for _, name := range fileNames {
+		if _, err := h.Write([]byte(name + "\t" + fileHashes[name] + "\n")); err != nil {
```
prashantv: nit: use `io.WriteString`, that way your code will be simpler without casts,
or
```go
fmt.Fprintf(h, "%v\t%v\n", name, fileHashes[name]
```

(it returns error as well)

gen/mixins.go
```diff
@@ -0,0 +1,60 @@
...
+
+	if _, ok := r.have[name]; ok {
+		return ok
+	}
+
+	r.have[name] = struct{}{}
+	return false
+}
+
+// hasLazyLists
```
prashantv: nit: incomplete comment

gen/mixins.go
```diff
@@ -0,0 +1,60 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package gen
```
prashantv: can this be moved to a separate PR

gen/enum.go
abhinav: Good idea. Created https://github.com/thriftrw/thriftrw-go/pull/85 with just these generator changes. Will rebase this PR on top.

gen/mixins.go
abhinav: :+1: 

Moved to https://github.com/thriftrw/thriftrw-go/pull/85 and rebased on top

abhinav:
I have also split out the changes to test.thrift into a separate commit so the actual business logic changes are in a single commit

gen/container_test.go
abhinav: sounds good

gen/generate.go
abhinav: I assumed it needed an octal literal when specifying permissions. It appears that it does not. 1 was for the directory flag but I guess that's not required here either. Switched to 755.

gen/generate.go
abhinav: True

gen/generate.go
abhinav: Correction: You still need octal literals

gen/generate_test.go
abhinav: :+1:

gen/helpers_test.go
abhinav: :+1:

gen/helpers_test.go
abhinav: :+1:

gen/mixins.go
abhinav: my bad

abhinav:
Updated based on comments. Will squash some of the related commits for this PR once it's time to merge.

abhinav:
Rebased on master

abhinav:
Rebased

gen/generate.go
```diff
@@ -21,60 +21,160 @@
...
+		g := NewGenerator(o.PackagePrefix, o.ThriftRoot, m.ThriftPath)
+
+		for _, typeName := range sortStringKeys(m.Types) {
+			if err := TypeDefinition(g, m.Types[typeName]); err != nil {
+				return err
+			}
+		}
+
+		buff := new(bytes.Buffer)
+		if err := g.Write(buff, token.NewFileSet()); err != nil {
```
prashantv: (not related to this diff) do we use the `token.FileSet` or do we expect to use it from outside of the generator?

gen/generate.go
```diff
@@ -21,60 +21,160 @@
...
 	}
 
+	// TODO(abg): Services
+
+	for relPath, contents := range files {
+		fullPath := filepath.Join(packageDir, relPath)
+		directory := filepath.Dir(fullPath)
+
+		if err := os.MkdirAll(directory, 0755); err != nil {
+			return fmt.Errorf("could not create directory %q: %v", packageDir, err)
```
prashantv: you probably want to use `directory` in the error

main.go
```diff
@@ -21,54 +21,193 @@
...
+				"%q is not contained in the %q directory tree",
+				m.ThriftPath, root)
+		}
+
+		return nil
+	})
+}
+
+// findCommonAncestor finds the deepest common ancestor for the Thrift file
+// for the given module and all the modules imported by it.
```
prashantv: the method doc is a little confusing "for the Thrift file for the given module". Maybe just "for the given module and all modules imported by it" is enough

main.go
```diff
@@ -21,54 +21,193 @@
...
+		if len(result) == 0 {
+			return fmt.Errorf(
+				"%q does not have a prefix in common with %q",
+				thriftPath, lastString)
+		}
+
+		lastString = thriftPath
+		return nil
+	})
+
```
prashantv: nit: delete blank

main.go
```diff
@@ -21,54 +21,193 @@
...
+// determinePackagePrefix determines the package prefix for Go packages
+// generated in this file.
+//
+// dir must be an absolute path.
+func determinePackagePrefix(dir string) (string, error) {
+	gopath := os.Getenv("GOPATH")
+	if gopath == "" {
+		return "", errors.New("$GOPATH is not set")
+	}
+
```
prashantv: GOPATH could have multiple colon separated paths, so split on `:` and iterate over all of them and check for at least one match.

main_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+			if tt.errMsg != "" {
+				if assert.Error(t, err, "determinePackagePrefix(%q) should fail", tt.dir) {
+					assert.Contains(t, err.Error(), tt.errMsg)
+				}
+			}
+		})
+	}
+}
+
```
prashantv: can we add tests for `findCommonAncestor` since it's kind of complicated (yay magic)

prashantv:
Completed a review pass, left a few comments

gen/generate.go
abhinav: No we don't use it.

gen/generate.go
abhinav: true

main.go
abhinav: :+1:

main.go
abhinav: :+1:

main.go
abhinav: Fixed

main_test.go
abhinav: true

abhinav:
Updated

gen/generate.go
prashantv: Cool, if we don't plan to use it, maybe we should remove it from the interface in a later diff.

abhinav: :+1:

main.go
```diff
@@ -21,54 +21,209 @@
...
+		}
+
+		lastString = thriftPath
+		return nil
+	})
+	if err != nil {
+		return "", err
+	}
+
+	if result[0] == "" {
```
prashantv: would be it be simpler to just use `strings.Join`, since you use `strings.Split`?

main.go
```diff
@@ -21,54 +21,209 @@
...
+//
+// dir must be an absolute path.
+func determinePackagePrefix(dir string) (string, error) {
+	gopathList := os.Getenv("GOPATH")
+	if gopathList == "" {
+		return "", errors.New("$GOPATH is not set")
+	}
+
+	for _, gopath := range filepath.SplitList(gopathList) {
+		// entry in GOPATH might be a relative path
```
prashantv: `GOPATH` must be absolute, if you try to run go with a relative `GOPATH`, you get:

```
go: GOPATH entry is relative; must be absolute path [...]
```

main.go
abhinav: indeed

main.go
abhinav: Oh, cool. I'll remove that step.

main_test.go
```diff
@@ -0,0 +1,439 @@
...
+		{
+			dir:    genDir,
+			result: "github.com/thriftrw/thriftrw-go/gen",
+		},
+		{
+			around: func(f func()) {
+				gopath := os.Getenv("GOPATH")
+				require.NoError(t, os.Setenv("GOPATH", tmpDir+":"+gopath))
+				defer os.Setenv("GOPATH", gopath)
+			},
```
prashantv: no call to `f`

the `around` functionality seems too generic and is more error-prone to bugs like this. suggest just adding a property like,
```go
  overrideGoPath func(curGoPath string) string
```

main.go
```diff
@@ -21,54 +21,209 @@
...
+		gopath, err := filepath.Abs(gopath)
+		if err != nil {
+			return "", err
+		}
+
+		packagePath, err := filepath.Rel(filepath.Join(gopath, "src"), dir)
+		if err != nil {
```
prashantv: don't think this is right, if `Rel` fails for one GOPATH, it may still succeed on another. you shouldn't be returning at that point. 

also update tests so that it only succeeds for the second path in `GOPATH` rather than the first every time.

main_test.go
abhinav: oh derp

main_test.go
```diff
@@ -0,0 +1,437 @@
...
+		func() {
+			if tt.overrideGoPath != nil {
+				fakeGoPath := tt.overrideGoPath(realGoPath)
+				if fakeGoPath == "" {
+					require.NoError(t, os.Unsetenv("GOPATH"))
+				} else {
+					require.NoError(t, os.Setenv("GOPATH", fakeGoPath))
+				}
+
+				defer os.Setenv("GOPATH", realGoPath)
```
prashantv: this inline func adds a bit too much nesting, maybe something like:
```go
realGoPath := os.Getenv("GOPATH")
defer os.Setenv("GOPATH", realGoPath)

for _, tt := range tests {
  os.Setenv("GOPATH", realGoPath)
  if tt.overrideGoPath != nil {
    fakeGoPath := tt.overrideGoPath(realGoPath)
    if fakeGoPath == "" {
      // unset
    } else {
      // set
    }
  }

  // test
```

prashantv:
lgtm

main_test.go
abhinav: :+1:

PR #85, opened 2016-03-25
Generate Reader() functions on-demand rather

abhinav:
Instead of always generating a `_$type_Read` function, we now generate them
on-demand only.

This will allow us to generate Reader functions for imported types on demand.


prashantv:
lgtm

PR #86, opened 2016-03-25
Move bulk of struct generation logic out

abhinav:
This will allow this logic to be re-used for function args and result structs.

Depends on #85.


abhinav:
Rebased on master

prashantv:
lgtm

PR #93, opened 2016-03-30
Detect and fail on complex typedef cycles

abhinav:
After the link stage, this will walk all `typedef` reference trees and look
for any cyclic references.

The full reference cycle is printed in the error message. For example, given
the input,

    typedef x foo
    struct x { 1: required y a }
    struct y { 1: optional map<string, z> b }
    struct z { 1: optional foo c }

The error is:

    cannot compile "[..]/tmp.thrift": cannot compile "foo": found a type reference cycle:
        foo
     -> x
     -> y
     -> map<string, z>
     -> z
     -> foo

If the cyclic references are spread across multiple files, the output also
includes the file names.

    cannot compile "[..]/tmp.thrift": cannot compile "foo": found a type reference cycle:
        foo ([..]/tmp.thrift)
     -> x ([..]/tmp.thrift)
     -> y ([..]/tmp.thrift)
     -> map<string, z>
     -> z ([..]/b.thrift)
     -> foo ([..]/tmp.thrift)

### DIFF
--- a/compile/cycle_test.go
+++ b/compile/cycle_test.go
@@ -1,0 +1,152 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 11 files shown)


abhinav:
CC @prashantv @breerly

compile/cycle_test.go
```diff
@@ -0,0 +1,157 @@
...
+				}
+			}),
+			msgs: []string{
+				"found a type reference cycle",
+				"   foo",
+				"-> bar",
+				"-> foo",
+			},
+		},
+		// TODO test for cycles from structs
```
prashantv: remove TODO

compile/cycle_test.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	for _, tt := range tests {
+		typ := mustLink(t, tt.typ, scope())
+
+		err := findTypeCycles(typ)
+		if len(tt.msgs) > 0 {
```
prashantv: since you always have messages, feel free to remove the if, and assume it should always error.

abhinav: :+1:

prashantv:
lgtm

PR #94, opened 2016-03-30
Generator should not care about import paths

abhinav:
Moved import path resolution logic upwards so that the Generator isn't munging paths to figure out import paths.

A `ServicePackage` method is provided which will be used during service code generation to import the service-specific sub packages.

Depends on https://github.com/thriftrw/thriftrw-go/pull/95

### DIFF
--- a/gen/generate.go
+++ b/gen/generate.go
@@ -71,41 +71,87 @@ func Generate(m *compile.Module, o *Options) error {
 			o.OutputDir)
 	}
 
+	importer := thriftPackageImporter{
+		ImportPrefix: o.PackagePrefix,
+		ThriftRoot:   o.ThriftRoot,
+	}
+
 	if o.NoRecurse {
-		return generateModule(m, o)
+		return generateModule(m, importer, o.OutputDir)
 	}
 
 	return m.Walk(func(m *compile.Module) error {
-		if err := generateModule(m, o); err != nil {
+		if err := generateModule(m, importer, o.OutputDir); err != nil {
 			return generateError{Name: m.ThriftPath, Reason: err}
 		}
 		return nil
 	})
 }
 
+type thriftPackageImporter struct {
+	ImportPrefix string
+	ThriftRoot   string
+}
+
+// RelativePackage returns the import path for the top-level package of the
+// given Thrift file relative to the ImportPrefix.
+func (i thriftPackageImporter) RelativePackage(file string) (string, error) {
+	return filepath.Rel(i.ThriftRoot, strings.TrimSuffix(file, ".thrift"))
+}
+
+// Package returns the import path for the top-level package of the given Thrift
+// file.
+func (i thriftPackageImporter) Package(file string) (string, error) {
+	pkg, err := i.RelativePackage(file)
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(i.ImportPrefix, pkg), nil
+}
+
+// ServicePackage returns the import path for the package for the Thrift service
+// with the given name defined in the given file.
+func (i thriftPackageImporter) ServicePackage(file, name string) (string, error) {
+	topPackage, err := i.Package(file)
+	if err != nil {
+		return "", err
+	}
+
+	return filepath.Join(topPackage, strings.ToLower(name)), nil
+}
+
 // generates code for only the given module, assuming that code for included
 // modules has already been generated.
-func generateModule(m *compile.Module, o *Options) error {
-	packagePath, err := filepath.Rel(o.ThriftRoot, m.ThriftPath)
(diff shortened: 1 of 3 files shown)


abhinav:
CC @prashantv

gen/field.go
```diff
@@ -28,6 +28,25 @@ type fieldGroupGenerator struct {
 	Fields compile.FieldGroup
 }
 
+func (f fieldGroupGenerator) Generate(g Generator) error {
```
prashantv: let's move this commit into a separate PR since it seems unrelated?

gen/field.go
```diff
@@ -28,6 +28,25 @@ type fieldGroupGenerator struct {
...
+	}
+
+	if err := f.FromWire(g); err != nil {
+		return err
+	}
+
+	if err := f.String(g); err != nil {
+		return err
+	}
+	return nil
```
prashantv: nit: blank before this (to match the blank between the other groups in this method)

gen/generate.go
```diff
@@ -71,41 +71,87 @@ func Generate(m *compile.Module, o *Options) error {
...
+
+// ServicePackage returns the import path for the package for the Thrift service
+// with the given name defined in the given file.
+func (i thriftPackageImporter) ServicePackage(file, name string) (string, error) {
+	topPackage, err := i.Package(file)
+	if err != nil {
+		return "", err
+	}
+
+	return filepath.Join(topPackage, strings.ToLower(name)), nil
```
prashantv: nit: should we do a "clean" of the name (which is lowercasing, removing any dashes or underscores etc?) or is just lowercase enough

gen/generate_test.go
```diff
@@ -62,3 +62,51 @@ func TestGenerateWithRelativePaths(t *testing.T) {
...
+		},
+		{
+			File:     "/src/thrift/shared/common.thrift",
+			Relative: "shared/common",
+			Package:  "github.com/myteam/myservice/shared/common",
+		},
+	}
+
+	for _, tt := range tests {
+		if tt.Relative != "" {
```
prashantv: `Relative` seems to be always set?

gen/generate_test.go
```diff
@@ -62,3 +62,51 @@ func TestGenerateWithRelativePaths(t *testing.T) {
...
+			got, err := importer.RelativePackage(tt.File)
+			if assert.NoError(t, err, "RelativePackage(%q)", tt.File) {
+				assert.Equal(t, tt.Relative, got, "RelativePackage(%q)", tt.File)
+			}
+		}
+
+		if tt.Package != "" {
```
prashantv: same with `Package`

gen/generate_test.go
```diff
@@ -62,3 +62,51 @@ func TestGenerateWithRelativePaths(t *testing.T) {
...
+			got, err := importer.Package(tt.File)
+			if assert.NoError(t, err, "Package(%q)", tt.File) {
+				assert.Equal(t, tt.Package, got, "Package(%q)", tt.File)
+			}
+		}
+
+		if tt.ServicePackage != "" {
```
prashantv: nit: using `ServiceName` for this seems better since that's the input

prashantv:
lgtm, just move the `Generate` fieldgroup changes out, and fix minor nits.

gen/generate_test.go
abhinav: Yeah, for all existing test cases, but I'm treating it as optional. If Relative is provided, that's the expected `RelativePackage`, if `ServicePackage` is provided, that`s the expected `ServicePackage` and so on.

gen/generate_test.go
abhinav: Answered above

gen/generate_test.go
abhinav: `ServiceName` is the name passed to `ServicePackage`. This is the expected output of calling `ServicePackage(file, name)`.

gen/field.go
abhinav: Cool. I figured it was a small enough change to bundle in this PR.

gen/generate.go
abhinav: Service names have to be valid identifiers so they won't contain spaces or dashes. Underscores may be present but I think that's okay to have because we want to support `service FooBar` and `service Foo_Bar` defined at the same time. We don't need to support `service foo` and `service Foo` defined at the same time because we validate case insensitivity at compile time.

gen/field.go
abhinav: :+1:

gen/generate.go
prashantv: cool, makes sense. sgtm.

PR #95, opened 2016-03-30
fieldGenerator: add top-level Generate method

abhinav:
This just calls all other generate methods for the fieldGenerator


prashantv:
lgtm

PR #96, opened 2016-03-31
Maintain field order in generated code

abhinav:
Instead of using `map[string]*FieldSpec`, use a `[]*FieldSpec` so that we
maintain field order when generating code. This is especially required for
method parameters.


abhinav:
@prashantv @breerly

prashantv:
Do we want to maintain field order as in the file, or field order based on the field ID?

abhinav:
Maintaining the same order as the file sounds like the least surprising behavior to me.

prashantv:
ok, lgtm.

Issue #61, opened 2016-03-07
Service code generation


abhinav:
Need to generate code for services

abhinav:
Given the following Thrift file,

```thrift
// kv.thrift

exception ResourceDoesNotExist {
    1: string required key
}

service Health {
    bool healthy()
}

service KeyValue {
    void setValue(1: string key, 2: string value)
    string getValue(1: string key)
        throws (1: ResourceDoesNotExist doesNotExist)
}
```

We considered two methods for service code generation:

-   Generate all types and services from the same Thrift file in the same
    package. This means that the argument structures and response unions must
    be prepended with the service name to avoid collisions between multiple
    methods with the same name across different services. This would look
    approximately like,

    ```go
    // kv/types.go

    package kv

    type ResourceDoesNotExist struct { /* ... */ }
    ```

    ```go
    // kv/health.go

    package kv

    type HealthHealthyResult struct{Success bool}

    type Health interface {
        bool Healthy(ctx context.Context)
    }
    ```

    ```go
    // kv/keyvalue.go

    package kv

    type KeyValueSetValueArgs struct {
        Key   string
        Value string
    }

    type KeyValueGetValueResult struct {
        Success *string
        DoesNotExist *ResourceDoesNotExist
    }

    type KeyValue interface {
        SetValue(ctx context.Context, key, value string)
        GetValue(ctx context.Context, key, value string)
    }
    ```

-   Generate types in a shared package and create a separate sub-package for
    each service.

    ```go
    // kv/types.go

    // same as above
    ```

    ```go
    // kv/health/result.go

    package health

    type HealthyResult struct{Success bool}
    ```

    ```go
    // kv/keyvalue/args.go

    package keyvalue

    type GetValueArgs struct{Key string}

    type SetValueArgs struct {
        Key, Value string
    }
    ```

    ```go
    // kv/keyvalue/result.go

    package keyvalue

    import "kv"

    type GetValueResult struct {
        Success *string
        DoesNotExist *kv.ResourceDoesNotExist
    }
    ```

After some discussion, we're leaning towards the second approach because the
names of the request/response structs and unions are a lot cleaner and we don't
have to worry about naming collisions with other types defined in the Thrift
file.

We want the responsibility of generating complex service interfaces and
building/destructuring the request and response structs to be transferred to
the library-specific code generation templates. tchannel-go, yarpc-go, and
ringpop-go will have their own templates for how a service should look and call
into these generates types.

CC @prashantv

[...]

PR #97, opened 2016-04-01
Generate req/res structs for service functions

### LINKED ISSUE #36: Provide a safe way to determine if the returned err is an application error
Given the IDL:
```
  ResolveReply Resolve(1: required ResolveRequest request) throws (
    1: InvalidServiceName invalidWatchServiceName
    2: InvalidEnsembleName invalidName
  )
```

and the call:
```
  res, err := client.Watch(ctx, request)
```

I need to differentiate between and application error and a transport error. Right now I have:

```
    switch t := err.(type) {
    case *idl.InvalidServiceName:
    case *idl.InvalidEnsembleName:
    default:
      log.Infof("tchannel error")
   }
```

But that's dangerous because if someone adds a new application exception and forgets to update the code we'll treat it as a tchannel error.

Since we probably can't change the signature of the Watch() function, we'll need something like
```
func IsWatchAppException(err) bool
```

in the generated code.

abhinav:
Each service gets its own submodule, and each function gets an Args struct, a
Result union, and an anonymous-struct-based variable that provides,

    func Args(arg1, arg2, ...) *MyFunctionArgs
    func IsException(error) bool
    func WrapResponse(returnType, error) (*MyFunctionResult, error)
    func UnwrapResponse(*MyFunctionResult) (returnType, error)

For example,

    // foo.thrift

    service Foo { string bar() }

    service Bar { void baz() }

Will generate,

    foo/        # package foo for foo.thrift
     |- foo/    # package foo for service Foo
         |- bar.go
             |- type BarArgs
             |- type BarResult
             |- Bar.Args() *BarArgs
             |- Bar.IsException(error) bool
             |- Bar.WrapResponse(string, error) (*BarResult, error)
             |- Bar.UnwrapResponse(*BarResult) (string, error)
     |- bar/    # package bar for service Bar
         |- baz.go
             |- type BazArgs
             |- type BazResult
             |- Baz.Args() *BazArgs
             |- Baz.IsException(error) bool
             |- Baz.WrapResponse(error) (*BazResult, error)
             |- Baz.UnwrapResponse(*BazResult) error

Note:

- The anonymous struct for void methods uses a single parameter rather than
  two.
- We're not generating a service interface. We're leaving that up to the
  library-specfic plugins.

Resolves #36, #61,


compile/service.go
```diff
@@ -207,6 +207,11 @@ func (f *FunctionSpec) Link(scope Scope) error {
 	return nil
 }
 
+// HasReturn returns true if this function has a non-void return.
+func (f *FunctionSpec) HasReturn() bool {
+	return f.ResultSpec != nil && f.ResultSpec.ReturnType != nil
```
prashantv: (unrelated to this PR), should we always have a `ResultSpec` for a function, even if it's void, and have the `ReturnType` be `nil`? I have a similar check in yab but it might be cleaner to assume `ResultSpec != nil`

gen/generator.go
```diff
@@ -58,7 +58,7 @@ type Generator interface {
 	// module.
 	Import(path string) string
 
-	// Write the generated code to the given Writer.
+	// Write the generated code to the given Writer and end this file.
```
prashantv: what does `end this file` mean?

gen/service.go
```diff
@@ -0,0 +1,225 @@
...
+//
+// Returns a map from file name to contents for that file. The file names are
+// relative to the package directory for the service.
+func Service(g Generator, s *compile.ServiceSpec) (map[string]*bytes.Buffer, error) {
+	files := make(map[string]*bytes.Buffer)
+
+	// TODO inherited service functions
+
+	for _, functionName := range sortStringKeys(s.Functions) {
+		err := ServiceFunction(g, s, s.Functions[functionName])
```
prashantv: nit: can inline into `if`

gen/service.go
```diff
@@ -0,0 +1,225 @@
...
+// ServiceFunction generates code for the given function of the given service.
+func ServiceFunction(g Generator, s *compile.ServiceSpec, f *compile.FunctionSpec) error {
+	if f.OneWay {
+		return fmt.Errorf(
+			"oneway functions are not yet supported: %s.%s is oneway",
+			s.Name, f.Name)
+	}
+
+	argsGen := fieldGroupGenerator{
+		Name:   goCase(f.Name) + "Args",
```
prashantv: TODO clashes -- if a service has functions `Foo` and `FooArgs`, it would clash. Same for the result struct.

gen/service.go
```diff
@@ -0,0 +1,225 @@
...
+		Fields: compile.FieldGroup(f.ArgsSpec),
+	}
+	if err := argsGen.Generate(g); err != nil {
+		return wrapGenerateError(fmt.Sprintf("%s.%s", s.Name, f.Name), err)
+	}
+
+	resultFields := make(compile.FieldGroup, 0)
+	if result := f.ResultSpec; result != nil {
```
prashantv: another place where we skip skip a `nil` check for the `ResultSpec` if it was always set

gen/service.go
```diff
@@ -0,0 +1,225 @@
...
+	return nil
+}
+
+func functionHelper(g Generator, f *compile.FunctionSpec) error {
+	return g.DeclareFromTemplate(
+		`
+		<$name := goCase .Name>
+
+		var <$name> = struct{
+			IsException func(error) bool
```
prashantv: Let's add method docs to the struct fields here.

gen/service.go
```diff
@@ -0,0 +1,225 @@
...
+	// throw a generic error.
+	return nil
+}
+
+func functionHelper(g Generator, f *compile.FunctionSpec) error {
+	return g.DeclareFromTemplate(
+		`
+		<$name := goCase .Name>
+
+		var <$name> = struct{
```
prashantv: docs for this helper

gen/service.go
```diff
@@ -0,0 +1,225 @@
...
+					if err == nil {
+						return &<$name>Result{}, nil
+					}
+			<end>
+					<if .ResultSpec>
+						<if .ResultSpec.Exceptions>
+							switch e := err.(type) {
+								<range .ResultSpec.Exceptions>
+								case <typeReferencePtr .Type>:
+									return &<$name>Result{<goCase .Name>: e}, nil
```
prashantv: some people hit the `err != nil` but it's actually a `nil` error with a type,
https://github.com/uber/tchannel-go/issues/233

It might be worth checking in here
```
if e == nil {
  // do something
}
```
here's the fix in thrift-gen,
https://github.com/uber/tchannel-go/pull/240/files

gen/service.go
```diff
@@ -0,0 +1,225 @@
...
+					if err == nil {
+						return &<$name>Result{Success: success}, nil
+					}
+			<else>
+				func(err error) (*<$name>Result, error) {
+					if err == nil {
+						return &<$name>Result{}, nil
+					}
+			<end>
+					<if .ResultSpec>
```
prashantv: maybe add `HasExceptions` so you oly have one `<if>`.

gen/service.go
```diff
@@ -0,0 +1,225 @@
...
+		`
+		<$name := goCase .Name>
+
+		var <$name> = struct{
+			IsException func(error) bool
+
+			Args func(
+				<range .ArgsSpec>
+					<if .Required>
+						<.Name> <typeReference .Type>,
```
prashantv: if the `Name` clashes with a go keyword, this will fail to compile. the full list is at 
https://golang.org/ref/spec#Keywords

gen/service_test.go
```diff
@@ -0,0 +1,402 @@
...
+		},
+		{
+			desc:           "deleteValue failure (2)",
+			unwrapResponse: keyvalue.DeleteValue.UnwrapResponse,
+			resultArg: &keyvalue.DeleteValueResult{
+				InternalError: &tv.InternalError{},
+			},
+			expectedError: &tv.InternalError{},
+		},
+	}
```
prashantv: add tests for:
 - multiple errors are set
 - error and result are set
 - only result is set for void method

(these are all invalid, but i want to make sure we understand what happens in invalid cases)

gen/service_test.go
```diff
@@ -0,0 +1,402 @@
...
+				InternalError: &tv.InternalError{},
+			},
+		},
+		{
+			desc: "deleteValue failure",
+			run: func() (interface{}, error) {
+				return keyvalue.DeleteValue.WrapResponse(errors.New("foo"))
+			},
+			expectedError: errors.New("foo"),
+		},
```
prashantv: add a case where you provide a result and an error

prashantv:
Completed review pass

compile/service.go
abhinav: Sure. I wanted to avoid changing that in this change as well. Created https://github.com/thriftrw/thriftrw-go/pull/98

gen/generator.go
abhinav: Clarified in docs

gen/service.go
abhinav: :+1:

gen/service.go
abhinav: As discussed, appended `Helper` to the variable name. We can iterate on this if we decide this isn't good enough for the plugins.

gen/service.go
abhinav: True. Simplified now that #98 is merged.

gen/service.go
abhinav: No longer necessary because Resultspec is always non-nil

gen/service.go
abhinav: Punting docs as discussed. Created https://github.com/thriftrw/thriftrw-go/issues/100

gen/service.go
abhinav: https://github.com/thriftrw/thriftrw-go/issues/100

gen/service.go
abhinav: :+1:

gen/service.go
abhinav: True

gen/service_test.go
abhinav: :+1: 

Not sure what you mean by the third one, though. You can't have success set for a void method because that field doesn't exist.

gen/service_test.go
abhinav: :+1:

abhinav:
Updated. Ready for second pass.

gen/generate.go
```diff
@@ -188,7 +186,29 @@ func generateModule(m *compile.Module, i thriftPackageImporter, outDir string) e
...
+				return err
+			}
+			packageName := filepath.Base(importPath)
+
+			// TODO inherited service functions
+
+			g := NewGenerator(i, importPath, packageName)
+			serviceFiles, err := Service(g, service)
+			if err != nil {
+				return fmt.Errorf("could not code for %s: %v", serviceName, err)
```
prashantv: I think you accidentally a word in "could not code for". also mention service maybe, "could not generate code code service %s"

gen/generator.go
```diff
@@ -37,15 +37,20 @@ import (
...
 	// context.
 	TextTemplate(s string, data interface{}, opts ...TemplateOption) (string, error)
 
-	// DeclareFromTemplate renders a template (in the text/template format)
-	// that generates Go code and includes all declarations from the template
-	// in the code generated by the generator.
+	// DeclareFromTemplate renders the given template, and includes the
+	// declarations from the generated Go code in the package being generated.
+	// The next call to Write will write these declarations and any other
+	// declerations that have not been written so far.
```
prashantv: typo "declerations"

gen/service.go
```diff
@@ -0,0 +1,278 @@
...
+				s.Name, functionName, err)
+		}
+
+		buff := new(bytes.Buffer)
+		if err := g.Write(buff, token.NewFileSet()); err != nil {
+			return nil, fmt.Errorf("could not write %s.%s: %v", s.Name, functionName, err)
+		}
+
+		// TODO check for conflicts
+		files[strings.ToLower(functionName)+".go"] = buff
```
prashantv: we should probably start with prefixing the service name, if we have 2 services with the same function name in the same thrift file, it'll clash straight away.

prashantv: ignore me, since we're in separate packages, doesn't matter.

gen/generate.go
abhinav: :+1: 

you accidentally the
the word code there

gen/generator.go
abhinav: :+1:

gen/service_test.go
```diff
@@ -0,0 +1,430 @@
...
+		{
+			desc: "deleteValue application error (2)",
+			run: func() (interface{}, error) {
+				return keyvalue.DeleteValueHelper.WrapResponse(&tv.InternalError{})
+			},
+			expectedResult: &keyvalue.DeleteValueResult{
+				InternalError: &tv.InternalError{},
+			},
+		},
+		{
```
prashantv: test where you call `WrapResponse` with non-nil type but nil value

gen/service_test.go
```diff
@@ -0,0 +1,430 @@
...
+		result, err := tt.run()
+		if tt.expectedError != nil {
+			assert.Equal(t, tt.expectedError, err, tt.desc)
+		} else {
+			assert.Equal(t, tt.expectedResult, result, tt.desc)
+		}
+	}
+}
+
+func TestWrapResponseNillError(t *testing.T) {
```
prashantv: nit: typo, s/Nill/Nil. Maybe also make the name more specific, "TestWrapResponseTypedNilError" or something

prashantv:
lgtm

kriskowal:
Music to my eyes.

gen/service_test.go
abhinav: :+1:

PR #99, opened 2016-04-01
Make service function helpers top-level functions

abhinav:
This takes away the possibility of name clashes with method names.

Other options:

- Prepend/append something to the helper name. Like, `GetValueHelper`, `SetValueHelper`. (Open to better names than `Helper`.)

- Create a top-level object to hold the helper structs:

        var Helper = struct{
            GetValue struct{
                IsException func(..)
                // ...
            }
            SetValue struct{
                // ...
            }
        }

        keyvalue.Helper.GetValue.Args(...)


abhinav:
Abandoning in favor of appending "Helper" to the name.

PR #102, opened 2016-04-02
TemplateFunc: auto-pass Generator if first argument

abhinav:
TemplateFunc will now automatically pass Generator as the first argument to
functions that accept it. This greatly simplifies the code for the default
FuncMap.


prashantv:
lgtm

PR #103, opened 2016-04-02
gen/namespace: don't allow reserved names

abhinav:


gen/reserved.go
```diff
@@ -0,0 +1,42 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package gen
+
+var _reservedNames = make(map[string]struct{})
+
+func init() {
+	reservedNames := []string{
```
prashantv: put a comment about where you got the list.

prashantv: also, their list skips `error` but that should be reserved too, see
https://github.com/uber/tchannel-go/blob/dev/thrift/thrift-gen/names.go#L28

prashantv:
lgtm, make sure to add `error`

gen/reserved.go
abhinav: :+1:

PR #106, opened 2016-04-05
services: Generate subpackages in `$out/service`

abhinav:
This will allow us to generate plugin-specific code in `$out` without any
namespace clashing concerns.

@prashantv


prashantv:
lgtm

PR #108, opened 2016-04-05
Abstract filesystem access from compiler into an interface that can be overriden

prashantv:
This allows callers of `Compile` to specify an alternate filesystem interface that can read from other sources (e.g. directly from memory).

### DIFF
--- a/compile/compiler_test.go
+++ b/compile/compiler_test.go
@@ -1,0 +1,86 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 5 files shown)


compile/options.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (realFS) Read(filename string) ([]byte, error) {
+	return ioutil.ReadFile(filename)
+}
+
+func (realFS) Abs(p string) (string, error) {
+	return filepath.Abs(p)
+}
+
+// Filesystem replaces the interface used to access the filesystem.
```
abhinav: Nit: Filesystem controls how the Thrift compiler accesses the filesystem.

compile/options.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"io/ioutil"
+	"path/filepath"
+)
+
+// Option represents a compiler option.
+type Option func(*compiler)
+
+// FS is the interface used by the compiler to interact with the filesystem.
```
abhinav: Nit: FS is used by the compiler to interact with the filesystem.

compile/doc.go
```diff
@@ -0,0 +1,26 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package compile provides specifications for the types and services defined
+// in a Thrift file.
+//
+// Files are parsed using the Compile method, which will return a Module that
+// contains the types and services defined in the Thrift file.
+package compile
```
abhinav: Thanks

compile/compiler_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"strings"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/thriftrw/thriftrw-go/wire"
```
abhinav: Import group ordering

abhinav:
LGTM

compile/options.go
prashantv: :+1:

compile/compiler_test.go
prashantv: Done

compile/options.go
prashantv: Done

PR #109, opened 2016-04-05
Generate code for YARPC's Thrift encoding

abhinav:
This gives us some basic working integration with yarpc-go with a `-yarpc` flag. The YARPC servers and clients for each Thrift service get generated under the `${serviceName}server` and `${serviceName}client` packages under the `yarpc` subtree.

Most of the code generation logic will go into the yarpc-go repo once we figure out the plugin API.

Depends on https://github.com/thriftrw/thriftrw-go/pull/110


abhinav:
CC @prashantv @breerly

breerly:
Seems wrong that ThriftRW needs to know about YARPC at all - what happened to IoC?

abhinav:
This is not the final product. As per the description:

> Most of the code generation logic will go into the yarpc-go repo once we figure out the plugin API.

The intention is for the plugin to be part of yarpc-go. thriftrw-go will not be aware of yarpc-go. Right now is not the correct time to commit to a plugin API. I want to get something working first.

gen/namespace.go
```diff
@@ -76,9 +91,27 @@ func (n *namespace) NewName(base string) string {
 		name = fmt.Sprintf("%s%d", base, i)
 	}
 	n.taken[name] = struct{}{}
+
+	if _, ok := n.gave[base]; !ok {
```
prashantv: since maps return default values (`nil` for slice) and `nil` is a valid slice, you can just do
```go
n.gave[base] = append(n.gave[base], name)
```
no need for the check here.

gen/service.go
```diff
@@ -117,7 +112,26 @@ func functionHelper(g Generator, f *compile.FunctionSpec) error {
 						<unreserve .Name> <typeReferencePtr .Type>,
 					<end>
 				<end>
-			) *<$name>Args
+        `,
+		f,
+		TemplateFunc("unreserve", func(n string) string {
```
prashantv: maybe rename `unreserve` to `goArgName` or something, the template just wants a valid go name

gen/yarpc.go
```diff
@@ -0,0 +1,329 @@
...
+				<$args := $params.NewName "args">
+				<$args> := <$servicePackage>.<goCase .Name>Helper.Args(
+					<range .ArgsSpec><$params.Rotate .Name>,<end>)
+
+				<$body := $params.NewName "body">
+				<$res := $params.NewName "res">
+				<$body>, <$res>, err := c.c.Call("<.Name>", <$req>, <$args>.ToWire())
+				if err != nil {
+					<if .ResultSpec.ReturnType>
+						return <zeroValue .ResultSpec.ReturnType>, <$res>, err
```
prashantv: instead of having a zeroValue, there's approaches to do it with less code (and probably less likely to have bugs):

- use named return value, and return that here
- create a temporary var,
```go
var emptyRet <.ResultSpec.ReturnType>
return emptyRet, <$res>, err
```

What do you think?

prashantv:
lgtm with minor comments

breerly:
:ship:

gen/namespace.go
abhinav: :+1:

gen/service.go
abhinav: Dropped in favor of `newNamespace` pattern used in YARPC code gen.

gen/yarpc.go
abhinav: Makes sense

abhinav:
Updated. Also updated code in https://github.com/yarpc/yarpc-go/pull/85

PR #110, opened 2016-04-05
{Unw,W}rapResponse: Primitive returns should not be pointers

abhinav:
This changes `WrapResponse` and `UnwrapResponse` to not use pointers
for primitive return values.


abhinav:
CC @prashantv

prashantv:
lgtm

Issue #77, opened 2016-03-22
JSON tags for generated structs


abhinav:
We need to add JSON tags to generated structs. See also https://github.com/thriftrw/thriftrw-go/issues/60

PR #111, opened 2016-04-06
Add json:".." tags to generated structs

abhinav:
An `omitempty` is added if the field is a pointer to a struct or an optional primitive.

Resolves https://github.com/thriftrw/thriftrw-go/issues/77

### DIFF
--- a/gen/field.go
+++ b/gen/field.go
@@ -65,15 +65,16 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
 		}`,
 		f,
 		TemplateFunc("tag", func(f *compile.FieldSpec) string {
-			// We want to add omitempty if the field is a struct or if it's an
-			// optional primitive so that we don't have "null" noise in the
-			// output. Note that binary is not a primitive.
+			// We want to add omitempty if the field is an optional struct or
+			// primitive to redure "null" noise. We won't add omitempty for
+			// optional collections because omitempty doesn't differentiate
+			// between nil and empty collections.
 
-			if isStructType(f.Type) || (isPrimitiveType(f.Type) && !f.Required) {
+			if (isStructType(f.Type) || isPrimitiveType(f.Type)) && !f.Required {
 				return fmt.Sprintf("`json:\"%s,omitempty\"`", f.Name)
 			}
 
-			return fmt.Sprintf("`json:\"%s\"`", f.Name)
+			return fmt.Sprintf("`json:%q`", f.Name)
 			// TODO(abg): Take go.tag and js.name annotations into account
 		}),
 	)
--- a/gen/field.go
+++ b/gen/field.go
@@ -20,7 +20,11 @@
 
 package gen
 
-import "github.com/thriftrw/thriftrw-go/compile"
+import (
+	"fmt"
+
+	"github.com/thriftrw/thriftrw-go/compile"
+)
 
 // fieldGroupGenerator is responsible for generating code for FieldGroups.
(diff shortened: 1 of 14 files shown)


abhinav:
@prashantv

gen/field.go
```diff
@@ -53,12 +57,26 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
-		}`, f)
+		}`,
+		f,
+		TemplateFunc("tag", func(f *compile.FieldSpec) string {
+			// We want to add omitempty if the field is a struct or if it's an
+			// optional primitive so that we don't have "null" noise in the
+			// output. Note that binary is not a primitive.
+
+			if isStructType(f.Type) || (isPrimitiveType(f.Type) && !f.Required) {
+				return fmt.Sprintf("`json:\"%s,omitempty\"`", f.Name)
```
prashantv: required structs should not need `omitempty` either since they should always be set, and if not, i think we want to see "null" so we know it's not set

gen/field.go
```diff
@@ -53,12 +57,26 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+			}
+
+			return fmt.Sprintf("`json:\"%s\"`", f.Name)
```
prashantv: nit: you probably want
```go
fmt.Sprintf("json:%q", f.Name)
```

gen/struct_test.go
```diff
@@ -677,3 +679,97 @@ func TestEmptyException(t *testing.T) {
...
+				"bool":   {BoolValue: boolp(true)},
+				"int64":  {Int64Value: int64p(42)},
+				"string": {StringValue: stringp("foo")},
+			}},
+			`{"listValue":null,"mapValue":{` +
+				`"bool":{"boolValue":true,"listValue":null,"mapValue":null},` +
+				`"int64":{"int64Value":42,"listValue":null,"mapValue":null},` +
+				`"string":{"stringValue":"foo","listValue":null,"mapValue":null}` +
+				`}}`,
+		},
```
prashantv: add a test where a required field of type struct is missing

gen/field.go
abhinav: :+1:

gen/field.go
abhinav: :+1:

gen/struct_test.go
abhinav: :+1:

abhinav:
Updated

gen/field.go
```diff
@@ -53,12 +57,27 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+		}`,
+		f,
+		TemplateFunc("tag", func(f *compile.FieldSpec) string {
+			// We want to add omitempty if the field is an optional struct or
+			// primitive to redure "null" noise. We won't add omitempty for
+			// optional collections because omitempty doesn't differentiate
+			// between nil and empty collections.
+
+			if (isStructType(f.Type) || isPrimitiveType(f.Type)) && !f.Required {
+				return fmt.Sprintf("`json:\"%s,omitempty\"`", f.Name)
```
prashantv: (no action for now)
We may want "plugins" for tag generation in future, since there's also "db" tags, people may want custom "json" tags etc. I imagine we'd want some way to get a list of tags, probably like `map[string]string{"json": "name,omitempty"}` that we then convert to a single tag.

abhinav: true

prashantv:
lgtm

PR #112, opened 2016-04-06
yarpc: use separate namespaces for server methods

abhinav:
Updated code in https://github.com/yarpc/yarpc-go/pull/85

@prashantv


prashantv:
lgtm

PR #107, opened 2016-04-06
Make typedef cycle detection less strict

abhinav:
The following cycles is valid:

```
typedef node head

struct node {
    1: required i32 value
    2: optional head next
}
```

Depends on https://github.com/thriftrw/thriftrw-go/pull/113


prashantv:
lgtm

PR #116, opened 2016-04-19
Don't count vendored code in coverage.

abhinav:
This wasn't a problem for thriftrw so far because we only have test
dependencies vendored right now.

@prashantv


prashantv:
lgtm

PR #117, opened 2016-05-16
gen: Add helper to reduce duplication in tests

abhinav:
Most of the tests were doing this:

    v = tt.x.ToWire()
    // check if v == tt.v

    var x TypeOfX
    if assert.NoError(t, x.FromWire(tt.v)) {
        assert.Equal(t, x, tt.x)
    }

I moved this into a helper (which needs to use reflection for the second part
of the test), and changed all existing tests to use this.

This also allowed me to combine all the table tests for structs that just do round trips.


blampe:
Nice, 🚢

breerly:
Hard to complain 🐑

PR #118, opened 2016-05-16
Support for unhashable types

### LINKED ISSUE #75: Sets and maps with unhashable items
We should use slices for maps/sets when the key is an unhashable type

abhinav:
This adds support for using unhashable types as map keys and set items. A type
is considered hashable only if it is a non-`binary` primitive type. All other
types are considered unhashable.

Sets of unhashable objects are downgraded to lists, and maps with unhashable
keys are downgraded to lists of key-value pairs. De-duplication is left up to
the user.

Resolves #75.

----

Don't let the line count scare you. Most of it is generated code and tests.

### DIFF
--- a/gen/container_test.go
+++ b/gen/container_test.go
@@ -415,12 +415,12 @@ func TestCrazyTown(t *testing.T) {
 			"ListOfSets",
 			tc.ContainersOfContainers{
 				ListOfSets: []map[int32]struct{}{
-					map[int32]struct{}{
+					{
 						1: struct{}{},
 						2: struct{}{},
 						3: struct{}{},
 					},
-					map[int32]struct{}{
+					{
 						4: struct{}{},
 						5: struct{}{},
 						6: struct{}{},
@@ -458,12 +458,12 @@ func TestCrazyTown(t *testing.T) {
 			"ListOfMaps",
 			tc.ContainersOfContainers{
 				ListOfMaps: []map[int32]int32{
-					map[int32]int32{
+					{
 						1: 100,
 						2: 200,
 						3: 300,
 					},
-					map[int32]int32{
+					{
 						4: 400,
 						5: 500,
 						6: 600,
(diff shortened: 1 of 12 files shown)


abhinav:

gen/container_test.go
```diff
@@ -369,3 +369,465 @@ func TestListOfStructs(t *testing.T) {
 		assertRoundTrip(t, &tt.s, tt.v, "Graph")
 	}
 }
+
+func TestCrazyTown(t *testing.T) {
```
breerly: <3

breerly: This test is crystal clear - great work

breerly:
lgtm

blampe:
Impressive!

gen/container_test.go
```diff
@@ -369,3 +369,465 @@ func TestListOfStructs(t *testing.T) {
...
+						}),
+					}),
+				})},
+			}}),
+		},
+		{
+			"ListOfSets",
+			tc.ContainersOfContainers{
+				ListOfSets: []map[int32]struct{}{
+					map[int32]struct{}{
```
prashantv: you shouldn't need this `map[int32]struct{}` here,
https://play.golang.org/p/q-rkKLbw6z

abhinav: 👍

gen/type.go
```diff
@@ -157,6 +157,10 @@ func typeName(g Generator, spec compile.TypeSpec) (string, error) {
 		if err != nil {
 			return "", err
 		}
+		if !isPrimitiveType(s.KeySpec) {
```
prashantv: can we create a new function `isHashable` which does this check and is used here and below.

gen/map.go
```diff
@@ -61,16 +61,23 @@ func (m *mapGenerator) ItemList(g Generator, spec *compile.MapSpec) (string, err
...
+			<$i := newVar "i">
 			func (<$m> <.Name>) ForEach(<$f> func(<$wire>.MapItem) error) error {
-				for <$k>, <$v> := range <$m> {
-					err := <$f>(<$wire>.MapItem{
-						Key: <toWire .Spec.KeySpec $k>,
-						Value: <toWire .Spec.ValueSpec $v>,
-					})
-					if err != nil {
-						return err
+				<if isPrimitiveType .Spec.KeySpec>
```
prashantv: let's use a `isHashable` function here and for the sets.

prashantv:
lgtm, really cool that is needed minimal changes to the code. just one minor nit, let's add a new function `isHashable` or something, I think it'll make the code clearer, and makes it easier to change if we wanted to.

gen/type.go
abhinav: 👍

gen/map.go
abhinav: 👍

abhinav:
Updated

PR #119, opened 2016-05-17
constant values: Support casting

abhinav:
This adds support for constant values to be casted to types based on context.
The following casts are supported:

-   `int` to `double`
-   `int` to `bool`
-   `int` to `enum`
-   `map[string]*` to `struct`
-   `list` to `set`

This is the same behavior supported by Apache Thrift, but with an extra level
of safety and type checking.

We need this before we can implement code generation for constant values.

### DIFF
--- a/compile/constant_value.go
+++ b/compile/constant_value.go
@@ -96,8 +96,7 @@ func (c ConstantInt) Link(scope Scope, t TypeSpec) (ConstantValue, error) {
 	case DoubleSpec:
 		return ConstantDouble(float64(c)).Link(scope, t)
 	case BoolSpec:
-		v := int64(c)
-		switch v {
+		switch v := int64(c); v {
 		case 0, 1:
 			return ConstantBool(v == 1).Link(scope, t)
 		default:
--- a/compile/constant_value.go
+++ b/compile/constant_value.go
@@ -40,6 +40,8 @@ func compileConstantValue(v ast.ConstantValue) ConstantValue {
 		return nil
 	}
 
+	// TODO(abg): Support typedefs
+
 	switch src := v.(type) {
 	case ast.ConstantReference:
 		return constantReference(src)
@@ -147,6 +149,20 @@ type ConstantStruct struct {
 	Fields map[string]ConstantValue
 }
 
+// buildConstantStruct builds a constant struct from a ConstantMap.
+func buildConstantStruct(c ConstantMap) (*ConstantStruct, error) {
+	fields := make(map[string]ConstantValue, len(c))
+	for _, pair := range c {
+		s, isString := pair.Key.(ConstantString)
+		if !isString {
+			return nil, fmt.Errorf(
+				"%v is not a string: all keys must be strings", pair.Key)
+		}
+		fields[string(s)] = pair.Value
+	}
+	return &ConstantStruct{Fields: fields}, nil
+}
+
 // Link for ConstantStruct
 func (c *ConstantStruct) Link(scope Scope, t TypeSpec) (ConstantValue, error) {
 	s, ok := t.(*StructSpec)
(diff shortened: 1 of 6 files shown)


abhinav:

compile/constant_value.go
```diff
@@ -76,24 +83,110 @@ type (
...
-	return c, nil
+func (c ConstantInt) Link(scope Scope, t TypeSpec) (ConstantValue, error) {
+	switch t {
+	case I8Spec, I16Spec, I32Spec, I64Spec:
+		// TODO bounds checks?
+		return c, nil
+	case DoubleSpec:
+		return ConstantDouble(float64(c)).Link(scope, t)
+	case BoolSpec:
+		v := int64(c)
```
prashantv: nit: inline in the switch, `switch v := int64(c); v {`

abhinav: 👍

compile/constant_value.go
```diff
@@ -76,24 +83,110 @@ type (
...
+					return nil, constantValueCastError{
+						Value:  c,
+						Type:   t,
+						Reason: fmt.Errorf("%q is a required field", field.Name),
+					}
+				}
+				continue
+			}
+			f = field.Default
+			c.Fields[field.Name] = f
```
prashantv: Do you need to do this? It loses information about whether the user specified an override which happened to be the default, and the user of the library would never be able to check that right?

compile/constant_value.go
```diff
@@ -118,20 +211,39 @@ type ConstantValuePair struct {
 }
 
 // Link for ConstantMap.
-func (c ConstantMap) Link(scope Scope) (ConstantValue, error) {
-	items := make([]ConstantValuePair, len(c))
+func (c ConstantMap) Link(scope Scope, t TypeSpec) (ConstantValue, error) {
+	if _, isStruct := t.(*StructSpec); isStruct {
```
prashantv: nit: let's move this logic into a separate function where we can reduce nesting

compile/constant_value.go
abhinav: We don't need to do this. This will apply the default at code-generation time. Should we apply it at send time instead?

compile/constant_value.go
```diff
@@ -143,6 +255,29 @@ func (c ConstantMap) Link(scope Scope) (ConstantValue, error) {
 	return ConstantMap(items), nil
 }
 
+// ConstantSet represents a set of constant values from the Thrift file.
+type ConstantSet []ConstantValue
+
+// Link for ConstantSet.
+func (c ConstantSet) Link(scope Scope, t TypeSpec) (ConstantValue, error) {
+	s, ok := t.(*SetSpec)
```
prashantv: optional: could you do:
```go
v, err := ConstantList(c).Link(Scope, t)
if err != nil {
  return nil, err
}
return ConstantSet(v), nil
```

compile/constant_value.go
```diff
@@ -179,9 +319,12 @@ type ConstReference struct {
 }
 
 // Link for ConstReference.
-func (c ConstReference) Link(scope Scope) (ConstantValue, error) {
-	// The reference has already been verified. Nothing to do.
-	return c, nil
+func (c ConstReference) Link(scope Scope, t TypeSpec) (ConstantValue, error) {
+	cv, err := typeEqualsOrCastError(c, c.Target.Type, t)
```
prashantv: does this mean you can't do
```thrift
const i32 A = 1
const double B = A
```

prashantv:
```
$ cat test.thrift
const i32 A = 1
const double B = A

$ thrift --gen go test.thrift

$ go build ./gen-go/test

```

compile/constant_value.go
```diff
@@ -76,24 +83,109 @@ type (
...
+		return nil, constantValueCastError{
+			Value: c,
+			Type:  t,
+			Reason: fmt.Errorf(
+				"%v is not a valid value for enum %q", int32(c), e.ThriftName()),
+		}
+	}
+
+	return nil, constantValueCastError{Value: c, Type: t}
+	// TODO: AST for constants will need to track positions for us to
```
breerly: +1

compile/constant_value.go
abhinav: 👍

compile/constant_value.go
abhinav: The problem with that is that now ConstantList needs to have more special handling for when TypeSpec is a `*SetSpec` rather than just delegating to ConstantSet.

compile/constant_value.go
abhinav: As discussed offline: If the types match, we'll use the constant reference as-is (so the generated code will have a reference to a declared constant), otherwise, we'll resolve the reference away to a new constant value.

abhinav:
Updated

compile/constant_value.go
abhinav: As per offline discussion: We're going to apply the defaults for constant values in the generated code.

prashantv:
lgtm

PR #121, opened 2016-05-18
Generate code for consts

### LINKED ISSUE #70: Constant value code generation
We need to generate code for constants and defaults declared in the Thrift file

abhinav:
This adds support for generating code for `const` declarations in Thrift files.

Resolves #70


abhinav:
This is ready for review.

compile/constant_value.go
```diff
@@ -274,6 +274,7 @@ func (c ConstantSet) Link(scope Scope, t TypeSpec) (ConstantValue, error) {
 	}
 
 	// TODO(abg): Track whether things are linked so that we don't re-link here
+	// TODO(abg): Fail for duplicates
```
breerly: any reason this can't go in this PR?

gen/constant.go
```diff
@@ -20,17 +20,222 @@
...
+}
+
+// helper to generate pointers to primitives
+func primitiveValueRef(g Generator, c compile.ConstantValue, t compile.TypeSpec) (string, error) {
+	var toptr string
+	switch t {
+	case compile.BoolSpec:
+		toptr = "_boolptr"
+		err := g.EnsureDeclared(
+			`func _boolptr(v bool) *bool { return &v }`, nil)
```
breerly: oh no lol

gen/testdata/constants/constants.go
```diff
@@ -0,0 +1,51 @@
...
+
+func _boolptr(v bool) *bool {
+	return &v
+}
+
+func _i64ptr(v int64) *int64 {
+	return &v
+}
+
+func _stringptr(v string) *string {
```
breerly: @prashantv is there a way to do this in Go without generating these lil helpers?

breerly:
lgtm

(wondering if those constant helpers should go in the lib instead of being generated, but I don't feel strongly).

compile/constant_value.go
abhinav: It's non-trivial to implement because the objects aren't necessarily hashable, and it's low priority enough that it made sense to defer.

gen/constant.go
```diff
@@ -20,17 +20,222 @@
...
+		if v {
+			return "true", nil
+		}
+		return "false", nil
+	case compile.ConstantDouble:
+		return fmt.Sprint(float64(v)), nil
+	case compile.ConstantInt:
+		return fmt.Sprint(int(v)), nil
+	case compile.ConstantList:
+		l, ok := t.(*compile.ListSpec)
```
prashantv: can we break the more complex constant generation things into their own functions. there doesn't seem to be any advantage to having them inline here. maybe the type check can be done in this function, but then you can call a function to do the rest?

gen/constant.go
```diff
@@ -20,17 +20,222 @@
...
+		if !ok {
+			return "", fmt.Errorf(
+				"cannot cast list %v to %v", v, t.ThriftName())
```
prashantv: what does this error message look like? does it say something like "list ConstantList string" ?

gen/constant.go
```diff
@@ -20,17 +20,222 @@
...
+
+// helper to generate pointers to primitives
+func primitiveValueRef(g Generator, c compile.ConstantValue, t compile.TypeSpec) (string, error) {
+	var toptr string
+	switch t {
+	case compile.BoolSpec:
+		toptr = "_boolptr"
+		err := g.EnsureDeclared(
+			`func _boolptr(v bool) *bool { return &v }`, nil)
+		if err != nil {
```
prashantv: nit: since the error checking and handling is the same, you could declare `var err error` before the switch, and just do `err = g.EnsureDeclared(...)` in your cases

gen/constant.go
```diff
@@ -20,17 +20,222 @@
...
+			return "", err
+		}
+	case compile.StringSpec:
+		toptr = "_stringptr"
+		err := g.EnsureDeclared(
+			`func _stringptr(v string) *string { return &v }`, nil)
+		if err != nil {
+			return "", err
+		}
+	default:
```
prashantv: here and below, don't need empty default case

gen/constant.go
```diff
@@ -20,17 +20,222 @@
...
+	switch t {
+	case compile.BoolSpec:
+		toptr = "_boolptr"
+		err := g.EnsureDeclared(
+			`func _boolptr(v bool) *bool { return &v }`, nil)
+		if err != nil {
+			return "", err
+		}
+	case compile.I8Spec:
+		toptr = "_i8ptr"
```
prashantv: optional: since we parse the template and get the function declaration we find, should `EnsureDeclared` just return the name rather than duplicating it?

gen/constant.go
```diff
@@ -20,17 +20,222 @@
...
+		panic(fmt.Sprintf(
+			"primitiveValueRef called with %v which is not a primitive",
+			t.ThriftName()))
+	}
+
+	s, err := ConstantValue(g, c, t)
+	if err == nil {
+		s = fmt.Sprintf("%v(%v)", toptr, s)
+	}
+	return s, err
```
prashantv: do we want to return the string even if there's an error?
why not write this as
```go
s, err := ConstantValue(..)
if err != nil {
  return "", err
}

return fmt.Sprintf(...)
```

gen/constant_test.go
```diff
@@ -0,0 +1,248 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package gen
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
prashantv: nit: import order

gen/testdata/constants/constants.go
prashantv: unfortunately nope, no other way for literal values. we could move these into a `thriftrw` package though, rather than generating them with every package.

gen/testdata/thrift/constants.thrift
```diff
@@ -0,0 +1,86 @@
+include "./other_constants.thrift"
+include "./containers.thrift"
+include "./enums.thrift"
+include "./exceptions.thrift"
+include "./structs.thrift"
+include "./unions.thrift"
+
+const containers.PrimitiveContainers primitiveContainers = {
+    "listOfInts": other_constants.listOfInts, // imported constant
```
prashantv: github seems to be unhappy with `// ` comments in this file. should we be using something else?

prashantv:
Looks pretty good, just a few small comments

gen/constant.go
abhinav: 👍

gen/constant.go
abhinav: It looks like,

    cannot cast [1 2 3] to foo

Although this specific error won't usually happen in this part of the code because the compile step already type checked everything. It *would* be okay to panic here but I figured it's safer to fail with an actual error regardless.

gen/constant.go
abhinav: 👍

gen/constant.go
abhinav: 👍

gen/constant.go
abhinav: If we do that, we will have to limit `EnsureDeclared` to a single declaration. Right now, it's the same as `DeclareFromTemplate`: You can have any number of declarations in it.

gen/constant.go
abhinav: How about,

    s, err := ...
    s = fmt.Sprintf(...)
    return s, err

The value of s will get ignored in case of an error anyway.

prashantv: sure

gen/constant_test.go
abhinav: 👍 

I really wish goimports would take care of this

gen/testdata/thrift/constants.thrift
abhinav: This is a valid comment as per the [Thrift scanner](https://github.com/apache/thrift/blob/775671aea41ea55427dd78d7ce68e282cc9b8487/compiler/cpp/src/thriftl.ll#L116). Not sure how to make GH happy.

abhinav:
Updated

prashantv:
lgtm

PR #123, opened 2016-05-18
gen/constant: Simplify const-pointer gen logic

abhinav:
Turns out we don't need to do all that special casing.


abhinav:

prashantv:
lgtm

Issue #100, opened 2016-04-01
Comments in the generated code


abhinav:
Our current code generation schemes makes it difficult to include comments in the generated code. If we use the `parser.ParseComments` flag, the comments appear out-of-place in the output. I think we need to fix the `token.Pos`s associated with comments.

PR #124, opened 2016-05-19
structs: Apply default values at encode/decode time

abhinav:
This changes code generation for structs to apply default values at
encode/decode time if they're nil. Fields that have a default value are
effectively optional in the generated code.

CC @prashant @breerly @kriskowal

### DIFF
--- a/gen/testdata/thrift/structs.thrift
+++ b/gen/testdata/thrift/structs.thrift
@@ -1,3 +1,5 @@
+include "./enums.thrift"
+
 struct EmptyStruct {}
 
 //////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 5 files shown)


breerly:
Just to be clear - "effectively optional" is the behavior of python, node, and java in the same case?

abhinav:
Almost. Python always applies the defaults if you the required-with-defaults fields are None, so that's basically the same behavior. Node applies the defaults at deserialization time, but not at serialization time (this will be fixed in https://github.com/thriftrw/thriftrw-node/issues/138 CC @kriskowal).

gen/testdata/thrift/structs.thrift
```diff
@@ -74,4 +76,26 @@ struct Node {
     2: optional List next
 }
 
-// TODO: Default values
+
+//////////////////////////////////////////////////////////////////////////////
+// Default values
+
+struct DefaultsStruct {
```
breerly: You copying snippets from ThriftTest.thrift? Or just writing them by hand as needed?

abhinav: By hand

gen/testdata/thrift/structs.thrift
```diff
@@ -74,4 +76,26 @@ struct Node {
...
+    1: required i32 requiredPrimitive = 100
```
breerly: wondering if its worth generating a comment near required-w-default fields like:

```
// required fields with a default are effectively optional
1: required i32 requiredPrimitive = 100
```

breerly:
lgtm 🐑

gen/testdata/thrift/structs.thrift
abhinav: Our code generation scheme doesn't support comments yet. https://github.com/thriftrw/thriftrw-go/issues/100

prashantv:
looks good, 🚢

kriskowal:
Let’s write this behavior down in the ThriftRW repository.

PR #125, opened 2016-05-20
Add support for message envelopes

### LINKED ISSUE #105: Message envelope support
thriftrw go needs to support thrift message envelopes before we can support HTTP

prashantv:
Fixes #105

### DIFF
--- a/protocol/binary/envelope.go
+++ b/protocol/binary/envelope.go
@@ -1,0 +1,134 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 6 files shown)


blampe:
Awesome!

Are the tests borrowed from node/python?

prashantv:
The non-strict test is, but the strict tests are different since python writes non-strict envelopes, while Go writes strict envelopes.

protocol/binary_test.go
```diff
@@ -669,3 +669,151 @@ func TestStructOfContainers(t *testing.T) {
...
 }
+
+func TestBinaryEnvelopeErrors(t *testing.T) {
+	tests := []struct {
+		encoded []byte
+		errMsg  string
+	}{
+		{
+			encoded: []byte{
+				0x80, 0x02, 0x00, 0x01, // version|type:4 = 2 | call
```
abhinav: This doesn't seem right. All the examples I'm seeing in thriftrw python and node have `0x80, 0x02` as the first 2 bytes.

https://github.com/thriftrw/thriftrw-python/blob/master/tests/protocol/test_binary.py#L405
https://github.com/thriftrw/thriftrw-node/blob/master/test/message.js#L89-90

abhinav:
yay

protocol/protocol.go
```diff
@@ -33,6 +33,14 @@ type Protocol interface {
 	// Encode the given Value and write the result to the given Writer.
 	Encode(v wire.Value, w io.Writer) error
 
+	// EncodeEnveloped encodes the enveloped value and writes the result
+	// to the given Writer.
+	EncodeEnveloped(e wire.Envelope, w io.Writer) error
```
breerly: This api is clean 👍

protocol/binary/envelope.go
```diff
@@ -0,0 +1,134 @@
...
+// Name (4 byte length prefixed string)
+// Sequence ID (4 bytes)
+//
+// When reading payloads, we need to support both strict and non-strict
+// payloads. To do this, we read the first 4 byte. Non-strict payloads
+// will always have a size >= 0, while strict payloads have selected
+// version numbers such that the value will always be negative.
+func (bw *Reader) ReadEnveloped() (wire.Envelope, error) {
+	var e wire.Envelope
+	initial, off, err := bw.readInt32(0)
```
kriskowal: I think the first byte `&^ 0x80` is a version number if first byte < 0, or the version is `0`. Four bytes are overkill.

prashantv: Haha, it is indeed overkill by 4x the amount of bytes we need. However, we pass the `initial` in to `readStrictNameType` which does need the rest of the bytes (rather than reading the 4 bytes again).

Mostly because at some point I'd like to move away from `io.ReaderAt` and make it just an `io.Reader`, where we can only consume bytes once.

PR #126, opened 2016-05-20
Support serializing arg structs with an envelope

prashantv:


compile/service.go
```diff
@@ -207,6 +210,20 @@ func (f *FunctionSpec) Link(scope Scope) error {
...
+}
+
+// Type returns the envelope type that is used when making enveloped requests
+// for this function.
+func (f *FunctionSpec) Type() wire.EnvelopeType {
+	if f.OneWay {
+		return wire.OneWay
+	}
+	return wire.Call
+}
```
abhinav: FunctionSpec probably shouldn't implement this interface.

prashantv: Renamed to `CallType` so it doesn't implement the interface, but still gives access to the `wire.EnvelopeType`

prashantv:
Updated with support for replies.

envelope/envelope.go
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
abhinav: 2016

abhinav: Use the `updateLicenses` script from the scripts directory

gen/service.go
```diff
@@ -280,3 +286,37 @@ func functionUnwrapResponse(g Generator, f *compile.FunctionSpec) (string, error
...
+		`
+		<$wire := import "github.com/thriftrw/thriftrw-go/wire">
+		<$argType := printf "%v%v" (goCase .Name) "Args">
+		<$v := newVar "v">
+
+		func (<$v> *<$argType>) MethodName() string {
+			return "<.MethodName>"
+		}
+
+		func (<$v> *<$argType>) Type() <$wire>.EnvelopeType {
```
abhinav: This makes it so `type` and `methodname` (and other-cased variants) can't be used as parameter names. `type` is already reserved by the namespacer, but `methodname` is not. You should probably add a TODO somewhere about handling that conflict (probably in the same place with the TODO for `towire/fromwire` conflicts).

gen/service_test.go
```diff
@@ -446,3 +450,51 @@ func TestUnwrapResponse(t *testing.T) {
 		}
 	}
 }
+
+func TestServiceArgsEnveloper(t *testing.T) {
```
abhinav: Outdated test name. This tests the replies too.

abhinav:
👍

envelope/envelope.go
prashantv: :+1:

gen/service.go
prashantv: Good point, perhaps we should rename the interface methods now, since `type` will show up pretty often in user types.

I'll call it `EnvelopeType` and put a TODO in to figure out how to deal with conflicts.

gen/service_test.go
prashantv: 👍

gen/service.go
abhinav: I meant that `Type()` is okay because `type` is already a reserved keyword, but yeah, `EnvelopeType` works better because `Type` is fairly general.

PR #128, opened 2016-05-21
wire: Add EvaluateValue

abhinav:
This will fully evaluate a Value object and raise any lazy errors.

CC @breerly @prashant


breerly:
Thx for combining those
🚢

PR #129, opened 2016-05-21
gen: Validate structs and unions

### LINKED ISSUE #76: Struct validation
We need to validate that required fields of structs (and exceptions) are present and that unions have at exactly one populated field.

abhinav:
This changes the generated code for struct-like types to check the following:

-   structs and exceptions: All required fields must be set. Note that
    primitive required fields are always set so they cannot be checked during
    serialization.
-   unions: For non-empty unions, exactly one field must be set.
-   function results: Exactly one field must be set if the function has a
    non-void return value. At most one field must be set if the function
    has a void return value.

Resolves #76


abhinav:
This is ready for review. CC @prashantv @breerly @kriskowal

gen/field.go
```diff
@@ -81,57 +86,95 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
-		func (<$v> *<.Name>) ToWire() <$wire>.Value {
+		func (<$v> *<.Name>) ToWire() (<$wire>.Value, error) {
     		<$fields := newVar "fields">
     		<$i := newVar "i">
-			// TODO check if required fields that are reference types are nil
+			<$wVal := newVar "w">
 
-			var <$fields> [<len .Fields>]<$wire>.Field
-			<$i> := 0
+			<if and .IsUnion (len .Fields)>
```
prashantv: can we move the validation logic to a separate `Validate` function? Will clean up the `ToWire` code and allow us to only run validation.

prashantv: actually never mind, I see you're validating `required` before generation, which is fine.

Maybe we should move just the `IsUnion` validation to a separate function that's private?

gen/field.go
```diff
@@ -81,57 +86,95 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+				<$fmt := import "fmt">
+				<$count := newVar "count">
+				<$count> := 0
+				<range .Fields>
+					if <$v>.<goCase .Name> != nil { <$count>++ }
```
prashantv: any reason why this validation is done separately, instead of just using `<$i>` after all the field validation? While required validation is done inline

gen/field.go
```diff
@@ -81,57 +86,95 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+			<$structName := .Name>
 			<range .Fields>
 				<$f := printf "%s.%s" $v (goCase .Name)>
 				<if .Required>
-					<$wVal := toWire .Type $f>
-					<$fields>[<$i>] = <$wire>.Field{ID: <.ID>, Value: <$wVal>}
-					<$i>++
-				<else>
-					<if .Default>
+					<if not (isPrimitiveType .Type)>
```
prashantv: do we have `Required` types that end up being primitives that we can't do a `nil` check on?

gen/field.go
```diff
@@ -81,57 +86,95 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+							if err != nil {
+								return <$wVal>, err
+							}
 							<$fields>[<$i>] = <$wire>.Field{
 								ID: <.ID>,
-								Value: <toWirePtr .Type $f>,
+								Value: <$wVal>,
 							}
 							<$i>++
+					<if not .Default>
```
prashantv: this is kind of confusing. what might be clearer is even in the `if .Default` case,  to just open a new brace, so that you always need to close regardless.

gen/field.go
abhinav: Ah, I think I just copied this blob over from the `FromWire` validator. This can be done with `<$i>` after serializing, yeah. Updating.

abhinav: Turns out count tracking for `ToWire` for unions can similarly be interleaved with the logic to parse values

gen/field.go
abhinav: Now that it's interleaved with how values are read/written, that's slightly more difficult to do.

gen/field.go
abhinav: Discussed offline. Invalid comment.

gen/field.go
abhinav: 👍

abhinav:
Updated. Ready for another look.

gen/field.go
```diff
@@ -157,22 +214,51 @@ func (f fieldGroupGenerator) FromWire(g Generator) error {
 						<end>
 						if err != nil {
 							return err
+							// TODO: Nest the error inside a "failed to read
+							// field X of struct Y" error.
 						}
+						<if .Required>
+							<$isSet.Rotate (printf "%sIsSet" .Name)> = true
+						<end>
+						<if $isUnion><$count>++<end>
```
prashantv: What if you receive the same field ID twice? I think that should only count as a single field, and shouldn't cause an error. (let's also add a unit test for that case)

gen/field.go
```diff
@@ -81,57 +86,95 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
-					<end>
+				<end>
+			<end>
+
+			<if and .IsUnion (len .Fields)>
+				<$fmt := import "fmt">
+				<if .AllowEmptyUnion>
+					if <$i> > 1 {
+						return <$wire>.Value{}, <$fmt>.Errorf(
+							"<.Name> should receive at most one field value: received %v values", <$i>)
```
prashantv: is `receive` the right word for `ToWire`? should we say only a single field should be set?

prashantv: also since this check is almost the same (maybe with a slight message difference) with `FromWire`, is it worth sharing in some way?

prashantv:
Is there a test for an empty union? We have checks for `IsUnion && len(fields)` which I assume is to allow an empty union to skip checks. What should an empty union actually mean?

gen/field.go
abhinav: Derp. Good catch.

gen/field.go
abhinav: True. I'll fix the error message.

It's non-trivial to share this code because they still have different return values. We'll probably want a library-level error type which is instantiated by that code.

abhinav:
There are tests for empty unions. https://github.com/thriftrw/thriftrw-go/pull/129/files#diff-415115ad6a457ba2707d1ff03de07c96R366

Empty union is equivalent to empty struct.

abhinav:
Updated

prashantv:
lgtm

PR #132, opened 2016-05-25
compile: Add a non-strict mode for compilation

prashantv:
In tools like yab, we want to be able to opt-out of strict mode.

### DIFF
--- a/compile/compiler_test.go
+++ b/compile/compiler_test.go
@@ -84,3 +84,27 @@ func TestCompile(t *testing.T) {
 	require.NoError(t, err, "Lookup KeyValue failed")
 	require.NotNil(t, kvSvc, "KeyValue service is nil")
 }
+
+func TestCompileNonStrict(t *testing.T) {
+	files := map[string]string{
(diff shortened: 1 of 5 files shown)


compile/compiler_test.go
```diff
@@ -84,3 +84,23 @@ func TestCompile(t *testing.T) {
...
+
+	fs := dummyFS{"/some/prefix/", files}
+
+	module, err := Compile("main.thrift", Filesystem(fs), NonStrict())
+	require.NoError(t, err, "Compile failed")
+
+	sType, err := module.LookupType("S")
+	require.NoError(t, err, "Lookup S failed")
+	require.NotNil(t, sType, "Type S is nil")
+	assert.Equal(t, wire.TStruct, sType.TypeCode(), "Type mismatch")
```
abhinav: You should verify that the uuid field is optional

abhinav:
LGTM

compile/compiler_test.go
prashantv: 👍

Issue #48, opened 2016-01-20
Strict mode


abhinav:
We should consider implementing a strict mode at the library level that does this:
- If a field type is a mismatch, instead of ignoring it, fail the request.
- If a container's element type does not match what is expected from that container, fail the request.

(Update this issue with any other requirements that come up for strict mode.)

prashantv:
I think we should be strict by default (at least in YARPC). Errors that this strict mode would catch would protect against backwards incompatible changes rather than ignoring the fields

cc @yarpc

breerly:
Not a terrible idea - @abhinav what do you think?

blampe:
Was this resolved by 785a6a0?

abhinav:
No, that just adds non-strict mode to loosen up current thriftrw restrictions. Prashant is proposing adding a more strict mode.

prashantv:
Our naming isn't great. There's 2 kinds of strict mode:
 - one for parsing of the Thrift file (which is what https://github.com/thriftrw/thriftrw-go/commit/785a6a01f441db024e5666df9ebff001ba50c13d lets you turn off)
 - one for parsing of bytes read on the wire (which is what this issue is about)

Right now, we're very lax when we receive binary payloads we don't expect, there's no warnings, or errors, but we just skip bad data. This means that when someone makes a backwards incompatible change is made by a service, it's not easily noticed (e.g. if they deploy a canary) since there's no errors or any way to track this. I think we should reject payloads instead so we catch these issues earlier.

abhinav:
This is unlikely to happen unless someone makes a PR for it. I'm closing it
until future updates.

PR #133, opened 2016-05-25
envelope: Decode exception payloads using TApplicationException

prashantv:
We might want to expose methods to check properties on `TApplicationException` in future.


envelope/envelope.go
```diff
@@ -48,3 +52,31 @@ func Write(p protocol.Protocol, w io.Writer, seqID int32, e Enveloper) error {
 		Value: body,
 	}, w)
 }
+
+// ReadReply reads an envelope from the given reader, and decodes any
+// Exception paylods.
```
abhinav: Nit: Decoding exception payloads is an implementation detail. Just mentioning that we're reading enveloped responses should be enough.

envelope/envelope.go
```diff
@@ -48,3 +52,31 @@ func Write(p protocol.Protocol, w io.Writer, seqID int32, e Enveloper) error {
...
+	envelope, err := p.DecodeEnveloped(r)
+	if err != nil {
+		return wire.Value{}, 0, err
+	}
+
+	switch {
+	case envelope.Type == wire.Reply:
+		return envelope.Value, envelope.SeqID, nil
+	case envelope.Type != wire.Exception:
+		return envelope.Value, envelope.SeqID, fmt.Errorf("unknown envelope for reply, got %v", envelope.Type)
```
abhinav: "unknown envelope type"?

envelope/exception.thrift
```diff
@@ -0,0 +1,18 @@
...
+  BAD_SEQUENCE_ID = 4
+  MISSING_RESULT = 5
+  INTERNAL_ERROR = 6
+  PROTOCOL_ERROR = 7
+  INVALID_TRANSFORM = 8
+  INVALID_PROTOCOL = 9
+  UNSUPPORTED_CLIENT_TYPE = 10
+}
+
+struct TApplicationException {
```
abhinav: Why make this a struct instead of exception? It'll implement the `error` interface if you make it an `exception`.

envelope/envelope.go
prashantv: done

envelope/envelope.go
prashantv: 👍

envelope/exception.thrift
prashantv: good idea

prashantv:
Updated

envelope/envelope_test.go
```diff
@@ -0,0 +1,204 @@
...
+				0x00, // stop
+			},
+			want: wire.NewValueStruct(wire.Struct{
+				Fields: []wire.Field{
+					{1, wire.NewValueI32(1)},
+				},
+			}),
+			wantSeqID: 1234,
+			// TODO: This should probably fail to decode. Right now, it's being ignored.
+			// wantErr:   "failed to decode exception",
```
abhinav: This won't fail to decode because type mismatches for fields are ignored and both these fields are optional

envelope/envelope_test.go
```diff
@@ -0,0 +1,204 @@
...
+				0x80, 0x01, 0x00, 0x03, // version|type:4 = 3 | exception
+				0x00, 0x00, 0x00, 0x03, // name length = 3
+				'a', 'b', 'c', // "abc"
+				0x00, 0x00, 0x04, 0xd2, // seqID:4 = 1234
+
+				// <struct>
+				0x0b,       // type:1 = string
+				0x00, 0x01, // id:2 = 1
+				0x00, 0x00, 0x00, 0x06, // length = 3
+				'e', 'r', 'r', 'M', 's', 'g', // "msg"
```
abhinav: outdated comment

envelope/envelope_test.go
```diff
@@ -0,0 +1,204 @@
...
+			},
+			want: wire.NewValueStruct(wire.Struct{
+				Fields: []wire.Field{
+					{1, wire.NewValueString("errMsg")},
+					{2, wire.NewValueI32(1)},
+				},
+			}),
+			wantSeqID: 1234,
+			// TODO: The type should be "UknownMethod".
+			wantErr: "TApplicationException{Message: errMsg, Type: 1}",
```
abhinav: You should be ablet o use `ExceptionTypeUnknownMethod` for the Type

abhinav: My bad. Fixed in https://github.com/thriftrw/thriftrw-go/pull/136

envelope/envelope_test.go
prashantv: Yeah found https://github.com/thriftrw/thriftrw-go/issues/48 which is what I was expecting. I'm not sure how I feel about this "ignore backwards incompatible breaking change" behaviour, but we can discuss that separately.

envelope/envelope_test.go
prashantv: fixed

prashantv:
Rebased and updated

abhinav:
LGTM

PR #140, opened 2016-05-27
compile: Move dummyFS into separate file

abhinav:
Just moving to a more discoverable place. I plan to eventually change some of the tests
that have specs written by hand to use this instead.


prashantv:
lgtm

PR #141, opened 2016-05-27
compile: Add RootTypeSpec function

### LINKED ISSUE #138: compile: Provide a way to get the root type (traverse all typedefs)
Currently, users of `compile.TypeSpec` have to loop over the type while they see a `Typedef`. There should be a way to get the root type of a `TypeSpec`

abhinav:
This retrieves the root TypeSpec a typedef chain points to.

Resolves #138

### DIFF
--- a/compile/type.go
+++ b/compile/type.go
@@ -53,6 +53,17 @@ type TypeSpec interface {
 	ForEachTypeReference(func(TypeSpec) error) error
 }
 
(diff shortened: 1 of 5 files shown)


abhinav:
CC @prashantv @kriskowal

kriskowal:
:ok:

prashantv:
lgtm

PR #142, opened 2016-05-27
constants: Support typedefs

### LINKED ISSUE #120: Constant values for typedefs
We need to support things like,

    typedef i32 Foo

    const Foo x = 42

Currently, the checks performed by the constant value linker will fail for these cases.

abhinav:
Resolves #120

CC @prashantv @kriskowal

### DIFF
--- a/gen/constant.go
+++ b/gen/constant.go
@@ -69,46 +69,43 @@ func ConstantValue(g Generator, c compile.ConstantValue, t compile.TypeSpec) (st
 	}
 }
 
-func constantBool(g Generator, v compile.ConstantBool, t compile.TypeSpec) (string, error) {
+func castConstant(g Generator, t compile.TypeSpec, s string) (string, error) {
+	n, err := typeName(g, t)
+	if err != nil {
+		return "", err
+	}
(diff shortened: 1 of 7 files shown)


gen/constant.go
```diff
@@ -45,73 +45,97 @@ func Constant(g Generator, c *compile.Constant) error {
...
+	s := fmt.Sprint(int(v))
+	switch t {
+	case compile.I8Spec, compile.I16Spec, compile.I32Spec, compile.I64Spec:
+		// do nothing
+	default:
+		n, err := typeName(g, t)
+		if err != nil {
+			return "", err
+		}
+		s = fmt.Sprintf("%v(%v)", n, s)
```
prashantv: since the casting logic is always the same, should we have a function that does that?
then you could just do,
```go
default:
  return wrapType(g, t, s)
```

prashantv:
lgtm with small comment

gen/constant.go
abhinav: 👍

abhinav:
Updated

prashantv:
lgtm

PR #144, opened 2016-05-27
compile: Expose an empty scope

### LINKED ISSUE #135: compile: Expose an empty scope
We should expose an `EmptyScope` global in the compile package that defines a `compile.Scope` that always fails to look up. Something like [fakeScope](https://github.com/thriftrw/thriftrw-go/blob/master/compile/scope_test.go#L26) but with no actual lookups.

abhinav:
Resolves #135


abhinav:
@prashantv

prashantv:
lgtm

PR #145, opened 2016-05-31
lazy lists: Embed type information in interfaces

abhinav:
This allows us to get rid of the expensive Set, Map, and List structs,
reducing the overall size of the Value object.

So instead of value being one of,

    Set(ValueType, Size, ValueList)
    List(ValueType, Size, ValueList)
    Map(KeyType, ValueType, Size, MapItemList)

It's simply one of the corresponding lazy lists, and the lazy lists themselves
define `ValueType()`, `Size()`, etc. functions.

----

Testing against the degenerative test case with a `list<map<i32,set<string>>>` where each collection is 1000 elements, this showed a _slight_ improvement in serialization performance and the same number of allocs.

Before:

```
BenchmarkThriftRWWriteNestedMixed-8	     300	   5954503 ns/op	    2101 B/op	      63 allocs/op
```

After

```
BenchmarkThriftRWWriteNestedMixed-8	     300	   5521817 ns/op	    1940 B/op	      63 allocs/op
```

### DIFF
--- a/gen/container_test.go
+++ b/gen/container_test.go
@@ -40,12 +40,8 @@ func TestCollectionsOfPrimitives(t *testing.T) {
 			"empty list",
 			tc.PrimitiveContainers{ListOfInts: []int64{}},
 			wire.NewValueStruct(wire.Struct{Fields: []wire.Field{{
-				ID: 2,
-				Value: wire.NewValueList(wire.List{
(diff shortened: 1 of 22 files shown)


abhinav:
CC @prashantv

prashantv:
lgtm

PR #146, opened 2016-06-01
wire/value: Reduce number of fields

abhinav:
This significantly decreases the size of Value objects by combining all lazy lists into one `interface{}`, and all numeric fields and boolean into one `uint64`.

### DIFF
--- a/wire/value.go
+++ b/wire/value.go
@@ -39,9 +39,7 @@ type Value struct {
 	ti64    int64
 	tbinary []byte
 	tstruct Struct
-	tmap    MapItemList
-	tset    ValueList
-	tlist   ValueList
+	tcoll   interface{} // set/map/list
 }
 
 // Type retrieves the type of value inside a Value.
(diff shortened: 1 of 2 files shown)


abhinav:
CC @prashantv

wire/value.go
```diff
@@ -53,108 +47,110 @@ func (v *Value) Type() Type {
...
+	case TSet, TList:
+		return v.GetList()
 	default:
 		panic(fmt.Sprintf("Unknown value type %v", v.typ))
 	}
 }
 
 // NewValueBool constructs a new Value that contains a boolean.
 func NewValueBool(v bool) Value {
+	var n uint64 = 0
```
prashantv: nit: can drop the `= 0` since it's the default anyway. or if you want the 0 to be explicit, `n := uint64(0)`

prashantv:
This is great, lgtm

wire/value.go
abhinav: 👍

PR #147, opened 2016-06-01
ptr - Create package to convert basic types to pointers

breerly:
@abhinav @nomis52 @prashantv @akshayjshah

### DIFF
--- a/ptr/ptr.go
+++ b/ptr/ptr.go
@@ -1,0 +1,46 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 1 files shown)


ptr/ptr.go
```diff
@@ -0,0 +1,46 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package ptr
```
prashantv: nit: can you add package docs here

ptr/ptr.go
```diff
@@ -0,0 +1,46 @@
...
+
+// Byte converts an int8 to a pointer
+func Byte(x int8) *int8 {
```
abhinav: Rename to Int8 to match the others.

akshayjshah: This is a bit confusing - `byte` is an alias for `uint8`. This should either be `Int8` or take a `uint8`.

ptr/ptr.go
breerly: Done.

abhinav:
:shipit:

ptr/ptr.go
```diff
@@ -0,0 +1,46 @@
...
+	return &x
+}
+
```
prashantv: since we have all the int types except for `Int16`, should we add that as well?

ptr/ptr.go
```diff
@@ -0,0 +1,46 @@
...
+// Int64 converts an int64 to a pointer
+func Int64(x int64) *int64 {
+	return &x
+}
+
+// Double converts a float64 to a pointer
+func Double(x float64) *float64 {
```
prashantv: @abhinav should we also have `Byte` to match `Double`? Or should this also be changed to `Float64`?

ptr/ptr.go
breerly: Agreed, adding 👍  

Can you think of any others?

breerly: Added `Int16`.

prashantv:
lgtm

ptr/ptr.go
breerly: Changed to `Float64`.

PR #149, opened 2016-06-01
yarpc: Plug in application errors API

abhinav:
This lets us inform YARPC that the response contains an application error.


breerly:
Yessum 🐑

prashantv:
lgtm

PR #151, opened 2016-06-08
ptr - Add Int()

breerly:
I'm using this to test some config values for yarpc-go, currently copy pasting a `intp()` func everywhere.


abhinav:
Closing as discussed offline.

PR #154, opened 2016-06-15
enums: Don't generate const decl for empty enums

abhinav:
This fixes a `const ()` declaration being generated for empty enums.

@prashantv

### DIFF
--- a/gen/testdata/enums/types.go
+++ b/gen/testdata/enums/types.go
@@ -10,8 +10,6 @@ import (
 
 type EmptyEnum int32
 
-const ()
-
 func (v EmptyEnum) ToWire() (wire.Value, error) {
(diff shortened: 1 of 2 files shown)


breerly:
test?

gen/testdata/enums/types.go
```diff
@@ -10,8 +10,6 @@ import (
 
 type EmptyEnum int32
 
-const ()
```
breerly: fair enough

breerly:
🚢

abhinav:
The updated generated file is the test. (We already had an empty enum test.) This doesn't break compilation in any way.

prashantv:
lgtm

PR #153, opened 2016-06-15
Gofmt with simplify flag

sectioneight:
This follows the Uber standard, and generally makes tests easier to read.

NOTE: I ran this across the entire codebase then did a `git add` of files that **did not** look autogenerated. I may have accidentally committed an autogenerated file, please let me know.

/r @abhinav @prashantv

PR #155, opened 2016-06-16
Make: Add lint target and run it from CI

abhinav:
Adding linting to CI as per #153.

Lint needs to be skipped for Go 1.5 because the rules have been changed/improved in 1.6 and we develop against that. zap and tchannel-go already do the same.

CC @prashantv @breerly


blampe:
gross 🚢

breerly:
:shipit: (I hereby declare abg graduated from makefile college)

prashantv:
I'm sorry you had to write that. 🚢

PR #156, opened 2016-06-16
Make: Second cut -b3- not required

abhinav:
`$(GO_FILES)` already strips the `./` at the start of every path.
Doing it again in `$(GENERATED_GO_FILES)` will end up dropping two more characters from the path.


blampe:
🚢

PR #158, opened 2016-06-22
gen/importer: Support symbols in package basename

abhinav:
This is needed to allow the generated code for yarpc to import `github.com/yarpc/yarpc-go`.

### DIFF
--- a/gen/imports.go
+++ b/gen/imports.go
@@ -24,6 +24,8 @@ import (
 	"go/ast"
 	"go/token"
 	"path/filepath"
+	"strings"
+	"unicode"
 )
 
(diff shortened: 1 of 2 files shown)


breerly:
Can you add a description explaining why we need this or at least a test showing the behavior from the outside?

abhinav:
True. Updated description. Tests incoming.

abhinav:
Updated

breerly:
lgtm 🚢 

(btw, you don't need to add a comment saying "Updated" anymore, Github started sending email updates for additional commits pushed. I got one for "tests++" which was probably enough in this case)

prashantv:
lgtm

PR #159, opened 2016-06-22
yarpc: Use top-level {Req,Res}Meta* objects

abhinav:
This is the code generation change for https://github.com/yarpc/yarpc-go/pull/230


breerly:
:shipit:

PR #164, opened 2016-07-12
binary: Handle invalid strings with large lengths

abhinav:
Currently, we always pre-allocate the buffer to read strings into. This means
that an invalid request with a really large string length can simply lock the
system up. For example, `0x22, 0x6e, 0x6f, 0x74` will request a string 500+
MB long. Even if that many bytes don't follow, we'll end up allocating that
much space.

This changes the system to use a dynamically resizing buffer if the requested
length is larger than 1 MB. This will cause invalid requests to fail gracefully
because that much data will not be available over the wire.

### DIFF
--- a/protocol/binary/reader.go
+++ b/protocol/binary/reader.go
@@ -21,11 +21,16 @@
 package binary
 
 import (
+	"bytes"
 	"io"
 	"math"
 
 	"github.com/thriftrw/thriftrw-go/wire"
 )
+
+// Requests for byte slices longer than this will use a dynamically resizing
+// buffer.
+const bytesAllocThreshold = 1048576 // 1 MB
 
 // Reader implements a parser for the Thrift Binary Protocol based on an
 // io.ReaderAt.
@@ -202,6 +207,19 @@ func (br *Reader) read(bs []byte, off int64) (int64, error) {
 	return off, err
 }
 
+// copyN copies the given number of bytes starting at the given position into
+// the Writer.
+func (br *Reader) copyN(w io.Writer, off int64, n int64) (int64, error) {
+	src := io.NewSectionReader(br.reader, off, n)
+	copied, err := io.CopyN(w, src, n)
+	off += copied
+	if err == io.EOF {
+		// All EOFs are unexpected for the decoder
+		err = io.ErrUnexpectedEOF
+	}
+	return off, err
+}
+
 func (br *Reader) readByte(off int64) (byte, int64, error) {
 	bs := br.buffer[0:1]
 	off, err := br.read(bs, off)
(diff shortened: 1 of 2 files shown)


abhinav:
CC @prashantv @breerly @bombela @kriskowal

protocol/binary_test.go
```diff
@@ -284,6 +285,18 @@ func TestBinary(t *testing.T) {
...
+	// 5 MB + 4 bytes for length
+	data := make([]byte, 5242880+4)
+	data[0], data[1], data[2], data[3] = 0x0, 0x50, 0x0, 0x0 // 5 MB
+
+	value, err := Binary.Decode(bytes.NewReader(data), wire.TBinary)
+	require.NoError(t, err, "failed to decode value")
+
+	want := wire.NewValueBinary(data[4:])
+	assert.True(t, wire.ValuesAreEqual(want, value), "values did not match")
+}
```
prashantv: this test will pass with or without this change right? i would have preferred a test where we have a small buffer but where the length is invalid (max size). However with 4 bytes, the length could be 4gb which may succeed with a make.

prashantv:
lgtm, 👍 for fixing this. wish we could test this better, but unless there's something trivial, don't waste time on it.

protocol/binary_test.go
abhinav: The other test tests for invalid really-large length. I couldn't find a way to limit the amount of memory the test is allowed to allocate, so it'll still pass without this change.

protocol/binary_test.go
```diff
@@ -297,6 +310,7 @@ func TestBinaryEOFFailure(t *testing.T) {
 		{},
 		{0x00}, // incomplete length
 		{0x00, 0x00, 0x00, 0x01}, // length mismatch
+		{0x22, 0x6e, 0x6f, 0x74}, // really long length
```
breerly: Suggest moving comments into `desc` var that can be used during failure.

protocol/binary/reader.go
```diff
@@ -202,6 +207,19 @@ func (br *Reader) read(bs []byte, off int64) (int64, error) {
 	return off, err
 }
 
+// copyN copies the given number of bytes starting at the given position into
+// the Writer.
+func (br *Reader) copyN(w io.Writer, off int64, n int64) (int64, error) {
```
breerly: I think we can name this better... `copyIntoWriter` perhaps?

breerly: Wow... just saw `io.CopyN`... Ok - I'm def wrong here :)

breerly: I think the comment on this function confused me. Suggest taking `io.CopyN`'s lead with something like:

```
// CopyN copies n bytes (or until an error) from ...
```

breerly:
lgtm (with nits)

protocol/binary/reader.go
abhinav: 💯

protocol/binary_test.go
abhinav: That will involve changing all `checkEOFError`-based tests. I would prefer to hold off on this change until this becomes an actual issue. Usually there are only 1-2 cases for each test based on that.

protocol/binary/reader.go
abhinav: Forgot to push the change into this PR. Fix: https://github.com/thriftrw/thriftrw-go/pull/165

blampe:
This is really great, how did you find this?

abhinav:
A yarpc test was timing out when it received an invalid payload.

PR #173, opened 2016-08-22
TApplicationException: Move to top-level internal package

abhinav:
This moves TApplicationException to the top-level internal package so that
it's accessible from more thriftrw packages. Specifically, this is needed for
the plugin system because we use Thrift envelopes to talk to the plugins.

@breerly @kriskowal @prashantv


prashantv:
lgtm

PR #174, opened 2016-08-22
internal: Add MultiErr type

abhinav:
This is used to collect and present multiple errors together.

CC @breerly @kriskowal @prashantv


breerly:
Clean. :shipit:

PR #175, opened 2016-08-22
internal: Implement concurrent.Range

abhinav:
This uses reflection to abstract away the common pattern where you need to run
a function concurrently on all elements on a list and wait for them all to
finish, any of the operations optionally failing with an error.

The following,

```go
err := concurrent.Range(xs, func(x ..) error {
    // do stuff with x
    return err
})
```

Is equivalent to,

```go
var (
    wg sync.WaitGroup
    lock sync.Mutex
    errors []error
)

for _, x := range xs {
    wg.Add(1)
    go func(x ...) {
        defer wg.Done()
        err := // do stuff with x
        if err != nil {
            lock.Lock()
            errors = append(errors, err)
            lock.Unlock()
        }
    }(x)
}

wg.Wait()
err = MultiError(errors)
```

CC @breerly @kriskowal @prashantv

### DIFF
--- a/internal/concurrent/range.go
+++ b/internal/concurrent/range.go
@@ -1,0 +1,148 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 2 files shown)


internal/concurrent/range.go
```diff
@@ -0,0 +1,148 @@
...
+
+// Range calls the function f on all items in c concurrently and waits for all
+// calls to finish.
+//
+// c may be a slice or a map. If c is a map, f must accept the key and the value
+// as its arguments, and otherwise it must accept an int index and the value as
+// its arguments.
+//
+// f may return nothing or error.
+func Range(c, f interface{}) error {
```
breerly: Suggest renaming `f` to `fn` or `func` - I think `f` is abbreviated too aggressively.

breerly: Also, what does `c` stand for? I spent almost a minute trying to guess - probably an indicator that the variable name is not clear enough.

breerly:
:shipit: besides naming nits.

This looks pretty useful, we should consider putting this in it's own lib at some point.

PR #176, opened 2016-08-23
internal: Add frame reader, writer, client & server

abhinav:
These allow sending and receiving framed requests over arbitrary
`io.{Reader,Writer}`s. Each request is prefixed with a 4-byte big-endian
integer containing the length of the rest of the frame. This is compatible
with Apache Thrift's `TFramedTransport`. ThriftRW plugins will use framed
requests and responses to talk to the ThriftRW process.

@breerly @kriskowal @prashantv


internal/frame/reader.go
```diff
@@ -0,0 +1,91 @@
...
+	if length < _fastPathFrameSize {
+		return r.readFastPath(length)
+	}
+
+	var buff bytes.Buffer
+	copied, err := io.CopyN(&buff, r.r, length)
+	if err != nil {
+		return nil, err
+	}
+	if copied < length {
```
prashantv: think you can skip this check,

From the godoc: https://golang.org/pkg/io/#CopyN
```
On return, written == n if and only if err == nil.
```

It will return EOF if there was less data:
https://golang.org/src/io/io.go?s=11645:11715#L325

internal/frame/reader.go
```diff
@@ -0,0 +1,91 @@
...
+		return nil, fmt.Errorf("unexpected end of input: expected %v bytes but got %v", length, copied)
+	}
+
+	return buff.Bytes(), nil
+}
+
+func (r *Reader) readFastPath(l int64) ([]byte, error) {
+	buff := make([]byte, l)
+	if l == 0 {
+		return buff, nil
```
prashantv: is returning `nil` for 0 length not OK?

internal/frame/reader.go
abhinav: 👍

internal/frame/reader.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	_, err := io.ReadFull(r.r, buff)
+	return buff, err
+}
+
+// Close closes the given Reader.
+func (r *Reader) Close() error {
+	if c, ok := r.r.(io.ReadCloser); ok {
```
prashantv: do you need `ReadCloser` or is just `Closer` fine?

internal/frame/server.go
```diff
@@ -0,0 +1,91 @@
...
+
+	for s.running.Load() {
+		req, err := s.r.Read()
+		if err != nil {
+			return err
+		}
+
+		res, err := h.Handle(req)
+		if err != nil {
+			return err
```
prashantv: if there's an error, should the writer be be closed? Otherwise the client reader will be blocked in the Read trying to read the response I think (can we write a test for this)

internal/frame/writer.go
```diff
@@ -0,0 +1,70 @@
...
+func NewWriter(w io.Writer) *Writer {
+	return &Writer{w: w}
+}
+
+// Write writes the given frame to the Writer.
+func (w *Writer) Write(b []byte) error {
+	w.Lock()
+	defer w.Unlock()
+
+	// TODO(abg): Bounds check?
```
prashantv: I don't think we need a bounds check here since it's a constant buffer size

internal/frame/writer.go
```diff
@@ -0,0 +1,70 @@
...
+		return nil
+	}
+
+	_, err := w.w.Write(b)
+	return err
+}
+
+// Close closes the given Writeer.
+func (w *Writer) Close() error {
+	if c, ok := w.w.(io.WriteCloser); ok {
```
prashantv: can probably use `io.Closer` here too

internal/frame/reader.go
abhinav: Discussed offline. It *should* be okay but it makes tests really annoying because you can no longer do `assert.Equal`.

internal/frame/reader.go
abhinav: Good call. Closer should be fine.

internal/frame/writer.go
abhinav: I meant bounds check for `len(b)`. Although it's unlikely, if `b` is larger than `math.MaxInt32`, we'll have undefined behavior.

internal/frame/writer.go
abhinav: 👍

kriskowal:
Of course you have to roll your own length-prefixed framing…

…

lgtm

internal/frame/server.go
abhinav: True. Will close for both, read errors and handler errors since they both terminate the Serve loop.

prashantv:
lgtm

PR #180, opened 2016-08-23
internal: Add `envelope.{Client, Server}`

abhinav:
`envelope.{Client, Server}` rely on primitive transport interfaces
(implemented by `internal/frame` in #176) to send and receive enveloped Thrift
requests and responses. Out-of-order of requests is not supported.

@breerly @prashantv @kriskowal


internal/envelope/mock_client_test.go
```diff
@@ -0,0 +1,73 @@
+// Automatically generated by MockGen. DO NOT EDIT!
```
breerly: Github doesn't know how to collapse deez? That's a shame.

internal/envelope/client.go
```diff
@@ -0,0 +1,104 @@
...
+
+	default:
+		return wire.Value{}, errUnknownEnvelopeType(resEnvelope.Type)
+	}
+}
+
+type errUnknownEnvelopeType wire.EnvelopeType
+
+func (e errUnknownEnvelopeType) Error() string {
+	return fmt.Sprintf("unknown envelope type: expected Reply, got %v", wire.EnvelopeType(e))
```
breerly: Perhaps "expected `wire.Reply`: ...` here would be more obvious?

breerly:
lgtm (thanks for breaking deez up)

prashantv:
lgtm

PR #182, opened 2016-08-24
internal: Add process.Client

abhinav:
`process.Client` provides an API similar to `frame.Client`. It sends framed
requests and receives frames responses from an external process. This is the
underlying transport we will use to talk to plugins.

### DIFF
--- a/internal/process/client.go
+++ b/internal/process/client.go
@@ -1,0 +1,99 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


internal/process/client.go
```diff
@@ -0,0 +1,99 @@
...
+
+	"github.com/thriftrw/thriftrw-go/internal"
+	"github.com/thriftrw/thriftrw-go/internal/frame"
+	"github.com/uber-go/atomic"
+)
+
+// Client sends framed requests and receives framed responses from an external
+// process.
+type Client struct {
+	running *atomic.Bool
```
prashantv: can just use the value `atomic.Bool` (no pointer needed) and then you don't need to initialize it

abhinav: Discussed offline. Need to initialize to true so zero-value is undesirable.

prashantv:
lgtm

PR #183, opened 2016-08-24
gen/containers: Gracefully handle nil items

abhinav:
It's incorrect to pass nil as an item in a list, set or map (whether as a key
or as a value). Thrift does not support nil values inside containers.

Previously, we would simply panic if a user tried to do this. (The system
would call `ToWire` on the nill value and panic.) With this change, we instead
send an error back to the user indicating that nil values are not allowed.

### DIFF
--- a/gen/container_test.go
+++ b/gen/container_test.go
@@ -27,6 +27,8 @@ import (
 	te "github.com/thriftrw/thriftrw-go/gen/testdata/enums"
 	ts "github.com/thriftrw/thriftrw-go/gen/testdata/structs"
 	"github.com/thriftrw/thriftrw-go/wire"
+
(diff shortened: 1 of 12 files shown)


breerly:
Why not just panic? How would you recover from this?

abhinav:
You wouldn't recover from this but a panic wouldn't result in a very informative error for this. Since this is technically a validation error similar to the "field is required" errors on structs, I think it's reasonable to bubble up an error message to the user.

breerly:
🚢

kriskowal:
I’m fine with turning this into an informative runtime error. Reduces the potential for interference between old well-behaved callers and new ill-behaved callers.

lgtm

prashantv:
lgtm

PR #184, opened 2016-08-24
plugin: Add IDL definining the core API

abhinav:
This defines the underlying Thrift API that ThriftRW will use to talk to
plugins. Plugin authors will be provided a nicer API on top of this.

@prashantv @kriskowal @breerly

PR #185, opened 2016-08-24
plugin: Add API for plugin authors

abhinav:
Plugin authors don't need to know how the plugin is talking to ThriftRW. With
this API, they can simply implement a `main` function that calls `plugin.Main`
with their plugin definition, and we handle setting up the communication with
ThriftRw.

Depends on #184.

@prashantv @kriskowal @breerly


abhinav:
CC @anuptalwalkar

prashantv:
lgtm

PR #186, opened 2016-08-24
plugin: Add templating system

abhinav:
Plugin authors will be able to use this to generate code using a templating
system similar to what ThriftRW uses internally. This is significantly simpler
than what ThriftRW uses internally but should suffice for most use cases.

Depends on #184.

@prashantv @kriskowal @breerly

### DIFF
--- a/plugin/template.go
+++ b/plugin/template.go
@@ -52,7 +52,7 @@ type TemplateOption struct {
 // 	)
 func TemplateFunc(name string, f interface{}) TemplateOption {
 	return TemplateOption{apply: func(t *goFileGenerator) {
-		t.funcs[name] = f
+		t.templateFuncs[name] = f
 	}}
 }
 
@@ -76,8 +76,8 @@ func GoFileImportPath(path string) TemplateOption {
 
 // goFileGenerator generates a single Go file.
 type goFileGenerator struct {
-	importPath string
-	funcs      map[string]interface{}
+	importPath    string
+	templateFuncs template.FuncMap
 
 	// Names of known globals. All global variables share this namespace.
 	globals map[string]struct{}
@@ -91,7 +91,7 @@ type goFileGenerator struct {
 
 func newGoFileGenerator(opts []TemplateOption) *goFileGenerator {
 	t := goFileGenerator{
-		funcs:         make(map[string]interface{}),
+		templateFuncs: make(template.FuncMap),
 		globals:       make(map[string]struct{}),
 		imports:       make(map[string]string),
 		importedNames: make(map[string]string),
@@ -207,7 +207,7 @@ func (g *goFileGenerator) Generate(filename, tmpl string, data interface{}) ([]b
 		"import":     g.Import,
 		"formatType": g.FormatType,
 	}
-	for k, v := range g.funcs {
+	for k, v := range g.templateFuncs {
 		funcs[k] = v
 	}
 
--- a/plugin/template.go
+++ b/plugin/template.go
@@ -56,12 +56,18 @@ func TemplateFunc(name string, f interface{}) TemplateOption {
 	}}
 }
 
-// GoFileImportPath is a TemplateOption that specifies the intended import path
-// for the file being generated.
+// GoFileImportPath is a TemplateOption that specifies the intended absolute
+// import path for the file being generated.
 //
-// If specified, this changes the behavior of the `formatType` template function
-// to NOT import this package and instead use the types directly since they are
-// available in the same package.
+// 	GoFileFromTemplate(
+// 		filename,
+// 		mytemplate,
+// 		GoFileImportPath("github.com/thriftrw/thriftrw-go/myservice"),
+// 	)
+//
+// If specified, this changes the behavior of the `formatType` template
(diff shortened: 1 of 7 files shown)


abhinav:
CC @anuptalwalkar

plugin/template.go
```diff
@@ -0,0 +1,284 @@
...
+// 		`package <lower "HELLO">`,
+// 		TemplateFunc("lower", strings.ToLower),
+// 	)
+func TemplateFunc(name string, f interface{}) TemplateOption {
+	return TemplateOption{apply: func(t *goFileGenerator) {
+		t.funcs[name] = f
+	}}
+}
+
+// GoFileImportPath is a TemplateOption that specifies the intended import path
```
prashantv: are these import paths absolute?

plugin/template.go
```diff
@@ -0,0 +1,284 @@
...
+func GoFileImportPath(path string) TemplateOption {
+	return TemplateOption{apply: func(t *goFileGenerator) {
+		t.importPath = path
+	}}
+}
+
+// goFileGenerator generates a single Go file.
+type goFileGenerator struct {
+	importPath string
+	funcs      map[string]interface{}
```
prashantv: can you specify that this is template functions, since it's in a struct with `globals` and `imports`, it's easy to mistake this as a list of functions in the file

plugin/template.go
abhinav: Yep. Will clarify in docs.

plugin/template.go
abhinav: Switched data type to `template.FuncMap` and renamed to `templateFuncs`.

plugin/template.go
```diff
@@ -0,0 +1,284 @@
...
+
+// Import the given import path and return the imported name for this package.
+func (g *goFileGenerator) Import(path string) string {
+	if name, ok := g.importedNames[path]; ok {
+		return name
+	}
+
+	name := goast.DeterminePackageName(path)
+
+	// Find an import name that does not conflict with any known globals.
```
prashantv: We should also make sure that the package name is not a reserved keyword or type. E.g. this is totally valid:

$GOPATH/src/foo/bool/bool.go:
```go
package bool

func Test() {
  println("test")
}
```

$GOPATH/src/foo/main.go:
```go
package main

import "foo/bool"

func main() {
  bool.Test()
}
```

plugin/template.go
```diff
@@ -0,0 +1,284 @@
...
+		default:
+			return "", fmt.Errorf("unknown simple type: %v", *t.SimpleType)
+		}
+	case t.SliceType != nil:
+		v, err := g.FormatType(t.SliceType)
+		if err != nil {
+			return "", err
+		}
+
+		return "[]" + v, nil
```
prashantv: you can do the same as below here,
```go
v, err := g.FormatType(t.SliceType)
return "[]" + v, err
```

plugin/template.go
```diff
@@ -0,0 +1,284 @@
...
+// goFileGenerator generates a single Go file.
+type goFileGenerator struct {
+	importPath string
+	funcs      map[string]interface{}
+
+	// Names of known globals. All global variables share this namespace.
+	globals map[string]struct{}
+
+	// import path -> empty string or import name for named imports
+	imports map[string]string
```
prashantv: Will there be some way for the user to come up with an unexported function name that won't clash with imports/globals?

plugin/template.go
abhinav: 👍

plugin/template.go
abhinav: We could add that, sure, although I think for most cases people will be able to get away with `_$pluginName_$functionName`. Adding things like `newVar`, etc. isn't very safe or always guaranteed to work because we've already evaluated the full template by the time we know what globals are being declared besides `import`.

To get a sort-of guarantee on globals, we would have to require that all globals be declared through functions. Like,

```
<$foo := import "...">

<func "myfunc">(....) {
  ..
}

<$bar := import ".../myfunc">  // this gets renamed
```

That's the only way to guarantee that there are no name clashes as the template is being evaluated.

plugin/template.go
abhinav: 👍

plugin/template.go
prashantv: Fair enough, 👍

plugin/template_test.go
```diff
@@ -0,0 +1,201 @@
...
+				package bar
+
+				func hello() <formatType .> {
+					return nil
+				}
+			`,
+			data: &api.Type{
+				ReferenceType: &api.TypeReference{
+					Name:    "Foo",
+					Package: "github.com/thriftrw/thriftrw-go/hello",
```
prashantv: nit: here and below, can we make this package `github.com/thriftrw/thriftrw-go/bar` to match the package name in the template. it doesn't really matter but it's just less confusing that way

prashantv:
lgtm

plugin/template_test.go
abhinav: 👍

PR #188, opened 2016-08-25
Fix cover.sh: in() not present in older jq versions

abhinav:
The version of jq on our Travis containers doesn't support the `in` function.
This changes it to use a simple map lookup.


prashantv:
lgtm

PR #190, opened 2016-08-26
internal: Add Plugin client

abhinav:
This adds a cleaner client API on top of generated Thrift APIs. This will be
used by the code generator to talk to plugins.


prashantv:
lgtm

breerly:
:shipit: (nice tests sir)

internal/plugin/transport.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/thriftrw/thriftrw-go/internal/multiplex"
+	"github.com/thriftrw/thriftrw-go/plugin/api"
+	"github.com/thriftrw/thriftrw-go/plugin/api/service/plugin"
+	"github.com/thriftrw/thriftrw-go/plugin/api/service/servicegenerator"
+	"github.com/thriftrw/thriftrw-go/protocol"
+
+	"github.com/uber-go/atomic"
+)
+
+var _proto = protocol.Binary
```
kriskowal: What’s the idea behind this underscore?

kriskowal:
lgtm

internal/plugin/transport.go
abhinav: Just a convention for variables that should be treated as local to the current file rather than the package.

PR #192, opened 2016-08-27
Switch to go-flags

abhinav:
We need this so that we can accept multiple `--plugin` arguments and have
hidden arguments for internal plugins.

This is technically a breaking change in that instead of `-no-recurse`,
`-yarpc`, etc., you have to use `--no-recurse`, `--yarpc`, etc.


breerly:
:shipit:

PR #193, opened 2016-08-29
plugin/server: Fix Stop() race condition

abhinav:
Calling `Stop` on the server was not closing the reader/writer so the loop was
stuck waiting for the next request. This changes it to close the reader and
let the writer be closed automatically when the loop exits.

Errors while reading are ignored if the server has already been closed.


prashantv:
lgtm

PR #194, opened 2016-08-29
plugins: Parse command line arguments

abhinav:
This adds support for parsing plugin arguments over the command line. It does
not actually integrate into the code generator yet.

Plugins are specified over the command line using the "-p" or "--plugin"
options. The value is in the form,

    NAME [ARGS ...]

Where `NAME` is the name of the plugin and `ARGS` is zero or more arguments to
pass to the plugin.

An executable with the name `thriftrw-plugin-${NAME}` is expected to be
available on `$PATH` to call the plugin. `ARGS` are passed to the plugin
executable.

For example,

    -p yarpc -p 'accessors --no-setters'

Will call,

-   `thrifrw-plugin-yarpc`
-   `thriftrw-plugin-accessors --no-setters`

Currently, a plugin can only be called once in a single invocation of
ThriftRW. That is, the following is currently unsupported:

    -p "foo --someflag" -p "foo --anotherflag"

### DIFF
--- a/gen/generate.go
+++ b/gen/generate.go
@@ -30,6 +30,7 @@ import (
 	"strings"
 
 	"github.com/thriftrw/thriftrw-go/compile"
+	"github.com/thriftrw/thriftrw-go/internal/plugin"
 )
 
 // Options controls how code gets generated.
@@ -56,6 +57,9 @@ type Options struct {
 	// NoRecurse determines whether code should be generated for included Thrift
 	// files as well. If true, code gets generated only for the first module.
 	NoRecurse bool
+
+	// Code generation plugin
+	Plugin plugin.Handle
 }
 
 // Generate generates code based on the given options.
@@ -87,6 +91,8 @@ func Generate(m *compile.Module, o *Options) error {
 		}
 		return nil
 	})
+
+	// TODO(abg): Generate code from opts.Plugin
 }
 
 // TODO(abg): Make some sort of public interface out of the Importer
--- a/internal/plugin/flag.go
+++ b/internal/plugin/flag.go
@@ -62,7 +62,10 @@ func (f *Flag) Handle() (Handle, error) {
 
 	handle, err := NewTransportHandle(f.Name, transport)
 	if err != nil {
-		return nil, fmt.Errorf("failed to open plugin %q: %v", f.Name, err)
+		return nil, internal.CombineErrors(
+			fmt.Errorf("failed to open plugin %q: %v", f.Name, err),
+			transport.Close(),
+		)
 	}
 
 	return handle, nil
@@ -70,7 +73,7 @@ func (f *Flag) Handle() (Handle, error) {
 
 // UnmarshalFlag parses a string specification of a plugin.
 func (f *Flag) UnmarshalFlag(value string) error {
-	tokens, err := shlex.Split(value, true)
+	tokens, err := shlex.Split(value, true /* posix */)
 	if err != nil {
 		return fmt.Errorf("invalid plugin %q: %v", value, err)
 	}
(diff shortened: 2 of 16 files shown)


gen/generate.go
```diff
@@ -87,6 +91,8 @@ func Generate(m *compile.Module, o *Options) error {
 		}
 		return nil
 	})
+
+	// TODO(abg): Generate code from opts.Plugin
```
breerly: Perhaps add a TODO to remove YARPC option

abhinav: It's there in the relevant part of main.go :)

internal/plugin/flag.go
```diff
@@ -0,0 +1,143 @@
...
+// The returned handle MUST be closed by the caller if error was nil.
+func (f *Flag) Handle() (Handle, error) {
+	transport, err := process.NewClient(f.Command)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open plugin %q: %v", f.Name, err)
+	}
+
+	handle, err := NewTransportHandle(f.Name, transport)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open plugin %q: %v", f.Name, err)
```
prashantv: do we need to close `transport` in this case?

internal/plugin/flag.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	return handle, nil
+}
+
+// UnmarshalFlag parses a string specification of a plugin.
+func (f *Flag) UnmarshalFlag(value string) error {
+	tokens, err := shlex.Split(value, true)
```
prashantv: can you put a comment after `true` that says what the arg name ism like `true /* posix */`

main.go
```diff
@@ -104,14 +111,23 @@ func main() {
 		}
 	}
 
+	pluginHandle, err := opts.Plugins.Handle()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if pluginHandle != nil {
+		defer pluginHandle.Close()
```
prashantv: can you just always `defer pluginHandle.Close()` and handle the `nil` case in the custom type.

internal/plugin/flag_test.go
```diff
@@ -0,0 +1,286 @@
...
+		var f Flag
+		err := f.UnmarshalFlag(tt.giveValue)
+		if tt.wantErrorPrefix != "" {
+			if assert.Error(t, err, "expected error for %q", tt.giveValue) {
+				assert.True(t,
+					strings.HasPrefix(err.Error(), tt.wantErrorPrefix),
+					"expected prefix %q on error message %q for %q",
+					tt.wantErrorPrefix, err.Error(), tt.giveValue,
+				)
+			}
```
prashantv: add `continue` and drop the `else`

internal/plugin/flag_test.go
```diff
@@ -0,0 +1,286 @@
...
+		}
+
+		if len(tt.wantErrors) > 0 {
+			if !assert.Error(t, err, "%v: expected error", tt.desc) {
+				continue
+			}
+
+			for _, msg := range tt.wantErrors {
+				assert.Contains(t, err.Error(), msg, "%v: error message mismatch", tt.desc)
+			}
```
prashantv: suggest adding `continue` and dropping `else` here as well.

prashantv:
lgtm with minor comment. We should not need the caller to do `nil` checks after getting a `MultiHandle`. Instead we should implement a `Close` that handles `nil`

internal/plugin/flag.go
abhinav: true

internal/plugin/flag.go
abhinav: 👍

main.go
abhinav: 👍

internal/plugin/flag_test.go
abhinav: reduced nesting

internal/plugin/flag_test.go
abhinav: 👍

abhinav:
Updated

prashantv:
lgtm

PR #195, opened 2016-08-30
Expose annotations

mihai-chiorean:
Exposed annotations for Felds, Services, Enums, Enum items, Functions, Structs

Type/Typedefs/primitiveType need a lot more changes since it would require changing the signature of some of the compileXXXX functions


blampe:
Hey @mihai-chiorean, can you describe the use case you had in mind here? In particular do you imagine it would be OK for another language (like Python or Java) to consume one of these Thrift files *without* getting these annotations exposed to the app?

mihai-chiorean:
I don't expect the other languages to fail if annotations are present in the file, if that's what you mean. My use case here is for a tool that parses a bunch of thrift files and builds some access policies based on annotations.

abhinav:
Hey, thanks for implementing this. I'll have a closer look soon, but general
thoughts on design for this (your current code already satisfies most of these
requirements):

----

We can have annotations on,

-   Type declarations

    ```
    struct Foo { ... } (foo, bar = "baz")
    
    typedef i64 Timestamp (a = "b")
    ```
    
    These should be attached to the respective `TypeSpec`s. Perhaps the
    `TypeSpec` interface should have an `Annotations()` function. This should
    work fine for typedefs too.

-   Enum items

    ```
    enum Status {
        DISABLED = 1 (go.constName = "Disabled"),
        ENABLED (go.constName = "Enabled"),
    }
    ```
    
    Attach to `EnumItem`s.
    
-   Struct fields

    ```
    struct User {
        1: required UUID id
        2: required string name (visibility = "private")
    }
    ```
    
    Attach to `FieldSpec`s.

-   Service declarations

    ```
    service Foo { .. } (access = "private")
    ```
    
    Attach to `ServiceSpec`s.
    
-   Service functions

    ```
    service KeyValue {
        void getValue(1: string key)
            throws (1: KeyDoesNotExist doesNotExist)
            (idempotent, maxAttempts = "3")
    }
    ```
    
    Attach to `FunctionSpec`.

-   ALL references to native types, including lists, sets, map and the
    references made by them.

    ```
    typedef i64 (js.type = "Long") Timestamp
    // This is different from
    typedef i64 Timestamp (js.type = "Long")
    
    struct Foo {
        1: required map<string (obfuscate), i64> (js.type = "array") values
    }
    ```
    
    Note that you CAN'T have annotations on references to user-defined types.
    
    ```
    struct Foo {
        1: required Timestamp (js.type = "Buffer") wrong // This is wrong
    }
    ```
    
    For containers, these can be attached to `{List, Set, Map}Spec`. For
    primitives, we'll first have to stop using singletons. Instead, we will
    need to declare a type for each primitive.
    
    ```go
    type BoolSpec struct {
        nativeThriftType
        
        Annotations []*AnnotationSpec
    }
    
    func (BoolSpec) TypeCode() wire.Type { return wire.TBool }
    func (BoolSpec) ThriftName() string  { return "bool"     }
    // ...
    ```
    
    If you're going to make the change, I recommend creating a separate pull
    request for just converting primitives to types (without annotations)
    since that will involve a lot of noise in just test updates, etc.

compile/annotation.go
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
sectioneight: Year is wrong

compile/service_test.go
```diff
@@ -60,53 +60,98 @@ func TestCompileService(t *testing.T) {
 	}
 
 	keyValueSpec := &ServiceSpec{
-		Name: "KeyValue",
-		File: "test.thrift",
+		Name:        "KeyValue",
+		File:        "test.thrift",
+		Annotations: []*AnnotationSpec{},
```
sectioneight: `nil` is a valid slice, why do you need to change all these?

mihai-chiorean:
I think at this point the only 2 things that are missing are,

> These should be attached to the respective TypeSpecs. Perhaps the
TypeSpec interface should have an Annotations() function. This should
work fine for typedefs too.

 and

> "ALL references to native types, including lists, sets, map and the
references made by them.",

Which I did not add to this PR because it would have made it too large. I want to make a new PR about those since it means a lot of noise in tests and some of the compile functions need to have a different signature that also returns an error

compile/service_test.go
mihai-chiorean: the Annotations slice is added even if empty; so the tests fail when comparing []*AnnotationSpec{} with []*AnnotationSpec(nil).

I could potentially not add the slice if len == 0

compile/annotation.go
mihai-chiorean: lol. I'll fix that

abhinav: @mihai-chiorean You should be able to use `scripts/updateLicenses.sh` to handle this.

compile/annotation.go
```diff
@@ -0,0 +1,52 @@
...
+
+// AnnotationSpec describes and a thrift annotation
+type AnnotationSpec struct {
+	Name  string
+	Value string
+}
+
+func compileAnnotations(annotations []*ast.Annotation) ([]*AnnotationSpec, error) {
+	namespace := newNamespace(caseInsensitive)
+	annotationSpecs := []*AnnotationSpec{}
```
abhinav: Use `var annotationSpecs []*AnnotationSpec` instead. It'll be nil when empty. This should make it so you don't have to change most of the existing tests.

blampe:
@mihai-chiorean thanks for clarifying my earlier answer, this makes a lot of sense. Thank you for the PR :)

I'd like to make sure that (internally) we have some clear guidance around how to name and use these annotations -- I'll start a separate conversation with you and @sectioneight for that.

compile/annotation_test.go
```diff
@@ -0,0 +1,108 @@
...
+package compile
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCompileAnnotation(t *testing.T) {
```
abhinav: These tests should be testing compileAnnotation only. You don't need a ServiceSpec to test `compileAnnotations`. All the stuff with `scope` and `Link` is unnecessary for this. Just hand-write a `[]*ast.Annotation` and call `compileAnnotations` on that.

compile/annotation_test.go
```diff
@@ -0,0 +1,108 @@
...
+		if assert.NoError(t, err, tt.desc) {
+			if assert.NoError(t, spec.Link(scope), tt.desc) {
+				assert.Equal(t, tt.spec, spec, tt.desc)
+			}
+		}
+	}
+}
+
+func TestCompileAnnotationFailure(t *testing.T) {
+
```
abhinav: General styling nit: Don't add an empty line at the start of a function.

compile/annotation_test.go
mihai-chiorean: fair enough. I'll make that change

compile/enum.go
```diff
@@ -27,15 +27,17 @@ import (
 
 // EnumSpec represents an enum defined in the Thrift file.
 type EnumSpec struct {
-	Name  string
-	File  string
-	Items []EnumItem
+	Name        string
+	File        string
+	Items       []EnumItem
+	Annotations []*AnnotationSpec
```
abhinav: If you add `Annotations()` to the TypeSpec interface, you'll probably want to rename this to `Annots`.

compile/enum.go
```diff
@@ -52,19 +54,26 @@ func compileEnum(file string, src *ast.Enum) (*EnumSpec, error) {
...
-
 		value := prev + 1
 		if astItem.Value != nil {
 			value = *astItem.Value
 		}
 		prev = value
 
+		itemAnnotations, err := compileAnnotations(astItem.Annotations)
+		if err != nil {
+			return nil, err
```
abhinav: The intention of `compileError` is to give you a message on exactly which component failed. You should use `compileError` here and a generic `annotationConflictError{Reason: err}` (define in compile/error.go) in `compileAnnotations`. That will give us an error message like,

    cannot compile "MyEnum.SomeItem" on line 42: annotation conflict: the name "foo" has already been used on line 45

compile/service.go
```diff
@@ -62,11 +63,18 @@ func compileService(file string, src *ast.Service) (*ServiceSpec, error) {
 		functions[function.Name] = function
 	}
 
+	annotations, err := compileAnnotations(src.Annotations)
+
+	if err != nil {
+		return nil, err
```
abhinav: This should probably be a `compileError` too to include the service name in the error.

compile/annotation.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+// AnnotationSpec describes and a thrift annotation
+type AnnotationSpec struct {
+	Name  string
+	Value string
+}
+
+func compileAnnotations(annotations []*ast.Annotation) ([]*AnnotationSpec, error) {
+	namespace := newNamespace(caseInsensitive)
```
abhinav: I'm fairly sure annotations are case sensitive in Thrift but because they don't actually affect behavior, we can choose to do it differently. @prashantv @kriskowal Do you think we should make annotations case insensitive?

compile/annotation.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"github.com/thriftrw/thriftrw-go/ast"
+)
+
+// AnnotationSpec describes and a thrift annotation
+type AnnotationSpec struct {
+	Name  string
+	Value string
+}
```
abhinav: I'm guessing your use case (and the majority use case for people using the
compile package) will involve looking annotations up by name. Maybe we should
use `type Annotations map[string]string` instead of `[]*AnnotationSpec`.

Although this may be prone to errors if we treat annotation names as case
insensitive because it's really easy to construct `Annotations{"Foo": "bar"}`
with keys that are not normalized. Further, if we normalize keys, we're now
losing the exact key that the user used (which may be something we want access
to).

How about keeping annotations case sensitive? This is compliant with what
Thrift already does. You'll just have to special case `compileAnnotations` to
return nil when `annotations` is empty.  For convenience, `type Annotations`
can have a `Get` function which handles the case where the map is nil.

```go
type Annotations map[string]string

func (a Annotations) Get(key string) (string, ok) {
  if a == nil {
    return "", false
  }
  ...
}
```

mihai-chiorean: I followed the pattern that's on go-thrift. So in my use case I am iterating through the annotations and have a switch where it matches the annotations we handle. 

I see your point though. "range" works on maps too.

compile/enum.go
mihai-chiorean: oh nice! I like that. Thanks

compile/annotation.go
```diff
@@ -0,0 +1,61 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package compile
+
+import (
+	"fmt"
+	"github.com/thriftrw/thriftrw-go/ast"
```
prashantv: nit: add a blank line between the standard library packages and this import

compile/annotation.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+// AnnotationSpec describes and a thrift annotation
```
prashantv: I think this comment is missing something (or maybe has extra words) "descibes and a thrift annotation". Also suggest capitalizing Thrift.

compile/annotation.go
prashantv: I'm fine with keeping annotations case sensitive if it makes more sense while iterating over them etc.

compile/annotation_test.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/thriftrw/thriftrw-go/ast"
+)
+
+func TestCompileAnnotation(t *testing.T) {
+
```
prashantv: nit: remove blank

compile/service_test.go
```diff
@@ -142,6 +177,16 @@ func TestCompileService(t *testing.T) {
 			keyValueSpec,
 		},
 		{
+			"service annotations",
+			`
+				service AnnotatedService {
+					void setValue(1: string key, 2: binary value) (test = "ok")
+				} (test = "test")
```
prashantv: suggest having tests where the key and value are different so we're not accidentally swapping them anywhere.

compile/annotation.go
```diff
@@ -0,0 +1,60 @@
...
+	msg := fmt.Sprint("annotation conflict")
+	if e.Reason != nil {
+		msg += fmt.Sprintf(": %v", e.Reason)
+	}
+	return msg
+}
+
+func compileAnnotations(annotations []*ast.Annotation) (Annotations, error) {
+	namespace := newNamespace(caseSensitive)
+	annotationSpecs := make(Annotations)
```
abhinav: Since we know the number of elements, we can probably specify the size in advance

    make(Annotations, len(annotations))

abhinav: Also, nit: There's no `annotationSpecs` now so we should probably rename these variables.

compile/annotation.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/thriftrw/thriftrw-go/ast"
+)
+
+// Annotations maps annotations
+type Annotations map[string]string
+
+type annotationConflictError struct {
+	Reason error
+}
```
abhinav: nit: you should move this to compile/errors.go

compile/annotation.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (e annotationConflictError) Error() string {
+	msg := fmt.Sprint("annotation conflict")
+	if e.Reason != nil {
+		msg += fmt.Sprintf(": %v", e.Reason)
+	}
```
abhinav: The usage for this will probably always include a reason so this doesn't need to be a conditional. `compileError` needed to be more flexible because of the different ways it is used.

compile/annotation.go
```diff
@@ -0,0 +1,60 @@
...
+
+	for _, a := range annotations {
+		if err := namespace.claim(a.Name, a.Line); err != nil {
+			return nil, annotationConflictError{Reason: err}
+		}
+
+		annotationSpecs[a.Name] = a.Value
+	}
+
+	if len(annotationSpecs) == 0 {
```
abhinav: This can probably be done before the for loop by checking `len(annotations)` instead.

compile/annotation_test.go
```diff
@@ -0,0 +1,97 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package compile
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/thriftrw/thriftrw-go/ast"
```
abhinav: nit: import group ordering. We usually go for three import groups (in this order):

- standard library imports
- local project imports
- third party libraries

So this should be,

```
"testing"

"github.com/thriftrw/thriftrw-go/ast"

```

compile/annotation_test.go
```diff
@@ -0,0 +1,97 @@
...
+		{
+			"simple annotation",
+			[]*ast.Annotation{
+				{
+					Name:  "test",
+					Value: "ok",
+					Line:  1,
+				},
+			},
+			scope(),
```
abhinav: scope not necessary in test table. it's not used anywhere.

compile/annotation_test.go
```diff
@@ -0,0 +1,97 @@
...
+			annotatedSpec,
```
abhinav: Nit: I would inline `annotatedSpec` rather than using a variable here.

compile/enum.go
```diff
@@ -52,19 +54,34 @@ func compileEnum(file string, src *ast.Enum) (*EnumSpec, error) {
...
 		value := prev + 1
 		if astItem.Value != nil {
 			value = *astItem.Value
 		}
 		prev = value
 
+		itemAnnotations, err := compileAnnotations(astItem.Annotations)
+		if err != nil {
+			return nil, compileError{
+				Target: astItem.Name,
```
abhinav: This should be `src.Name + "." + astItem.Name` so that we have the enum name in addition to the item name.

compile/service.go
```diff
@@ -62,11 +63,22 @@ func compileService(file string, src *ast.Service) (*ServiceSpec, error) {
 		functions[function.Name] = function
 	}
 
+	annotations, err := compileAnnotations(src.Annotations)
+
+	if err != nil {
```
abhinav: Nit: Usually, error checks for things are right next to the statement that created the error. No empty line.

compile/service_test.go
```diff
@@ -280,6 +319,28 @@ func TestCompileServiceFailure(t *testing.T) {
...
+			`,
+			[]string{
+				`the name "test" has already been used`,
+			},
+		},
+		{
+			"duplicate annotation name on function",
+			`
+				service AnnotatedService {
+					i32 bar() (functest = "test" functest = "t")
```
abhinav: Although the grammar makes separators optional, you should probably add a comma between the two annotations.

abhinav:
This is great! Just a few nits but LGTM otherwise.

abhinav:
LGTM besides broken lint. Will merge after the tests pass.

PR #196, opened 2016-08-31
internal/plugin: Expose MockHandle in handletest package

abhinav:
We need access to MockHandle in other places.


breerly:
doh :shipit:

PR #197, opened 2016-08-31
Rename import paths to go.uber.org/thriftrw

breerly:
@abhinav


breerly:
Going to re-open against dev

PR #198, opened 2016-08-31
Revamp README for v0.1.0

breerly:
Switching thriftrw over to the release process used in tchannel-go. Once this merges into `dev` I'm going to cut `v0.1.0` so that yarpc can pin to it.

@abhinav @kriskowal @prashantv


abhinav:
👍

PR #200, opened 2016-08-31
plugin: Generate code from plugins

abhinav:
This integrates plugins into the code generator and starts actually generating
code. With this change, the plugin system is actually fully functional.

Note that we're not dropping support for the `--yarpc` flag until YARPC has cut
a release with the YARPC plugin.


breerly:
Should this be going into `dev`?

abhinav:
Ugh. Right. Ignore.

PR #202, opened 2016-08-31
plugin: Generate code from plugins

abhinav:
This integrates plugins into the code generator and starts actually generating
those files. With this change, the plugin system is now fully functional.

Note that we are not dropping support for the `--yarpc` flag until YARPC has
cut a release with the YARPC plugin.

### DIFF
--- a/gen/generate.go
+++ b/gen/generate.go
@@ -30,6 +30,7 @@ import (
 	"strings"
 
 	"github.com/thriftrw/thriftrw-go/compile"
+	"github.com/thriftrw/thriftrw-go/internal"
 	"github.com/thriftrw/thriftrw-go/internal/plugin"
 )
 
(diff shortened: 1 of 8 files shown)


abhinav:
```
gen/plugin_test.go:117: compile.ArgsSpec composite literal uses unkeyed fields
gen/plugin_test.go:126: compile.FieldGroup composite literal uses unkeyed fields
gen/plugin_test.go:134: compile.FieldGroup composite literal uses unkeyed fields
gen/plugin_test.go:175: compile.ArgsSpec composite literal uses unkeyed fields
gen/plugin_test.go:370: compile.FieldGroup composite literal uses unkeyed fields
gen/plugin_test.go:457: compile.FieldGroup composite literal uses unkeyed fields
```

Those errors are wrong. Those literals aren't structs but aliases of collections. Feel free to ignore them.

kriskowal:
lgtm

gen/generate.go
```diff
@@ -81,18 +82,61 @@ func Generate(m *compile.Module, o *Options) error {
 		ThriftRoot:   o.ThriftRoot,
 	}
 
-	if o.NoRecurse {
-		return generateModule(m, importer, o)
-	}
+	// Mapping of files relative to OutputDir to their contents.
```
prashantv: nit: s/files/filenames

gen/generate.go
```diff
@@ -150,105 +205,101 @@ func generateModule(m *compile.Module, i thriftPackageImporter, o *Options) erro
...
-		files["types.go"] = buff
+		files["types.go"] = buff.Bytes()
 	}
 
 	if len(m.Services) > 0 {
 		for _, serviceName := range sortStringKeys(m.Services) {
 			service := m.Services[serviceName]
+
+			// If we called generateModule on something, that's a root
+			// service. We need plugins to generate code for them.
```
prashantv: I don't understand the comment about if we called `generateModule` on something, it's a root service.

Doesn't `Generate` call `generateModule` on all modules?

gen/generate_test.go
```diff
@@ -21,16 +21,32 @@
...
+	"github.com/thriftrw/thriftrw-go/plugin/api"
+	"github.com/thriftrw/thriftrw-go/plugin/plugintest"
 
+	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+func testdata(t *testing.T, paths ...string) string {
+	cwd, err := os.Getwd()
```
prashantv: by default all paths are relative to `cwd` so do you need to append this or can you just append to `./testdata`

gen/generate.go
abhinav: It calls them on all modules by default, but only on one if `--no-recurse` was used. In that case, we want to generate code only for those services in that file, but we still need information about the services it depends on.

abhinav: I'll try to clarify further in comments.

gen/generate_test.go
abhinav: The path needs to be absolute for `Generate()`.

abhinav: Will explain in comment

gen/plugin.go
```diff
@@ -0,0 +1,348 @@
...
+}
+
+// AddRootService adds a service as a root service to this request.
+func (g *generateServiceBuilder) AddRootService(spec *compile.ServiceSpec) (api.ServiceID, error) {
+	id, err := g.addService(spec)
+	if err != nil {
+		return id, err
+	}
+
+	if _, alreadyAdded := g.rootServices[id]; !alreadyAdded {
```
prashantv: Is it possible for a root service to be added twice?

gen/generate.go
prashantv: 👍

gen/generate.go
abhinav: 👍

gen/plugin.go
abhinav: No. `RootServices` is just a list of pointers to services for which code should be generated. You won't generate code for the same service twice because it would end up overwriting itself.

prashantv:
lgtm

PR #205, opened 2016-09-02
plugin: Don't require unique names

abhinav:
This changes MultiHandle and MultiServiceGenerator to stop keying off of the
plugin name. An advantage of this is that plugin names are no longer required
to be unique so we don't have to worry about builtin plugins conflicting with
user plugins. Additionally, this also makes parameterized plugins possible.
That is, you could now have a plugin which generates significantly different
code based on its command line arguments and have it used twice in the same
thriftrw invocation:

    -p 'myplugin --generateA' -p 'myplugin --generateB'

### DIFF
--- a/internal/plugin/multi.go
+++ b/internal/plugin/multi.go
@@ -22,6 +22,7 @@ package plugin
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/thriftrw/thriftrw-go/internal/concurrent"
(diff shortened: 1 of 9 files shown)


internal/plugin/multi.go
```diff
@@ -22,45 +22,58 @@ package plugin
...
 // ServiceGenerators of all plugins associated with this MultiHandle and
 // consolidates their results.
-func (mh MultiHandle) ServiceGenerator() api.ServiceGenerator {
-	msg := make(MultiServiceGenerator, len(mh))
-	for name, h := range mh {
+func (mh MultiHandle) ServiceGenerator() ServiceGenerator {
+	msg := make(MultiServiceGenerator, 0, len(mh))
+	for _, h := range mh {
 		sg := h.ServiceGenerator()
 		if sg != nil {
```
prashantv: nit: can inline into the `if`

prashantv:
lgtm

internal/plugin/multi.go
abhinav: 👍

PR #206, opened 2016-09-03
plugin: Implement internal plugin

abhinav:
This is used to generate code for the plugin system API itself. This plugin is
enabled by using the hidden --generate-plugin-api flag.


abhinav:

internal/plugin/builtin/pluginapigen/handle.go
```diff
@@ -0,0 +1,256 @@
...
+	"path/filepath"
+	"strings"
+
+	iplugin "github.com/thriftrw/thriftrw-go/internal/plugin"
+	"github.com/thriftrw/thriftrw-go/plugin"
+	"github.com/thriftrw/thriftrw-go/plugin/api"
+)
+
+// Handle is a plugin.Handle that generates code for the plugin system API.
+var Handle iplugin.Handle = handle{}
```
breerly: Suggest `intplugin` instead of `iplugin` - feels like an Apple product announcement otherwise.

internal/plugin/builtin/pluginapigen/handle.go
```diff
@@ -0,0 +1,256 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	return &api.GenerateServiceResponse{Files: files}, nil
+}
+
+// convenience function because "index .Request.Services .Service.ParentID"
+// doesn't work. Index expects a ServiceID, not *ServiceID.
+func getService(req *api.GenerateServiceRequest, id api.ServiceID) *api.Service {
```
breerly: Looks like this function should actually be a method on `req`.

internal/plugin/builtin/pluginapigen/handle.go
```diff
@@ -0,0 +1,256 @@
...
+	return req.Services[id]
+}
+
+var templateOptions = []plugin.TemplateOption{
+	plugin.TemplateFunc("basename", filepath.Base),
+	plugin.TemplateFunc("getService", getService),
+}
+
+const interfaceTemplate = `
+// Code generated by thriftrw --generate-plugin-api
```
breerly: Nice, code gen comments for Phabricator and Github, ty sir

breerly:
lgtm :shipit:

plugin/gen.go
```diff
@@ -0,0 +1,24 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package plugin
+
+//go:generate thriftrw-go --generate-plugin-api api.thrift
+//go:generate ../scripts/updateLicenses.sh
```
breerly: Is it necessary to put these everywhere?

internal/plugin/builtin/pluginapigen/handle.go
```diff
@@ -0,0 +1,256 @@
...
+// @generated
+
+<$module := index .Request.Modules .Service.ModuleID>
+package <basename $module.Package>
+
+type <.Service.Name> interface {
+	<if .Service.ParentID>
```
prashantv: this is fine for now, but I imagine this pattern will exist in almost service generator. We should think about how to make this easier (even if it's just a function that will do all this that users will explicitly need to register using `TemplateFunc`)

internal/plugin/builtin/pluginapigen/handle.go
```diff
@@ -0,0 +1,256 @@
...
+		return
+	}
+
+	body, err = c.client.Send("<.ThriftName>", body)
+	if err != nil {
+		return
+	}
+
+	var result <.Name>Result
+	if err = result.FromWire(body); err != nil {
```
prashantv: nit: use `:=`

prashantv: oh you're using named returns, never mind.

prashantv:
lgtm

internal/plugin/builtin/pluginapigen/handle.go
abhinav: `api.GenerateServiceRequest` is generated from the Thrift file. We can't attach methods to it (yet).

plugin/gen.go
abhinav: It's not necessary to put it everywhere in that it could be at a higher level, but that leads to an annoying workflow where you `go generate` a package and then have to go to a higher level and `go generate` again to add the license headers back.

internal/plugin/builtin/pluginapigen/handle.go
abhinav: 👍

internal/plugin/builtin/pluginapigen/handle.go
abhinav: Valid concern. We can evolve the templating API over time. Right now, it doesn't depend on having the original GenerateServiceRequest available, and that makes sense because it could be used for things besides GenerateServiceRequest.

PR #209, opened 2016-09-06
naming: Handle unknown abbreviations

### LINKED ISSUE #208: Leave all-caps attributes alone
Currently, if we have an all-caps attribute that is not a known abbreviation, we pascal case it because the system detects it as `SCREAMING_SNAKE_CASE`. 

    struct Foo { 1: required bool VIP }

Gets converted to,

    type Foo struct { Vip bool }

We should fix this.

abhinav:
This changes naming in generated code in the following ways:

-   If an attribute is all-caps and is not a known abbreviation, we use it as-
    is.
-   Even if an enum item is all caps, we PascalCase it when we prepend the
    enum name.

Resolves #208

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ v0.2.0 (unreleased)
     Initially, only the generated code for `service` declarations is
     customizable. Check the documentation for more details.
 
+-   Breaking: Fixed a bug where all-caps attributes that are not known
+    abbreviations were changed to PascalCase.
+
 
 v0.1.0 (2016-08-31)
 -------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ v0.2.0 (unreleased)
 -   Breaking: Fixed a bug where all-caps attributes that are not known
     abbreviations were changed to PascalCase.
 
+-   Breaking: The `String()` method for `enum` types now returns the name of
+    the item as specified in the Thrift file.
+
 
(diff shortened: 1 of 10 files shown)


abhinav:

APTy:
thanks for this! 👍

abhinav:
Derp. I'll have another look at the enum naming. We *should* be able to keep the current enum behavior because we always prepend the enum name anyway: `RecordTypeName`, 'RecordTypeHomeAddress`.

abhinav:
Updated to maintain current `enum` behavior (which IMO is preferable anyway).

prashantv:
lgtm

can you update the title + description when merging

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ v0.2.0 (unreleased)
     ThriftRW now provides a plugin system to allow customizing code generation.
     Initially, only the generated code for `service` declarations is
     customizable. Check the documentation for more details.
+-   Breaking: Fixed a bug where all-caps attributes that are not known
+    abbreviations were changed to PascalCase.
+-   Breaking: The `String()` method for `enum` types now returns the name of
```
breerly: Suggest formatting like so:

- **BREAKING** - blah blah blah.

abhinav: Done https://github.com/thriftrw/thriftrw-go/pull/217

(I think the all-caps is unnecessary but the bold works.)

PR #216, opened 2016-09-07
Naming: Fix all-caps constant naming

### LINKED ISSUE #215: Constant naming
As of #209,

```
const i32 VERSION = 1
```

Generates,

```
const VERSION int32 = 1
```

But

```
const i32 APP_VERSION = 1
```

Generates,

```
const AppVersion int32 = 1
```

We should either always PascalCase constants if they're not known abbreviations, or just revert #209 and go back to known-abbreviations only.

abhinav:
Follow up to #209: That change would leave single word all caps constants
in that form as well. This fixes the behavior for constants to always pascal
case.

Resolves #215

CC @prashantv @breerly @kriskowal @bombela

### DIFF
--- a/gen/string.go
+++ b/gen/string.go
@@ -78,12 +78,6 @@ func constantName(s string) string {
 	return pascalCase(false /* all caps */, strings.Split(s, "_")...)
 }
 
-// enumItemName returns the Go name that should be used for an enum item with
-// the given Thrift name.
-func enumItemName(enumName string, itemName string) string {
-	return enumName + pascalCase(false /* all caps */, strings.Split(itemName, "_")...)
-}
-
 // goCase converts strings into PascalCase.
 func goCase(s string) string {
 	if len(s) == 0 {
--- a/gen/string.go
+++ b/gen/string.go
@@ -39,7 +39,11 @@ func isAllCaps(s string) bool {
 }
 
 // pascalCase combines the given words using PascalCase.
-func pascalCase(words ...string) string {
+//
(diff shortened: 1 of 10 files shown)


abhinav:
#209 already added the entry to the changelog. This is part of that same breaking change.

gen/string.go
```diff
@@ -70,12 +74,14 @@ func pascalCase(words ...string) string {
...
+	return pascalCase(false /* all caps */, strings.Split(s, "_")...)
+}
+
 // enumItemName returns the Go name that should be used for an enum item with
 // the given Thrift name.
 func enumItemName(enumName string, itemName string) string {
-	words := []string{enumName}
-	words = append(words, strings.Split(itemName, "_")...)
-	return pascalCase(words...)
+	return enumName + pascalCase(false /* all caps */, strings.Split(itemName, "_")...)
```
prashantv: You no longer pass `enumName` to `pascalCase`, so if the enum name is lowercase, will this return an unexported name?

abhinav: enumName is generated by `typeName` which imports the relevant package (if necessary) and returns `foo.Bar` (or just `Bar` if the type is local). I'll add a test for lower case enums.

abhinav: Actually, I'm also going to change `enumItemName` to accept the `EnumSpec` and generate the name from that so that there's no change of accidentally calling it with the wrong enum name.

prashantv:
lgtm

PR #219, opened 2016-09-09
Prepare v0.2.0

abhinav:
@breerly @kriskowal @bombela


breerly:
lgtm.

On a side note - I'm still not sure how customers should be gracefully upgrading to deal with these breaking changes.

PR #225, opened 2016-09-15
lexer: Disallow all reserved keywords

### LINKED ISSUE #221: parser: Disallow all reserved keywords
We do not yet disallow all reserved keywords that Apache Thrift does. See,

https://github.com/apache/thrift/blob/master/compiler/cpp/src/thriftl.ll#L292

The following works in ThriftRW but not Apache Thrift.

```
union Operation {
  1: optional Insert insert 
  2: optional Update update
  3: optional Delete delete
}
```

abhinav:
This disallows use of keywords marked as reserved by Thrift. The full list is
available here: https://github.com/apache/thrift/blob/775671aea41ea55427dd78d7ce68e282cc9b8487/compiler/cpp/src/thriftl.ll#L266

Resolves #221

### DIFF
--- a/gen/constant_test.go
+++ b/gen/constant_test.go
@@ -216,12 +216,12 @@ func TestConstants(t *testing.T) {
 			&ts.Graph{
 				Edges: []*ts.Edge{
 					{
-						Start: &ts.Point{X: 1, Y: 2},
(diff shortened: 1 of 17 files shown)


breerly:
I guess we'll want to do this in thriftrw-python and thriftrw-node. 

cc @kriskowal

PR #228, opened 2016-09-16
Update CHANGELOG

abhinav:
Forgot to add entry for #225 and update release date when 0.2.0 was released.

CC @kriskowal @breerly @bombela

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,14 @@
 Releases
 ========
 
-v0.2.0 (unreleased)
+v0.3.0 (unreleased)
+------------------
+
+-   **Breaking**: Most keywords reserved by Apache Thrift are now disallowed as
+    identifiers in Thrift files.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,14 @@
 Releases
 ========
 
-v0.2.0 (unreleased)
+v0.3.0 (unreleased)
+------------------
+
+-   **Breaking**: Most keywords reserved by Apache Thrift are now disallowed as
```
breerly: What do you "most"?

abhinav: It's really "all" but if they add more in the future, this is no longer true. But you're right. I'll change this.

breerly:
Why not export Version?

abhinav:
I'll export it eventually in a submodule. Nobody should be importing the `main` module directly.

PR #233, opened 2016-09-27
plugin: Qualify imports for base name mismatch

abhinav:
Imports in plugin templates previously had this logic: If the package can be
imported (it exists on the `GOPATH`), import it and determine its name.
Otherwise, use the base path. This doesn't work if the package is not
available on the `GOPATH` and has a `-` in the name.

This commit makes two fixes to this behavior:

-   Always strip hyphens out of the import name.
-   If the base name and the package name do not match, always use a qualified
    import regardless of whether the package name was determined by importing
    it or guessed based on its base name.

This should ensure that even if we guess the package name wrong, the code
still works.

CC @prashantv @breerly

### DIFF
--- a/internal/goast/package.go
+++ b/internal/goast/package.go
@@ -23,6 +23,7 @@ package goast
 import (
 	"go/build"
 	"path/filepath"
+	"strings"
 )
 
 // DeterminePackageName determines the name of the package at the given import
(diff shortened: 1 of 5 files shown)


abhinav:
CC @srussell-uber

internal/goast/package.go
```diff
@@ -35,7 +36,15 @@ func DeterminePackageName(importPath string) string {
...
 	if err != nil {
-		return filepath.Base(importPath)
+		return guessPackageName(importPath)
 	}
 	return pkg.Name
 }
+
+func guessPackageName(importPath string) string {
+	packageName := filepath.Base(importPath)
+	if strings.HasSuffix(packageName, "-go") {
```
breerly: Will this be necessary when we go to go.uber.org import paths?

abhinav: It's not necessary at all. It's just a nice-to-have for the common pattern of "foo-go" packages using the name "foo". We'll still use a qualified import for this because "foo-go" is not equal to "foo".

PR #239, opened 2016-10-05
gen: Use go.uber.org/yarpc for --yarpc

abhinav:
The CLI now warns users that they should use `--plugin=yarpc`.

CC @breerly

### DIFF
--- a/version.go
+++ b/version.go
@@ -20,4 +20,4 @@
 
 package main // import "go.uber.org/thriftrw"
 
-var version = "v0.4.0"
+var version = "v0.3.2"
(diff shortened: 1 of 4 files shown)


version.go
```diff
@@ -20,4 +20,4 @@
 
 package main // import "go.uber.org/thriftrw"
 
-var version = "v0.4.0"
+var version = "v0.3.2"
```
bombela: Is this the correct version here?

abhinav: Yes, because this is a bugfix.

PR #242, opened 2016-10-10
Oneway support

### LINKED ISSUE #139: Support for oneway functions

abhinav:
This adds support for generating code for `oneway` functions. It also exposes
whether a function is oneway or not to code generation plugins.

Resolves #139

CC @prashantv @breerly @kriskowal @AlexAPB


breerly:
cc @thuningxu

thuningxu:
Maybe I am not understanding this correctly, it seems to me that in service.go, are you just saying "if the call is one way, I don't do anything to populate the return value"? I thought we need to return an Error indicating retriable or not to the transport layer.

abhinav:
This is not the YARPC interface. This is generic generated code that is required for all uses of thriftrw regardless of whether it's YARPC or not. YARPC code is generated by the [ThriftRW YARPC Plugin](https://github.com/yarpc/yarpc-go/blob/dev/encoding/thrift/thriftrw-plugin-yarpc/main.go) and work to support oneway in YARPC is underway here: https://github.com/yarpc/yarpc-go/pull/378

PR #243, opened 2016-10-10
compile: Expose annotations for native types

### LINKED ISSUE #222: compile: Expose annotations for primitive type specs
Follow up to #195 

Annotations were not added for primitive type specs because that will require some API changes. We should do that before 1.0.

abhinav:
Resolves #222

### DIFF
--- a/compile/primitive.go
+++ b/compile/primitive.go
@@ -27,67 +27,117 @@ import (
 	"go.uber.org/thriftrw/wire"
 )
 
-// TypeSpecs for primitive Thrift types.
-var (
-	BoolSpec   TypeSpec = primitiveTypeSpec{Name: "bool", Code: wire.TBool}
-	I8Spec     TypeSpec = primitiveTypeSpec{Name: "byte", Code: wire.TI8}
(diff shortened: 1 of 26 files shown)


compile/primitive.go
```diff
@@ -27,67 +27,117 @@ import (
...
+func (*BoolSpec) ThriftName() string   { return "bool" }
+func (*I8Spec) ThriftName() string     { return "byte" }
+func (*I16Spec) ThriftName() string    { return "i16" }
+func (*I32Spec) ThriftName() string    { return "i32" }
+func (*I64Spec) ThriftName() string    { return "i64" }
+func (*DoubleSpec) ThriftName() string { return "double" }
+func (*StringSpec) ThriftName() string { return "string" }
+func (*BinarySpec) ThriftName() string { return "binary" }
+
+func (t *BoolSpec) Link(Scope) (TypeSpec, error)   { return t, nil }
```
prashantv: can native thrift type implement `Link` and `ForEachTypeReference`?

PR #244, opened 2016-10-12
A top-level version constant.

### LINKED ISSUE #229: Expose Version as a library-level constant
We should expose the current version of thriftrw as a library-level constant

bombela:
- Now we have a subpackage version with a constant Version.

Fixes #229

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,7 @@ v0.4.0 (unreleased)
     types from values to types.
 -   Generate args structs and helpers for oneway functions.
 -   Expose whether a function is oneway to plugins.
--   Library version is exposed via the subpackage version (version.Version).
-
+-   Expose the version of the library under `go.uber.org/thriftrw/version.Version`.
 
 v0.3.2 (2016-10-05)
 -------------------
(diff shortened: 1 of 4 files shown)


prashantv:
Yeah I'd suggest a subpackage (either `thriftrw` or `version`)

bombela:
I had a weird magical thing happening with go build, which likely pulled the binary over and over again from some sort of cache, so I ended up pushing here to confirm that I wasn't supposed to work!

bombela:
Now we have version.Version ...

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ v0.4.0 (unreleased)
     types from values to types.
 -   Generate args structs and helpers for oneway functions.
 -   Expose whether a function is oneway to plugins.
+-   Library version is exposed via the subpackage version (version.Version).
```
breerly: Nit - The convention here is to use backticks, so `version.Version`.

breerly: Suggest renaming to:

- Expose the version of the library under `go.uber.org/thriftrw/version.Version`.

PR #247, opened 2016-10-19
Generated code will be checked for version compatibility

### LINKED ISSUE #246: Enforce compatible version of generated code and thriftrw library
Right now the code generated by `./thriftw` is independent from the thriftrw library. But we need to use `thriftrw/wire` and eventually `thriftrw/protocol` with generated types in order to serialize/deserialize them.

We would like to prevent subtle bugs because of differences between the version of thriftrw used to generate the code and the version used at runtime.

#### Solution 1 - compile time

Possible solution is a compile time error based on missing const/variable containing the version in the name. In every generated file, we add:

```go
const _ = version.GenCodeCompatibleVersions.v3_4
```

And if `thrtiftrw/version.GenCodeCompatibleVersions.v3_4` doesn't exist, compilation will fail. The error might not be the most readable nor understandable, but has the merit of being unavoidable.

#### Solution 2 - runtime

We add a new generated file per generated go packages (ie: per generated directory), containing something like the following:

```go
func init() {
  version.IHaveBeenGeneratedWithThriftrwVersion(3, 4, 0)
}
```

The function `IHaveBeenGeneratedWithThriftrwVersion` in thriftrw is responsible for checking versions compatibility and panicking at startup. The error message can be made satisfyingly clear, in the expense of waiting until runtime to fail.

bombela:
- Every generated package is now calling `thriftrw/version.CheckCompatibilityWithGeneratedCodeAt()`
- thriftrw can then check if the version of thriftrw used to generate the code is compatible with the current version

fix #246


gen/generate.go
```diff
@@ -213,6 +214,36 @@ func generateModule(m *compile.Module, i thriftPackageImporter, builder *generat
...
+		}{
+			Version: version.Version,
+			Package: importPath,
+		}
+
+		if err := g.DeclareFromTemplate(`
+		<$version := import "go.uber.org/thriftrw/version">
+
+		func init() {
+			<$version>.CheckCompatibilityWithGeneratedCodeAt("<.Version>", "<.Package>")
```
breerly: Nice method name! Might consider `CheckCompatWithGeneratedCodeAt` instead.

version/semver_test.go
```diff
@@ -0,0 +1,105 @@
...
+		{"1.2.3-a.10", "1.2.3-a.5"},
+		{"1.2.3-a.b", "1.2.3-a"},
+		{"1.2.3-a.b", "1.2.3-a.5"},
+		{"1.2.3-a.b.c.10.d.5", "1.2.3-a.b.c.5.d.100"},
+		{"2.0.0", "1.2.3"},
+		{"3.0.0", "2.7.2+asdf"},
+		{"3.0.0+foobar", "2.7.2"},
+		{"3.0.0-hello.42+foobar.meta.39", "3.0.0-42.42+barfoo.tame.93"},
+	}
+	for _, test := range tests {
```
breerly: Suggest using subtests here, we're using 1.7+ after all.

version/version.go
```diff
@@ -20,4 +20,45 @@
...
+
+// CheckCompatibilityWithGeneratedCodeAt will panic if the thriftrw version
+// used to generated code (given by `genCodeVer`) is not compatible with the
+// current version of thriftrw.
+// This function is designed to be called during initialization of the
+// generated code.
+//
+// Rational: Let's say you use thriftrw version 1.0 to generate some stubs.
+// Later on, you imports the stubs, but also thriftrw in version 1.2. Maybe
+// thriftrw 1.2 is not compatible in subtle ways with the generated code from
```
abhinav: How about,

> It is possible that the old generated code is not compatible with the new
> version of ThriftRW in subtle ways but still compiles successfully. This
> function will ensure that the version mismatch is detected and help avoid
> bugs that could be caused by this discrepancy.

version/version.go
```diff
@@ -20,4 +20,45 @@
...
+// version 1.0. This function will make sure to panic during initialization
+// preventing potential bugs.
+func CheckCompatibilityWithGeneratedCodeAt(genCodeVersion string, fromPkg string) {
+	genv := parseSemVerOrPanic(genCodeVersion)
+	compatible := (genv.Compare(&genCodeCompatbilityRange.begin) >= 0 &&
+		genv.Compare(&genCodeCompatbilityRange.end) < 0)
```
abhinav: I could be wrong but you probably don't need the `(` if you're putting the
`&&` at the end of the line.

version/version.go
```diff
@@ -20,4 +20,45 @@
...
+			fromPkg, &genCodeCompatbilityRange.begin,
+			&genCodeCompatbilityRange.end, &genv)
+	}
+}
+
+var genCodeCompatbilityRange = computeGenCodeCompabilityRange()
+
+type genCodeCompatbilityRangeHolder struct {
+	begin semVer
+	end   semVer
```
abhinav: You can store `*semVer` here to avoid the `&` in
`genCodeCompatbilityRange.begin/end` above.

version/version.go
```diff
@@ -20,4 +20,45 @@
...
+	if !compatible {
+		log.Panicf(`incompatible version from generaged package "%s", expected >=%s and <%s, got %s`,
+			fromPkg, &genCodeCompatbilityRange.begin,
+			&genCodeCompatbilityRange.end, &genv)
+	}
+}
+
+var genCodeCompatbilityRange = computeGenCodeCompabilityRange()
+
+type genCodeCompatbilityRangeHolder struct {
```
abhinav: You don't need to declare a type. Anonymous structs work just fine.

```
var compatibleRange = struct{
  Begin *semVer
  End *semVer
}{}
```

And then set it up in `init`.

```
func init() {
  begin := parseSemVerOrPanic(Version)
  begin.Patch = 0
  compatibleRange.Begin = begin

  end := *begin
  end.Minor++
  compatibleRange.End = end
}
```

version/semver.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// Following http://semver.org/spec/v2.0.0.html
+const numPart = `([0-9]+)\.([0-9]+)\.([0-9]+)`
+const preReleasePart = `-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)`
+const metaPart = `\+([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)`
+
+var semVerRegex = regexp.MustCompile(`^` + numPart + `(?:` + preReleasePart + `)?(?:` + metaPart + `)?$`)
+
+func parseSemVer(v string) (semVer, error) {
```
abhinav: Why does this and `parseSemVerOrPanic` return `semVer` when all the receivers
are on `*semVer`. If `*semVer` is the way to use these objects, these
functions should return pointers too.

gen/generate.go
```diff
@@ -213,6 +214,36 @@ func generateModule(m *compile.Module, i thriftPackageImporter, builder *generat
...
+		}
+
+		`, data); err != nil {
+			return nil, err
+		}
+
+		buff := new(bytes.Buffer)
```
bombela: meh, whatever, did it because I like you

version/semver.go
```diff
@@ -0,0 +1,158 @@
...
+	Minor uint
+	Patch uint
+	Pre   []string
+	Meta  string
+}
+
+// Following http://semver.org/spec/v2.0.0.html
+const numPart = `([0-9]+)\.([0-9]+)\.([0-9]+)`
+const preReleasePart = `-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)`
+const metaPart = `\+([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)`
```
bombela: yeah named group endup annoying to use in this specific case. i like 0-9 because it is very explicit, but \d it is now.

version/semver.go
bombela: because i hate the idea of allocation shit everywhere in ram, yes there is a GC, yes this is a cold path. But it makes me itch.

version/semver.go
```diff
@@ -0,0 +1,158 @@
...
+
+var semVerRegex = regexp.MustCompile(`^` + numPart + `(?:` + preReleasePart + `)?(?:` + metaPart + `)?$`)
+
+func parseSemVer(v string) (semVer, error) {
+	r := semVer{}
```
bombela: done

version/semver.go
```diff
@@ -0,0 +1,158 @@
...
+	parts := semVerRegex.FindStringSubmatch(v)
+	if parts == nil {
+		return r, fmt.Errorf(`cannot parse as semantic version: "%s"`, v)
```
bombela: done

version/semver.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		panic(err)
+	}
+	return semVer
+}
+
+func parseUint(s string) (uint, error) {
+	var v uint64
+	var err error
+	if v, err = strconv.ParseUint(s, 10, 31); err != nil {
```
bombela: done

version/semver.go
```diff
@@ -0,0 +1,158 @@
...
+		return 0, err
+	}
+	return uint(v), nil
+}
+
+func (v *semVer) String() string {
+	r := fmt.Sprintf("%d.%d.%d", v.Major, v.Minor, v.Patch)
+	if v.Pre != nil {
```
bombela: I thought Go made sure to match with nil even if the slide is allocated but empty? anyway, done.

version/semver.go
```diff
@@ -0,0 +1,158 @@
...
+		r = fmt.Sprintf("%s-%s", r, strings.Join(v.Pre, "."))
```
bombela: done

version/version.go
```diff
@@ -20,4 +20,45 @@
 
 package version
 
+import "log"
+
+// Version is the current thriftrw version.
```
bombela: done

version/version.go
```diff
@@ -20,4 +20,45 @@
...
 const Version = "0.4.0"
+
+// CheckCompatibilityWithGeneratedCodeAt will panic if the thriftrw version
+// used to generated code (given by `genCodeVer`) is not compatible with the
```
bombela: better safe than sorry.

version/version.go
```diff
@@ -20,4 +20,45 @@
...
+// current version of thriftrw.
+// This function is designed to be called during initialization of the
+// generated code.
+//
+// Rational: Let's say you use thriftrw version 1.0 to generate some stubs.
```
bombela: thats definitively true. corrected.

version/version.go
bombela: nice, done.

version/version.go
bombela: i am not even gonna try.

version/version.go
```diff
@@ -20,4 +20,45 @@
...
+// Later on, you imports the stubs, but also thriftrw in version 1.2. Maybe
+// thriftrw 1.2 is not compatible in subtle ways with the generated code from
+// version 1.0. This function will make sure to panic during initialization
+// preventing potential bugs.
+func CheckCompatibilityWithGeneratedCodeAt(genCodeVersion string, fromPkg string) {
+	genv := parseSemVerOrPanic(genCodeVersion)
+	compatible := (genv.Compare(&genCodeCompatbilityRange.begin) >= 0 &&
+		genv.Compare(&genCodeCompatbilityRange.end) < 0)
+	if !compatible {
+		log.Panicf(`incompatible version from generaged package "%s", expected >=%s and <%s, got %s`,
```
bombela: done thanks

version/version.go
bombela: but i dont want to use an init() function. my version cannot really forget to init the variable...

version/version.go
bombela: me no like pointer

version/semver.go
abhinav: I'm happy to use pass-by-value as well. I don't feel strongly about this; I
just dislike the `Compare(&compatibleRange.begin)` business. I would like the
object to be stored in the way it's used.

version/semver.go
abhinav: Huh? I don't understand.

version/version.go
abhinav: I'm not really concerned about the `init`. I just think the type is
unnecessary and an anonymous struct will suffice. The reason to do it in
`init` is because returning the anonymous type from a function will mean
repeating it.

Also, correction about the anonymous struct example: You would actually just
do:

```
var compatibleRange struct {
    Begin *semVer
    End *semVer
}

func init() {
    compatibleRange.Begin = ...
    compatibleRange.End = ...
}
```

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ v0.4.0 (unreleased)
 -   Generate args structs and helpers for oneway functions.
 -   Expose whether a function is oneway to plugins.
 -   Expose the version of the library under `go.uber.org/thriftrw/version.Version`.
+-   Generated code will test for version compatibility with the current version
+    of thriftrw during initialization.
```
abhinav: ThriftRW

version/version.go
```diff
@@ -20,4 +20,44 @@
...
+// Rationale: It is possible that the old generated code is not compatible with
+// the new version of ThriftRW in subtle ways but still compiles successfully.
+// This function will ensure that the version mismatch is detected and help
+// avoid bugs that could be caused by this discrepancy.
+func CheckCompatWithGeneratedCodeAt(genCodeVersion string, fromPkg string) {
+	genv := parseSemVerOrPanic(genCodeVersion)
+	compatible := (genv.Compare(&genCodeCompatbilityRange.begin) >= 0 &&
+		genv.Compare(&genCodeCompatbilityRange.end) < 0)
+	if !compatible {
+		log.Panicf(`incompatible version from generated package "%s", expected >=%s and <%s, got %s`,
```
abhinav: Use `%q` instead of `"%s"`

CHANGELOG.md
bombela: done

version/version.go
bombela: done

PR #248, opened 2016-10-25
compile: Expose annotations on typedefs

abhinav:
This is a follow up to #195 and #243. It exposes annotations on the
compiled representation of typedefs. With this, all elements of the
Thrift AST that can have annotations now expose them in the compiled
specs.

CC @breerly @kriskowal @bombela

### DIFF
--- a/compile/typedef.go
+++ b/compile/typedef.go
@@ -29,9 +29,10 @@ import (
 type TypedefSpec struct {
 	linkOnce
 
-	Name   string
-	File   string
-	Target TypeSpec
+	Name        string
+	File        string
+	Target      TypeSpec
+	Annotations Annotations
 
 	root TypeSpec
 }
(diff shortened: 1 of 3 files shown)


compile/typedef.go
```diff
@@ -43,10 +44,16 @@ func compileTypedef(file string, src *ast.Typedef) (*TypedefSpec, error) {
 		return nil, err
 	}
 
+	annotations, err := compileAnnotations(src.Annotations)
+	if err != nil {
+		return nil, err
```
bombela: should this code builds a more complicated error? like at https://github.com/thriftrw/thriftrw-go/blob/0f849e351d9a2f5ee0553480b8e2f8b7cbc2668b/compile/enum.go#L65

abhinav: True. Fixed

PR #249, opened 2016-10-25
support (go.name) annotations

### LINKED ISSUE #223: Name normalization collisions
Consider,

```
struct User {
  1: required string firstName
  2: required string first_name
}
```

This will fail or generate invalid code because both attributes become `FirstName`. We need to account for collisions in attribute naming when normalizing during code generation.

We can pick `FirstName2` for the second one by default and maybe allow overriding it with a `go.name` annotation.

```
struct User {
  1: required string firstName
  2: required string first_name (go.name = "OtherFirstName")
}
```

bombela:
Also adds early detection of name normalization collisions. Avoiding collisions
can be done via (go.name).

fixes #223


gen/string.go
```diff
@@ -25,8 +25,19 @@ import (
 	"strings"
 	"unicode"
 	"unicode/utf8"
+
+	"go.uber.org/thriftrw/compile"
 )
 
+// ItemSpec is any Thrift compile items with a name and annotations.
+type ItemSpec interface {
```
kriskowal: Perhaps just call this Spec and hoist it up somewhere common. This is an interface that captures all of the spec types.

gen/collision_test.go
```diff
@@ -0,0 +1,205 @@
...
+	}{
+		{tc.MyEnum2X, "X", 12},
+	}
+	for _, tt := range tests {
+		assert.Equal(t, tt.n, tt.e.String())
+		assert.Equal(t, tt.v, int64(tt.e))
+	}
+}
+
+func troughWire(t *testing.T, x thriftType, msg string, args ...interface{}) thriftType {
```
bombela: yeah roundTrip is used already in another file. Also, yes, I meant through. I never know which one is which.

gen/collision_test.go
```diff
@@ -0,0 +1,205 @@
...
+	message := fmt.Sprintf(msg, args...)
```
bombela: You did the same thing in another test file, I thought it was some good practice to follow :) You are my inspiration man!

gen/string.go
bombela: discussed offline:
 - renamed to namedEntity, private interface to strings.go
 - TypeSpec doesn't require ThriftAnnotations as not all TypeSpec are offering annotations.

Composition wins!

gen/string.go
```diff
@@ -90,6 +101,41 @@ func goCase(s string) string {
...
+		return "", fmt.Errorf("%q is not a go style identifier (suggestion: %q)", name, goCase(name))
+	}
+
+	return name, nil
+}
+
+func goName(spec ItemSpec) (string, error) {
+	name, err := goNameAnnotation(spec)
+	if err == nil && name == "" {
+		name = goCase(spec.ThriftName())
```
bombela: you never know :p

gen/collision_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+	xType := reflect.TypeOf(x)
+	if xType.Kind() == reflect.Ptr {
+		xType = xType.Elem()
+	}
+
+	gotX := reflect.New(xType)
+	errval := gotX.MethodByName("FromWire").
+		Call([]reflect.Value{reflect.ValueOf(v)})[0].
+		Interface()
```
abhinav: I suspected it would play weirdly with non-nil errors that hold nil values
but after trying it out (https://play.golang.org/p/KPeBM_TaR6) it seems that
it will be fine.

gen/plugin.go
```diff
@@ -250,9 +254,10 @@ func (g *generateServiceBuilder) buildType(spec compile.TypeSpec, required bool)
 		if err != nil {
 			return nil, err
 		}
+		name, err := goName(s)
```
abhinav: handle the error

gen/string.go
```diff
@@ -90,6 +92,55 @@ func goCase(s string) string {
...
+	if !capitalized || underscore {
+		var emsg []string
+		if underscore {
+			emsg = append(emsg, "contains underscores")
+		}
+		if !capitalized {
+			emsg = append(emsg, "is not capitalized")
+		}
+
+		return "", fmt.Errorf("%q (from go.name annotation) is not a Go style public identifier (%s), suggestion: %q)", name, strings.Join(emsg, ", "), goCase(name))
```
abhinav: nice

gen/plugin.go
bombela: done

bombela:
- Ok so as we found out, there is no annotations on constant.
 - changelog updated

PR #254, opened 2016-10-31
Generated enums first class JSON support

### LINKED ISSUE #214: ThriftRW-go should marshall enums to JSON as the name string
State of things:

Regardless of implementation, the intent is for the user to use the name of an enum and not to cast a number out of thin air. The number value is intended for the wire, and the name intended for code.

In Java, Python, and Go, enums are named numbers in code. The generated code produces a constant with the corresponding number. In JavaScript, consequently TCurl, enums values are expressed with the string name and only represented as a number on the wire. This is to allow users to express the name in both TCurl and in code, which is especially important because the idiom in JavaScript is to express Thrift requests and responses without necessarily importing the IDL model.

Consequently, when marshaling a Thrift model in Go to JSON, it should produce JSON suitable for use in JavaScript. Enum values should marshal to the string name.

bombela:
Implements https://golang.org/pkg/encoding/json/#Marshaler for generated enum types.

enums are marshaled to a JSON string, with fall back to a JSON number. unmarshalling is symetrical.

fix #214

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ v0.4.0 (2016-11-01)
 -   **Breaking**: The `compile` API now exposes annotations made while
     referencing native Thrift types. This changes the `TypeSpec`s for primitive
     types from values to types.
+-   **Breaking**: Generated enums now have first class JSON support. Enums are
+    (un)marshalled from/to strings if possible with fall back to integer.
 -   The `compile` API now also exposes annotations for `typedef` declarations.
 -   Generate args structs and helpers for oneway functions.
 -   Expose whether a function is oneway to plugins.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ v0.4.0 (2016-11-01)
     referencing native Thrift types. This changes the `TypeSpec`s for primitive
     types from values to types.
 -   **Breaking**: Generated enums now have first class JSON support. Enums are
-    (un)marshalled from/to strings if possible with fall back to integer.
+    (un)marshalled from/to strings if possible with fallback to integer for
+	unrecognized values.
 -   The `compile` API now also exposes annotations for `typedef` declarations.
 -   Generate args structs and helpers for oneway functions.
 -   Expose whether a function is oneway to plugins.
--- a/gen/enum.go
+++ b/gen/enum.go
@@ -144,10 +144,10 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 					return <$err>
 				}
 				if <$x> <">="> 0x80000000 {
-					return <$fmt>.Errorf("enum overflow from JSON %q for \"<$enumName>\"", text)
+					return <$fmt>.Errorf("enum overflow from JSON %q for %q", text, "<$enumName>")
 				}
 				if <$x> <"<"> -0x80000000 {
-					return <$fmt>.Errorf("enum underflow from JSON %q for \"<$enumName>\"", text)
+					return <$fmt>.Errorf("enum underflow from JSON %q for %q", text, "<$enumName>")
 				}
 				*<$v> = (<$enumName>)(<$x>)
 				return nil
@@ -162,11 +162,11 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 						return nil
 				<end>
 					default:
-						return <$fmt>.Errorf("unknown enum value %q for \"<$enumName>\"", <$w>)
+						return <$fmt>.Errorf("unknown enum value %q for %q", <$w>, "<$enumName>")
 				}
 			}
 
-			return <$fmt>.Errorf("invalid JSON value %T $q to unmarshal into \"<$enumName>\"", <$t>, <$t>)
+			return <$fmt>.Errorf("invalid JSON value $q (%T) to unmarshal into %q", <$t>, <$t>, "<$enumName>")
 		}
 
 		func (<$v> *<$enumName>) UnmarshalJSON2(text []byte) error {
--- a/gen/enum.go
+++ b/gen/enum.go
@@ -156,7 +156,7 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 				<$enum := .Spec>
 				<range .Spec.Items>
 					case "<.Name>":
-						*<$v> = <enumItemName $enum .Name>
+						*<$v> = <enumItemName $enumName .>
 						return nil
 				<end>
 					default:
--- a/gen/enum.go
+++ b/gen/enum.go
@@ -166,39 +166,6 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 
 			return <$fmt>.Errorf("invalid JSON value %q (%T) to unmarshal into %q", <$t>, <$t>, "<$enumName>")
 		}
-
-		func (<$v> *<$enumName>) UnmarshalJSON2(<$text> []byte) error {
-			<$e := newVar "e">
-
-			<$w>, err := <$strconv>.ParseInt(string(<$text>), 10, 32)
-			if err == nil {
-				*<$v> = (<$enumName>)(<$w>)
-				return nil
-			}
-
-			<$e> := err.(*strconv.NumError)
-			if <$e>.Err != <$strconv>.ErrSyntax {
-				return err
-			}
-
-			<$s := newVar "s">
-			var <$s> string
-			if err := <$json>.Unmarshal(<$text>, &<$s>); err != nil {
-				return err
-			}
-
-			<if len .Spec.Items>
-				switch string(<$s>) {
-				<range .Spec.Items>
-					case "<.Name>":
-						*<$v> = (<$enumName>)(<.Value>)
-						return nil
-				<end>
-				}
-			<end>
-
-			return <$fmt>.Errorf("unknown enum value %q for \"<$enumName>\"", <$s>)
-		}
 		`,
 		struct {
 			Spec        *compile.EnumSpec
--- a/gen/enum.go
+++ b/gen/enum.go
@@ -112,8 +112,7 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 
 		func (<$v> <$enumName>) MarshalJSON() ([]byte, error) {
 			<if len .Spec.Items>
-				<$w> := int32(<$v>)
-				switch <$w> {
+				switch int32(<$v>) {
 				<range .UniqueItems>
 					case <.Value>:
 						return ([]byte)("\"<.Name>\""), nil
@@ -123,31 +122,30 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 			return ([]byte)(<$strconv>.FormatInt(int64(<$v>), 10)), nil
 		}
 
-		func (<$v> *<$enumName>) UnmarshalJSON(text []byte) error {
+		<$text := newVar "text">
+		func (<$v> *<$enumName>) UnmarshalJSON(<$text> []byte) error {
 			<$d := newVar "d">
 			<$t := newVar "t">
-			<$err := newVar "err">
 
-			<$d> := <$json>.NewDecoder(<$bytes>.NewReader(text))
+			<$d> := <$json>.NewDecoder(<$bytes>.NewReader(<$text>))
 			<$d>.UseNumber()
-			<$t>, <$err> := <$d>.Token()
-
-			if <$err> != nil {
-				return <$err>
+			<$t>, err := <$d>.Token()
+			if err != nil {
+				return err
 			}
 
 			<$ok := newVar "ok">
 			if <$w>, <$ok> := <$t>.(<$json>.Number); <$ok> {
 				<$x := newVar "x">
-				<$x>, <$err> := <$w>.Int64()
-				if <$err> != nil {
-					return <$err>
+				<$x>, err := <$w>.Int64()
+				if err != nil {
+					return err
 				}
 				if <$x> <">="> 0x80000000 {
-					return <$fmt>.Errorf("enum overflow from JSON %q for %q", text, "<$enumName>")
+					return <$fmt>.Errorf("enum overflow from JSON %q for %q", <$text>, "<$enumName>")
 				}
 				if <$x> <"<"> -0x80000000 {
-					return <$fmt>.Errorf("enum underflow from JSON %q for %q", text, "<$enumName>")
+					return <$fmt>.Errorf("enum underflow from JSON %q for %q", <$text>, "<$enumName>")
 				}
 				*<$v> = (<$enumName>)(<$x>)
 				return nil
@@ -166,28 +164,27 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 				}
 			}
 
-			return <$fmt>.Errorf("invalid JSON value $q (%T) to unmarshal into %q", <$t>, <$t>, "<$enumName>")
+			return <$fmt>.Errorf("invalid JSON value %q (%T) to unmarshal into %q", <$t>, <$t>, "<$enumName>")
 		}
 
-		func (<$v> *<$enumName>) UnmarshalJSON2(text []byte) error {
-			<$err := newVar "err">
+		func (<$v> *<$enumName>) UnmarshalJSON2(<$text> []byte) error {
 			<$e := newVar "e">
 
-			<$w>, <$err> := <$strconv>.ParseInt(string(text), 10, 32)
-			if <$err> == nil {
+			<$w>, err := <$strconv>.ParseInt(string(<$text>), 10, 32)
+			if err == nil {
 				*<$v> = (<$enumName>)(<$w>)
 				return nil
 			}
 
-			<$e> := <$err>.(*strconv.NumError)
+			<$e> := err.(*strconv.NumError)
 			if <$e>.Err != <$strconv>.ErrSyntax {
-				return <$err>
+				return err
 			}
 
 			<$s := newVar "s">
 			var <$s> string
-			if <$err> := <$json>.Unmarshal(text, &<$s>); <$err> != nil {
-				return <$err>
+			if err := <$json>.Unmarshal(<$text>, &<$s>); err != nil {
+				return err
 			}
 
 			<if len .Spec.Items>
--- a/gen/enum.go
+++ b/gen/enum.go
@@ -151,7 +151,7 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 				*<$v> = (<$enumName>)(<$x>)
 				return nil
 			case string:
-				switch string(<$w>) {
+				switch <$w> {
 				<$enum := .Spec>
 				<range .Spec.Items>
 					case "<.Name>":
--- a/gen/enum.go
+++ b/gen/enum.go
@@ -69,6 +69,8 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 	err := g.DeclareFromTemplate(
 		`
 		<$wire := import "go.uber.org/thriftrw/wire">
+		<$strconv := import "strconv">
+		<$fmt := import "fmt">
 
 		<$enumName := goName .Spec>
 		type <$enumName> int32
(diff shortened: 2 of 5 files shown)


gen/enum.go
```diff
@@ -101,6 +106,99 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 			<end>
 			return fmt.Sprintf("<$enumName>(%d)", <$w>)
 		}
+
+		func (<$v> <$enumName>) MarshalJSON() ([]byte, error) {
+			<if len .Spec.Items>
+				<$w> := int32(<$v>)
+				switch <$w> {
```
abhinav: Not using `<$w>` outside of switch. Can do `switch int32(<$v>)`.

gen/enum.go
```diff
@@ -101,6 +106,99 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+				<range .UniqueItems>
+					case <.Value>:
+						return ([]byte)("\"<.Name>\""), nil
```
abhinav: Can do `<printf "[]byte(%q)" .Name>` instead.

gen/enum.go
```diff
@@ -101,6 +106,99 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+				<end>
+				}
+			<end>
+			return ([]byte)(<$strconv>.FormatInt(int64(<$v>), 10)), nil
```
abhinav: Why use `FormatInt` instead of `fmt.Sprint(int32(<$v>))`?

gen/enum.go
```diff
@@ -101,6 +106,99 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+		}
+
+		func (<$v> *<$enumName>) UnmarshalJSON(text []byte) error {
+			<$d := newVar "d">
+			<$t := newVar "t">
+			<$err := newVar "err">
```
abhinav: Don't use `newVar` for err. That isn't done anywhere else. Assumption is that err is always shadowed.

gen/enum.go
```diff
@@ -101,6 +106,99 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+				<range .UniqueItems>
+					case <.Value>:
+						return ([]byte)("\"<.Name>\""), nil
+				<end>
+				}
+			<end>
+			return ([]byte)(<$strconv>.FormatInt(int64(<$v>), 10)), nil
+		}
+
+		func (<$v> *<$enumName>) UnmarshalJSON(text []byte) error {
```
abhinav: Use `newVar` for `text`.

gen/enum.go
```diff
@@ -101,6 +106,99 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+			<$d := newVar "d">
+			<$t := newVar "t">
+			<$err := newVar "err">
+
+			<$d> := <$json>.NewDecoder(<$bytes>.NewReader(text))
+			<$d>.UseNumber()
+			<$t>, <$err> := <$d>.Token()
+
```
abhinav: nitty nit: rm empty line

gen/enum.go
```diff
@@ -101,6 +106,99 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+			<$ok := newVar "ok">
+			if <$w>, <$ok> := <$t>.(<$json>.Number); <$ok> {
+				<$x := newVar "x">
+				<$x>, <$err> := <$w>.Int64()
+				if <$err> != nil {
+					return <$err>
+				}
+				if <$x> <">="> 0x80000000 {
+					return <$fmt>.Errorf("enum overflow from JSON %q for %q", text, "<$enumName>")
+				}
```
abhinav: nice

gen/enum.go
```diff
@@ -101,6 +106,99 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+
+			<$ok := newVar "ok">
+			if <$w>, <$ok> := <$t>.(<$json>.Number); <$ok> {
+				<$x := newVar "x">
+				<$x>, <$err> := <$w>.Int64()
+				if <$err> != nil {
+					return <$err>
+				}
+				if <$x> <">="> 0x80000000 {
+					return <$fmt>.Errorf("enum overflow from JSON %q for %q", text, "<$enumName>")
```
abhinav: Don't need to format `<$enumName>` using `Errorf`. It's known at codegen time so you can make it part of the string.

gen/enum.go
```diff
@@ -101,6 +106,99 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+					case "<.Name>":
+						*<$v> = <enumItemName $enum .Name>
+						return nil
+				<end>
+					default:
+						return <$fmt>.Errorf("unknown enum value %q for %q", <$w>, "<$enumName>")
+				}
+			}
+
+			return <$fmt>.Errorf("invalid JSON value $q (%T) to unmarshal into %q", <$t>, <$t>, "<$enumName>")
```
abhinav: `$q`?

gen/enum.go
```diff
@@ -101,6 +106,99 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+		}
+
+		func (<$v> *<$enumName>) UnmarshalJSON2(text []byte) error {
```
abhinav: what?

gen/enum.go
bombela: Because its easy to read and faster. Doing less work is always better. Less code, less bug. Less everything. Let's just delete everything, nothing cannot have bugs.

```go
func Sprint(a ...interface{}) string {
	p := newPrinter()
	p.doPrint(a) // WE GO THERE
	s := string(p.buf)
	p.free()
	return s
}

func (p *pp) doPrint(a []interface{}) {
	prevString := false
	for argNum, arg := range a {
		isString := arg != nil && reflect.TypeOf(arg).Kind() == reflect.String
		// Add a space between two non-string arguments.
		if argNum > 0 && !isString && !prevString {
			p.buf.WriteByte(' ')
		}
		p.printArg(arg, 'v') // AND NOW HERE
		prevString = isString
	}
}

func (p *pp) printArg(arg interface{}, verb rune) {
	p.arg = arg
	p.value = reflect.Value{}

	if arg == nil {
		switch verb {
		case 'T', 'v':
			p.fmt.padString(nilAngleString)
		default:
			p.badVerb(verb)
		}
		return
	}

	// Special processing considerations.
	// %T (the value's type) and %p (its address) are special; we always do them first.
	switch verb {
	case 'T':
		p.fmt.fmt_s(reflect.TypeOf(arg).String())
		return
	case 'p':
		p.fmtPointer(reflect.ValueOf(arg), 'p')
		return
	}

	// Some types can be done without reflection.
	switch f := arg.(type) {
	case bool:
		p.fmtBool(f, verb)
	case float32:
		p.fmtFloat(float64(f), 32, verb)
	case float64:
		p.fmtFloat(f, 64, verb)
	case complex64:
		p.fmtComplex(complex128(f), 64, verb)
	case complex128:
		p.fmtComplex(f, 128, verb)
	case int:
		p.fmtInteger(uint64(f), signed, verb)
	case int8:
		p.fmtInteger(uint64(f), signed, verb)
	case int16:
		p.fmtInteger(uint64(f), signed, verb)
	case int32:
		p.fmtInteger(uint64(f), signed, verb) // WE GET HERE NOW
	case int64:
		p.fmtInteger(uint64(f), signed, verb)
	case uint:
		p.fmtInteger(uint64(f), unsigned, verb)
	case uint8:
		p.fmtInteger(uint64(f), unsigned, verb)
	case uint16:
		p.fmtInteger(uint64(f), unsigned, verb)
	case uint32:
		p.fmtInteger(uint64(f), unsigned, verb)
	case uint64:
		p.fmtInteger(f, unsigned, verb)
	case uintptr:
		p.fmtInteger(uint64(f), unsigned, verb)
	case string:
		p.fmtString(f, verb)
	case []byte:
		p.fmtBytes(f, verb, "[]byte")
	case reflect.Value:
		p.printValue(f, verb, 0)
	default:
		// If the type is not simple, it might have methods.
		if !p.handleMethods(verb) {
			// Need to use reflection, since the type had no
			// interface methods that could be used for formatting.
			p.printValue(reflect.ValueOf(f), verb, 0)
		}
	}
}

func (p *pp) fmtInteger(v uint64, isSigned bool, verb rune) {
	switch verb {
	case 'v':
		if p.fmt.sharpV && !isSigned {
			p.fmt0x64(v, true)
		} else {
			p.fmt.fmt_integer(v, 10, isSigned, ldigits) // RIGHT THERE DEEPER
		}
	case 'd':
		p.fmt.fmt_integer(v, 10, isSigned, ldigits)
	case 'b':
		p.fmt.fmt_integer(v, 2, isSigned, ldigits)
	case 'o':
		p.fmt.fmt_integer(v, 8, isSigned, ldigits)
	case 'x':
		p.fmt.fmt_integer(v, 16, isSigned, ldigits)
	case 'X':
		p.fmt.fmt_integer(v, 16, isSigned, udigits)
	case 'c':
		p.fmt.fmt_c(v)
	case 'q':
		if v <= utf8.MaxRune {
			p.fmt.fmt_qc(v)
		} else {
			p.badVerb(verb)
		}
	case 'U':
		p.fmt.fmt_unicode(v)
	default:
		p.badVerb(verb)
	}
}
```

So we finally reach `func (f *fmt) fmt_integer(u uint64, base int, isSigned bool, digits string)`.

Let's see the other one:

```go
func FormatInt(i int64, base int) string {
	_, s := formatBits(nil, uint64(i), base, i < 0, false)
	return s
}
```

And we reach out to `formatBits` which is similar to the `fmt_integer` from above.

And if you are really curious:

```go
func (f *fmt) fmt_integer(u uint64, base int, isSigned bool, digits string) {
	negative := isSigned && int64(u) < 0
	if negative {
		u = -u
	}

	buf := f.intbuf[0:]
	// The already allocated f.intbuf with a capacity of 68 bytes
	// is large enough for integer formatting when no precision or width is set.
	if f.widPresent || f.precPresent {
		// Account 3 extra bytes for possible addition of a sign and "0x".
		width := 3 + f.wid + f.prec // wid and prec are always positive.
		if width > len(buf) {
			// We're going to need a bigger boat.
			buf = make([]byte, width)
		}
	}

	// Two ways to ask for extra leading zero digits: %.3d or %03d.
	// If both are specified the f.zero flag is ignored and
	// padding with spaces is used instead.
	prec := 0
	if f.precPresent {
		prec = f.prec
		// Precision of 0 and value of 0 means "print nothing" but padding.
		if prec == 0 && u == 0 {
			oldZero := f.zero
			f.zero = false
			f.writePadding(f.wid)
			f.zero = oldZero
			return
		}
	} else if f.zero && f.widPresent {
		prec = f.wid
		if negative || f.plus || f.space {
			prec-- // leave room for sign
		}
	}

	// Because printing is easier right-to-left: format u into buf, ending at buf[i].
	// We could make things marginally faster by splitting the 32-bit case out
	// into a separate block but it's not worth the duplication, so u has 64 bits.
	i := len(buf)
	// Use constants for the division and modulo for more efficient code.
	// Switch cases ordered by popularity.
	switch base {
	case 10:
		for u >= 10 {
			i--
			next := u / 10
			buf[i] = byte('0' + u - next*10)
			u = next
		}
	case 16:
		for u >= 16 {
			i--
			buf[i] = digits[u&0xF]
			u >>= 4
		}
	case 8:
		for u >= 8 {
			i--
			buf[i] = byte('0' + u&7)
			u >>= 3
		}
	case 2:
		for u >= 2 {
			i--
			buf[i] = byte('0' + u&1)
			u >>= 1
		}
	default:
		panic("fmt: unknown base; can't happen")
	}
	i--
	buf[i] = digits[u]
	for i > 0 && prec > len(buf)-i {
		i--
		buf[i] = '0'
	}

	// Various prefixes: 0x, -, etc.
	if f.sharp {
		switch base {
		case 8:
			if buf[i] != '0' {
				i--
				buf[i] = '0'
			}
		case 16:
			// Add a leading 0x or 0X.
			i--
			buf[i] = digits[16]
			i--
			buf[i] = '0'
		}
	}

	if negative {
		i--
		buf[i] = '-'
	} else if f.plus {
		i--
		buf[i] = '+'
	} else if f.space {
		i--
		buf[i] = ' '
	}

	// Left padding with zeros has already been handled like precision earlier
	// or the f.zero flag is ignored due to an explicitly set precision.
	oldZero := f.zero
	f.zero = false
	f.pad(buf[i:])
	f.zero = oldZero
}

func formatBits(dst []byte, u uint64, base int, neg, append_ bool) (d []byte, s string) {
	if base < 2 || base > len(digits) {
		panic("strconv: illegal AppendInt/FormatInt base")
	}
	// 2 <= base && base <= len(digits)

	var a [64 + 1]byte // +1 for sign of 64bit value in base 2
	i := len(a)

	if neg {
		u = -u
	}

	// convert bits
	if base == 10 {
		// common case: use constants for / because
		// the compiler can optimize it into a multiply+shift

		if ^uintptr(0)>>32 == 0 {
			for u > uint64(^uintptr(0)) {
				q := u / 1e9
				us := uintptr(u - q*1e9) // us % 1e9 fits into a uintptr
				for j := 9; j > 0; j-- {
					i--
					qs := us / 10
					a[i] = byte(us - qs*10 + '0')
					us = qs
				}
				u = q
			}
		}

		// u guaranteed to fit into a uintptr
		us := uintptr(u)
		for us >= 10 {
			i--
			q := us / 10
			a[i] = byte(us - q*10 + '0')
			us = q
		}
		// u < 10
		i--
		a[i] = byte(us + '0')

	} else if s := shifts[base]; s > 0 {
		// base is power of 2: use shifts and masks instead of / and %
		b := uint64(base)
		m := uintptr(b) - 1 // == 1<<s - 1
		for u >= b {
			i--
			a[i] = digits[uintptr(u)&m]
			u >>= s
		}
		// u < base
		i--
		a[i] = digits[uintptr(u)]

	} else {
		// general case
		b := uint64(base)
		for u >= b {
			i--
			q := u / b
			a[i] = digits[uintptr(u-q*b)]
			u = q
		}
		// u < base
		i--
		a[i] = digits[uintptr(u)]
	}

	// add sign, if any
	if neg {
		i--
		a[i] = '-'
	}

	if append_ {
		d = append(dst, a[i:]...)
		return
	}
	s = string(a[i:])
	return
}
```

gen/enum.go
bombela: done

gen/enum.go
bombela: done

gen/enum.go
bombela: there is a good reason: this keeps strings literal identical for all enums. The Go compiler will coalesce them all into a single one.

gen/enum.go
bombela: %q it is

gen/enum.go
bombela: a different version not using the json library. forgot to get rid of it :$

gen/enum.go
```diff
@@ -101,6 +106,96 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 			<end>
 			return fmt.Sprintf("<$enumName>(%d)", <$w>)
 		}
+
+		func (<$v> <$enumName>) MarshalJSON() ([]byte, error) {
+			<if len .Spec.Items>
+				switch int32(<$v>) {
+				<range .UniqueItems>
+					case <.Value>:
+						return ([]byte)("\"<.Name>\""), nil
```
prashantv: nit: no need for parens around `[]byte`, and you might want to use backtick quotes here,
```go
return []byte(`"<.Name>"`)
```

Also, is there any concern about the `.Name` containing quotes?

gen/enum.go
```diff
@@ -101,6 +106,96 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+					return <$fmt>.Errorf("enum overflow from JSON %q for %q", <$text>, "<$enumName>")
+				}
+				if <$x> <"<"> -0x80000000 {
+					return <$fmt>.Errorf("enum underflow from JSON %q for %q", <$text>, "<$enumName>")
+				}
+				*<$v> = (<$enumName>)(<$x>)
+				return nil
+			}
+
+			if <$w>, <$ok> := <$t>.(string); <$ok> {
```
abhinav: Styling nit: If you're doing two type matches, a type switch might be
preferable.

```
switch <$w> := <$t>.(type) {
case <$json>.Number:
    <$x>, err := <$w>.Int64()
    // ..
case string:
    switch <$w> {
    // ..
    }
default:
    return <$fmt>.Errorf("...")
}
```

gen/enum.go
```diff
@@ -101,6 +106,96 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+				<end>
+					default:
+						return <$fmt>.Errorf("unknown enum value %q for %q", <$w>, "<$enumName>")
+				}
+			}
+
+			return <$fmt>.Errorf("invalid JSON value %q (%T) to unmarshal into %q", <$t>, <$t>, "<$enumName>")
+		}
+
+		func (<$v> *<$enumName>) UnmarshalJSON2(<$text> []byte) error {
```
abhinav: You were going to delete this?

gen/enum.go
```diff
@@ -101,6 +106,96 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+			}
+
+			<$ok := newVar "ok">
+			if <$w>, <$ok> := <$t>.(<$json>.Number); <$ok> {
+				<$x := newVar "x">
+				<$x>, err := <$w>.Int64()
+				if err != nil {
+					return err
+				}
+				if <$x> <">="> 0x80000000 {
```
prashantv: Suggest using `> math.MaxInt32` and `< math.MinInt32` instead since it's more readable

gen/enum.go
```diff
@@ -101,6 +106,96 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+				}
+				*<$v> = (<$enumName>)(<$x>)
+				return nil
+			}
+
+			if <$w>, <$ok> := <$t>.(string); <$ok> {
+				switch string(<$w>) {
+				<$enum := .Spec>
+				<range .Spec.Items>
+					case "<.Name>":
```
prashantv: are the enums case insensitive? If so, should unmarshalling also be case insensitive or is this intentionally case sensitive?

gen/enum.go
prashantv: is this second implementation used or just another option?

gen/enum.go
```diff
@@ -101,6 +106,96 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 			<end>
 			return fmt.Sprintf("<$enumName>(%d)", <$w>)
 		}
+
+		func (<$v> <$enumName>) MarshalJSON() ([]byte, error) {
```
kriskowal: Should $v be parameterized? The variable name is scoped to the function. Does giving this variable freedom introduce the possibility of a name collision?

gen/enum.go
kriskowal: What is the distinction between UnmarshalJSON{, and 2}?

gen/enum.go
```diff
@@ -101,6 +106,96 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+				switch string(<$s>) {
+				<range .Spec.Items>
+					case "<.Name>":
+						*<$v> = (<$enumName>)(<.Value>)
+						return nil
+				<end>
+				}
+			<end>
+
+			return <$fmt>.Errorf("unknown enum value %q for \"<$enumName>\"", <$s>)
```
kriskowal: Is there no facility for quoting template variables?

gen/enum.go
bombela: good idea!

gen/enum.go
bombela: not used, and it was another way to implement it. one without using the json library, and likely faster because of that. both version worked with my unit tests the same. Anyway, we decided to go for the json version with abhinav, and I forgot to remove the code. It is now done.

gen/enum.go
bombela: right now we enforce enum items uniqueness case insensitively (`enum { x X }` is not allowed). so yes, in theory we could safely parse enum item names as case insensitive.

for now, let's keep things simple, and match as it is in the thrift file. we can always decide later to relax one of the two rule.

gen/enum.go
bombela: nope.

gen/enum.go
prashantv: good point, i did forget that!

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ v0.4.0 (2016-11-01)
 -   **Breaking**: The `compile` API now exposes annotations made while
     referencing native Thrift types. This changes the `TypeSpec`s for primitive
     types from values to types.
+-   **Breaking**: Generated enums now have first class JSON support. Enums are
+    (un)marshalled from/to strings if possible with fall back to integer.
```
abhinav: "fallback to integer for unrecognized values"

gen/enum.go
```diff
@@ -104,6 +110,61 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+				if <$x> <">"> <$math>.MaxInt32 {
+					return <$fmt>.Errorf("enum overflow from JSON %q for %q", <$text>, "<$enumName>")
+				}
+				if <$x> <"<"> <$math>.MinInt32 {
+					return <$fmt>.Errorf("enum underflow from JSON %q for %q", <$text>, "<$enumName>")
+				}
+				*<$v> = (<$enumName>)(<$x>)
+				return nil
+			case string:
+				switch string(<$w>) {
```
abhinav: `<$w>` is already of type string. You don't need to cast it.

PR #257, opened 2016-11-07
gen: Inline constant references

### LINKED ISSUE #251: Non-primitive constant should be "inline" constants to prevent mutation
Generated code for non-primitve type constants looks like this:

``` go
var StructConstant *MyStruct = &MyStruct{
   SomeField1: false, SomeField2: "false"}
```

The user code is indeed free to modify this "constant". In order to avoid side effect to the generated code, we should probably "inline" the constant where it is used in generated code.

Bonus question: do we want to offer a function returning a new const instead of a variable to our users to prevent this to ever happen in the first place?

bombela:
fixes #251

### DIFF
--- a/gen/constant_test.go
+++ b/gen/constant_test.go
@@ -27,11 +27,13 @@ import (
 	tc "go.uber.org/thriftrw/gen/testdata/containers"
 	te "go.uber.org/thriftrw/gen/testdata/enums"
 	tx "go.uber.org/thriftrw/gen/testdata/exceptions"
+	tok "go.uber.org/thriftrw/gen/testdata/other_constants"
 	ts "go.uber.org/thriftrw/gen/testdata/structs"
 	td "go.uber.org/thriftrw/gen/testdata/typedefs"
 	tu "go.uber.org/thriftrw/gen/testdata/unions"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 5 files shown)


abhinav:
changelog please

gen/constant_test.go
```diff
@@ -292,3 +294,24 @@ func TestConstants(t *testing.T) {
 		assert.Equal(t, tt.o, tt.i, tt.name)
 	}
 }
+
+func TestConstantsMutation(t *testing.T) {
+	originalX := tok.SomePoint.X
+
+	wireVal, err := tk.Graph.ToWire()
+	require.NoError(t, err)
+	g := ts.Graph{}
```
abhinav: nit: Use `var g ts.Graph` for zero-value structs

gen/constant_test.go
```diff
@@ -292,3 +294,24 @@ func TestConstants(t *testing.T) {
...
+	err = g.FromWire(wireVal)
+	require.NoError(t, err)
+	assert.Equal(t, g.Edges[0].StartPoint.X, originalX)
+
+	tok.SomePoint.X += 42
+	assert.Equal(t, tok.SomePoint.X, originalX+42.)
```
abhinav: Did you add the "." to make it a float? Can you add a 0 as well? "42.0"?

bombela: yes I did, wtf you want a zero for?

PR #269, opened 2016-11-10
Forbid generating code with thriftrw reserved identifiers

bombela:
Any advice how to test this?

### DIFF
--- a/gen/field.go
+++ b/gen/field.go
@@ -125,11 +125,11 @@ func (f *fieldGroupGenerator) declFieldName(fs *compile.FieldSpec) (string, erro
 		var note string
 		switch {
 		case originalName && fromAnnotation:
-			note = "(from go.name annotation)"
+			note = " (from go.name annotation)"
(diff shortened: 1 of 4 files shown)


gen/field.go
```diff
@@ -97,11 +115,23 @@ func (f *fieldGroupGenerator) declFieldName(fs *compile.FieldSpec) (string, erro
...
+	}
+
+	if err != nil {
+		originalName := (name == fs.ThriftName())
+		var note string
+		switch {
+		case originalName && fromAnnotation:
+			note = "(from go.name annotation)"
+		case !originalName && fromAnnotation:
+			note = fmt.Sprintf(" (from %q go.name annotation): %v", fs.ThriftName())
```
abhinav: inconsistent spacing: there will be a space before the `(` in the other two cases but not the first one.

abhinav: There's an extra `: %v` in this and the next case

PR #270, opened 2016-11-10
Properly check version of generated code

bombela:
with current version: `A.B.C`
generated code version is compatible when it is:
 - `>= A.0.0`
 - `< A.B+1.0`


version/version.go
```diff
@@ -34,7 +34,12 @@ func computeGenCodeCompabilityRange() (r genCodeCompatbilityRangeHolder) {
 	r.begin = parseSemVerOrPanic(Version)
 	r.begin.Patch = 0
 	r.end = r.begin
-	r.end.Minor++
+	if r.begin.Major == 0 {
+		r.end.Minor++
+	} else {
+		r.end.Minor = 0
```
abhinav: Don't we want to flip `begin.Minor` to 0 too? Otherwise we're enforcing the latest ('current') version of thriftrw as the minimum version for code that was generated earlier.

PR #272, opened 2016-11-11
Put ThriftRW version into a Thrift file

abhinav:
This puts the version of ThriftRW into an version/internal.thrift file.

@bombela @kriskowal


abhinav:
Discussed offline. We can do this without exposing it at the Thrift level.

bombela:
+1

PR #273, opened 2016-11-11
plugin: Send ThriftRW version in Handshake

abhinav:
This changes the plugin handshake protocol:

-   API Version was bumped. This will continue to serve its original purpose:
    Facilitate breaking changes in the plugin protocol.
-   The ThriftRW version itself is now sent over the wire. A future change
    will match this against the version of ThriftRW to ensure compatibility.
    This will enable adding features to the plugin protocol in a backwards
    compatible way by allowing plugins to state the version of ThriftRW they
    expect.
-   The VERSION constant was renamed to API_VERSION to clarify its meaning.

@kriskowal @bombela

### DIFF
--- a/plugin/api.thrift
+++ b/plugin/api.thrift
@@ -1,14 +1,9 @@
 /**
- * Package api defines the core interface which plugins use to talk to
- * ThriftRW.
+ * API_VERSION is the version of the plugin API.
+ *
+ * This MUST be provided in the HandshakeResponse.
  */
-
-/**
- * Version is the plugin API version specified in this IDL. This MUST be
- * provided in HandshakeResponses to specify which version of the API the
- * plugin implements.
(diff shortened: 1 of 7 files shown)


plugin/api.thrift
```diff
@@ -235,15 +230,20 @@ struct HandshakeResponse {
     /**
      * Version of the plugin API.
      *
-     * This is NOT the version of the plugin but the version of the plugin API
-     * which the plugin respects. This MUST match the version that ThriftRW
-     * expects or the program will fail.
+     * This MUST be set to API_VERSION by the plugin.
      */
-    2: required i32 apiVersion
+    2: required i32 apiVersion (go.name = "APIVersion")
```
bombela: ahah +1

PR #275, opened 2016-11-12
version: Refactor compatiblity check

abhinav:
This pulls compatibility check code into an internal package so we can re-use
it for plugins.

@bombela

### DIFF
--- a/internal/semver/compat_test.go
+++ b/internal/semver/compat_test.go
@@ -18,49 +18,52 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package version
+package semver
 
 import (
 	"testing"
-
-	"go.uber.org/thriftrw/internal/semver"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-func TestCodeCompatRange(t *testing.T) {
+func TestCompatibleRange(t *testing.T) {
 	test := []struct {
-		libVer     string
-		compatVer  []string
-		invalidVer []string
+		library      string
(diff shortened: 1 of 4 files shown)


internal/semver/compat_test.go
```diff
@@ -18,49 +18,52 @@
...
 
-func TestCodeCompatRange(t *testing.T) {
+func TestCompatibleRange(t *testing.T) {
 	test := []struct {
-		libVer     string
-		compatVer  []string
-		invalidVer []string
+		library      string
+		compatible   []string
+		incompatible []string
```
bombela: nice renaming

internal/semver/range.go
```diff
@@ -0,0 +1,32 @@
...
+package semver
+
+// Range is the range of version numbers that lie in [Begin, End).
+type Range struct {
+	Begin Version
+	End   Version
+}
+
+// Matches returns true if the given semver version number is in this range.
+func (r *Range) Matches(other Version) bool {
```
bombela: I think its better to use the word "contains" here.

version/check.go
```diff
@@ -35,10 +50,13 @@ import "log"
 // This function will ensure that the version mismatch is detected and help
 // avoid bugs that could be caused by this discrepancy.
 func CheckCompatWithGeneratedCodeAt(genCodeVersion string, fromPkg string) {
-	genv := parseSemVerOrPanic(genCodeVersion)
-	if !genCodeCompatbilityRange.IsCompatibleWith(genv) {
+	v, err := semver.Parse(genCodeVersion)
```
bombela: why not using `semver.ParseOrPanic` ?

abhinav: We don't have ParseOrPanic. I removed it.

PR #276, opened 2016-11-12
plugin: Verify ThriftRW version in handshake

abhinav:
This changes the plugin handshake to verify that the version of ThriftRW used
to compile the plugin is compatible with the current version of ThriftRW.

@bombela @kriskowal

### DIFF
--- a/internal/plugin/transport.go
+++ b/internal/plugin/transport.go
@@ -28,13 +28,27 @@ import (
 	"go.uber.org/thriftrw/internal"
 	"go.uber.org/thriftrw/internal/envelope"
 	"go.uber.org/thriftrw/internal/multiplex"
+	"go.uber.org/thriftrw/internal/semver"
 	"go.uber.org/thriftrw/plugin/api"
 	"go.uber.org/thriftrw/protocol"
+	"go.uber.org/thriftrw/version"
 
 	"go.uber.org/atomic"
 )
 
-var _proto = protocol.Binary
+var (
+	_proto      = protocol.Binary
+	compatRange = computeCompatibleRange()
+)
+
+func computeCompatibleRange() semver.Range {
+	v, err := semver.Parse(version.Version)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


internal/plugin/transport.go
```diff
@@ -70,7 +84,31 @@ func NewTransportHandle(name string, t envelope.Transport) (Handle, error) {
 	if handshake.APIVersion != api.APIVersion {
 		return nil, errHandshakeFailed{
 			Name:   name,
-			Reason: errVersionMismatch{Want: api.APIVersion, Got: handshake.APIVersion},
+			Reason: errAPIVersionMismatch{Want: api.APIVersion, Got: handshake.APIVersion},
```
bombela: Thanks to your PR description I understand that .Version is the version of the thriftrw library that was used during compilation of the plugin itself and .APIVersion is the plugin API version itself.

So  what is really the goal of checking both APIVersion and thriftrw Version? APIVersion should be sufficient isn't? As long as you speak the right API, the interpretation of it on both sides should stay the same no? And if something changes in the thriftrw API enough, it warrants a new plugin API, ensuring only updated plugins will work with the new library.

On the other hand since the plugin API will never change without changing the thriftrw version, maybe .APIVersion is not needed after all, as we can assume the thriftrw API includes the plugin API.

A the very least, maybe rename handshake.Version to handshake.ThriftRWVersion or .LibVersion, just to make it a tad easier to read.

abhinav: API Version is the Thrift IDL version. The intention is that you can use the handshake to negotiate breaking changes to the API. As long as we don't add new required fields to `HandshakeResponse`, we can make breaking changes to the rest of the IDL and bump the API_VERSION if necessary. API_VERSION will not change every time we make ThriftRW releases.

bombela: with our offline discussions, I get it now. Let's just rename .Version for better distinction.

PR #279, opened 2016-12-14
Fix lint

peter-edge:
This fixes make lint by fixing some errors in idl/parser_test.go, and adding compile/primitive.go to `LINT_EXCLUDE_EXTRAS` (there are a ton of uncommented public types that are not worth commenting for now). It also uncomments `make lint_ci` in .travis.yml, as this appears to work again (let's see what travis says).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,7 @@ install:
     - make install_ci
 
 script:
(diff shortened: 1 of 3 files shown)


peter-edge:
Travis build is green, I should be whitelisted for CLA as of yesterday, will check again

PR #280, opened 2016-12-14
make lint: Add errcheck and fix errors

peter-edge:
This PR adds errcheck as a linter used, and fixes the code to handle any
issues found by it.

-   I did not include `-blank` as a default flag to errcheck, I actually like
    being able to do this in completely obvious places. There are a few
    functions here that I use blank errors for, because it is known no error
    will be returned (and a comment is added to that effect)
-   I changed `main` to call a function `do` that returns an error. This is
    because with how the code was written with `log.Fatalf` killing the
    program, the `Close` on `pluginHandle` would not be called in error cases,
    so now the `log.Fatalf` call is at the top level.

Note this will not pass until #279 is merged, and then this is rebased.

### DIFF
--- a/internal/frame/server.go
+++ b/internal/frame/server.go
@@ -25,6 +25,7 @@ import (
 	"io"
 
 	"go.uber.org/atomic"
+	"go.uber.org/thriftrw/internal"
 )
 
 // Handler handles incoming framed requests.
@@ -59,24 +60,13 @@ func NewServer(r io.Reader, w io.Writer) *Server {
 // there is an IO error or an unhandled error is received from the Handler.
 //
 // This blocks until the server is stopped using Stop.
-func (s *Server) Serve(h Handler) (retErr error) {
+func (s *Server) Serve(h Handler) (err error) {
 	if s.running.Swap(true) {
 		return fmt.Errorf("server is already running")
 	}
 
 	defer func() {
-		if err := s.r.Close(); err != nil && retErr == nil {
-			// TODO: by returning this error, internal/plugin/flag_test.go fails
-			// we should be able to handle this error appropriately
-			//retErr = err
-		}
-	}()
-	defer func() {
-		if err := s.w.Close(); err != nil && retErr == nil {
-			// TODO: by returning this error, internal/plugin/flag_test.go fails
-			// we should be able to handle this error appropriately
-			//retErr = err
-		}
+		err = internal.CombineErrors(err, s.r.Close(), s.w.Close())
 	}()
 
 	for s.running.Load() {
--- a/internal/frame/server.go
+++ b/internal/frame/server.go
@@ -59,13 +59,21 @@ func NewServer(r io.Reader, w io.Writer) *Server {
 // there is an IO error or an unhandled error is received from the Handler.
 //
 // This blocks until the server is stopped using Stop.
-func (s *Server) Serve(h Handler) error {
+func (s *Server) Serve(h Handler) (retErr error) {
 	if s.running.Swap(true) {
 		return fmt.Errorf("server is already running")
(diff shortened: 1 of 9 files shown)


peter-edge:
Note this WILL pass travis, since `make lint_ci` is not currently called from dev. I pushed a fix, and also commented out some error handling that causes an unwanted non-zero exit status for now.

internal/frame/server.go
```diff
@@ -59,13 +59,25 @@ func NewServer(r io.Reader, w io.Writer) *Server {
...
 	if s.running.Swap(true) {
 		return fmt.Errorf("server is already running")
 	}
 
-	defer s.r.Close()
-	defer s.w.Close()
+	defer func() {
+		if err := s.r.Close(); err != nil && retErr == nil {
+			// TODO: by returning this error, internal/plugin/flag_test.go fails
+			// we should be able to handle this error appropriately
```
abhinav: Actually, you should be able to do,

```go
import "go.uber.org/thriftrw/internal"

defer func() {
    retErr = internal.CombineErrors(
        retErr,
        s.r.Close(),
        s.w.Close(),
    )
}
```

(At which point you can call it `err`.)

main.go
```diff
@@ -68,25 +78,25 @@ func main() {
...
 	if opts.DisplayVersion {
 		fmt.Printf("thriftrw v%s\n", version.Version)
-		os.Exit(0)
+		return nil
 	}
 
 	if len(args) != 1 {
 		parser.WriteHelp(os.Stdout)
-		os.Exit(1)
+		return errors.New("") // will return non-nil error but print nothing
```
abhinav: This feels a bit hacky. I would rather call `WriteHelp` on a `bytes.Buffer`
and return that as the error.

```
var buff bytes.Buffer
parser.WriteHelp(&buff)
return errors.New(buff.String())
```

internal/frame/server.go
abhinav: Also, can you let this fail in Travis. I'd like to see how this fails so that
we can fix it.

main.go
```diff
@@ -127,25 +137,30 @@ func main() {
...
 
 	if gopts.GeneratePluginAPI {
 		pluginHandle = append(pluginHandle, pluginapigen.Handle)
 	}
 
-	defer pluginHandle.Close()
+	defer func() {
+		// TODO: should this be an error that results in a non-zero exit status?
+		if err := pluginHandle.Close(); err != nil {
+			log.Println(err.Error())
```
abhinav: Yeah, probably unless we find a strong reason to ignore this.

internal/frame/server.go
peter-edge: Done

main.go
peter-edge: Done

main.go
peter-edge: Changed

peter-edge:
Comments should be addressed

main.go
```diff
@@ -68,25 +80,26 @@ func main() {
...
-		os.Exit(0)
+		return nil
 	}
 
 	if len(args) != 1 {
-		parser.WriteHelp(os.Stdout)
-		os.Exit(1)
+		var buffer bytes.Buffer
+		parser.WriteHelp(&buffer)
+		return errors.New(buffer.String())
```
peter-edge: Note that this will now print to stderr, not stdout

abhinav:
I found why the test was failing: `r.Close()` was getting called twice in
`internal/frame/server.go`. I changed the reader and writer to support calling
Close multiple times.

peter-edge:
OK, then we're good to go I think!

PR #281, opened 2016-12-14
Add --no-{types,constants,service-helpers} flags

peter-edge:
This adds flags to disable code generation for types, constants and
service helpers defined in Thrift files. This is useful in situations
where you don't actually want or need the extra Go code, and just want
to use ThriftRW's plugin interface.

### DIFF
--- a/gen/generate.go
+++ b/gen/generate.go
@@ -62,8 +62,14 @@ type Options struct {
 	// Code generation plugin
 	Plugin plugin.Handle
 
-	// Only run the plugin, do not generate the standard golang code.
-	PluginOnly bool
+	// Do not generate types.go
+	NoTypes bool
+
+	// Do not generate constants.go
+	NoConstants bool
+
+	// Do not generate service helpers
+	NoServiceHelpers bool
 }
 
 // Generate generates code based on the given options.
@@ -117,19 +123,14 @@ func Generate(m *compile.Module, o *Options) error {
 	if plug == nil {
 		plug = plugin.EmptyHandle
 	}
-	var pluginFiles map[string][]byte
+
 	if sgen := plug.ServiceGenerator(); sgen != nil {
 		res, err := sgen.Generate(genBuilder.Build())
 		if err != nil {
 			return err
 		}
-		pluginFiles = res.Files
-	}
 
-	if o.PluginOnly {
-		files = pluginFiles
-	} else if pluginFiles != nil {
-		if err := mergeFiles(files, pluginFiles); err != nil {
+		if err := mergeFiles(files, res.Files); err != nil {
 			return err
 		}
 	}
@@ -242,7 +243,9 @@ func generateModule(m *compile.Module, i thriftPackageImporter, builder *generat
 		}
 
 		// TODO(abg): Verify no file collisions
-		files["constants.go"] = buff.Bytes()
+		if !o.NoConstants {
+			files["constants.go"] = buff.Bytes()
+		}
 	}
 
 	if len(m.Types) > 0 {
@@ -259,7 +262,9 @@ func generateModule(m *compile.Module, i thriftPackageImporter, builder *generat
 		}
 
 		// TODO(abg): Verify no file collisions
-		files["types.go"] = buff.Bytes()
+		if !o.NoTypes {
+			files["types.go"] = buff.Bytes()
+		}
 	}
 
 	// Services must be generated last because names of user-defined types take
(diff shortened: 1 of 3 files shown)


abhinav:
I'm not sure we should do this this way. If we want the tool to be more
generic, I think the way forward is to treat the existing code generation like
a plugin that is enabled by default and opt-out of parts of it explicitly.
Otherwise the meaning of "plugin only" is broken if we add support for plugins
that integrate with existing codegen. For example, a struct tag plugin would
be called by the Go codegen logic while generating struct definitions. At that
point, plugin-only loses meaning because you can't generate code for the tag
plugin without the regular Go codegen.

Instead, it might make sense to have the following flags: `--no-types`,
`--no-constants`, and `--no-service-helpers`. I can see scenarios where we
want to disable these components to let a different plugin handle that
specific responsibility. Also, since the service helpers depend on the types,
`--no-service-helpers` would also set `--no-types`.

peter-edge:
I can buy that, I'll make the change shortly (or tomorrow).

peter-edge:
I think you mean `--no-types` sets `--no-service-helpers`. I updated the diff to use these flags, so this should be all set.

gen/generate.go
```diff
@@ -234,7 +243,9 @@ func generateModule(m *compile.Module, i thriftPackageImporter, builder *generat
 		}
 
 		// TODO(abg): Verify no file collisions
-		files["constants.go"] = buff.Bytes()
+		if !o.NoConstants {
+			files["constants.go"] = buff.Bytes()
```
abhinav: Maybe do this in the `if` check above:

    if !o.NoConstants && len(m.Constants) > 0 {
      ..

gen/generate.go
```diff
@@ -251,7 +262,9 @@ func generateModule(m *compile.Module, i thriftPackageImporter, builder *generat
 		}
 
 		// TODO(abg): Verify no file collisions
-		files["types.go"] = buff.Bytes()
+		if !o.NoTypes {
```
abhinav: same as above

gen/generate.go
```diff
@@ -278,8 +291,10 @@ func generateModule(m *compile.Module, i thriftPackageImporter, builder *generat
 					serviceName, err)
 			}
 
-			for name, buff := range serviceFiles {
-				files[name] = buff.Bytes()
+			if !o.NoServiceHelpers {
```
abhinav: ditto

main.go
```diff
@@ -52,6 +52,9 @@ type genOptions struct {
 
 	GeneratePluginAPI bool `long:"generate-plugin-api" hidden:"true" description:"Generates code for the plugin API"`
 	NoVersionCheck    bool `long:"no-version-check" hidden:"true" description:"Does not add library version checks to generated code."`
+	NoTypes           bool `long:"no-types" description:"Do not generate types.go, implies --no-service-helpers."`
```
abhinav: Use "Do not generate code for types" instead

main.go
```diff
@@ -52,6 +52,9 @@ type genOptions struct {
...
+	NoConstants       bool `long:"no-constants" description:"Do not generate constants.go."`
```
abhinav: Do not generate code for `const` declarations

main.go
```diff
@@ -52,6 +52,9 @@ type genOptions struct {
...
+	NoServiceHelpers  bool `long:"no-service-helpers" description:"Do not generate service helper go files."`
```
abhinav: Do not generate service helpers.

abhinav:
> I think you meant --no-types sets --no-service-helpers. I updated the diff to use these flags, so this should be all set.

Oh right. Yeah, thanks.

gen/generate.go
peter-edge: I didn't want to do this because if you do this for services (below), the GenerateServiceRequest will not be populated properly. builder.AddRootService has to be called, so I have to do that logic regardless, and then I did the same logic here to be consistent and in case there are side effects from functions called inside the if statement (I think in the case of types and constants, there aren't, but this needs to be guaranteed by the documentation of functions called).

gen/generate.go
peter-edge: see reply

main.go
peter-edge: Done

abhinav:
Please add an entry to the changelog and I will merge this

peter-edge:
Done

Issue #230, opened 2016-09-19
Plugins: Struct tags


abhinav:
We should allow customization of struct tags through plugins. The plugin would probably need access to *some* information about the field for which the tag is being generated. We could provide built-in plugins for JSON/YAML tags.

Motiejus:
cc @edast

We would appreciate this feature for our GraphQL endpoints, which could use thriftrw-based schemas.

I.e. in some GraphQL internal services we emit JSON. Currently, that JSON is not typed. We would like to specify the returned JSON with Thrift. However, the rules for JSON field names differ for GraphQL and JSON.

With this feature, our returned JSON could be specified in Thrift. If you make this happen or prioritize, please let us know.

martynasb:
+1

peter-edge:
+1 to this too.

Short: I started writing a plugin and then realized I can't get information on all typedefs/structs/etc, just on arguments to service methods.

Long: I'm sure there's a design decision around it, but why not just send over all information about a module to a plugin, as opposed to having a feature set (which is just now SERVICE_GENERATOR)? ie why not send over information on all the typedefs/structs/etc defined in the module, why is this limited to services (which have struct information as it is, for arguments)?

protobuf has been doing this for some time via https://github.com/google/protobuf/blob/9d709f446f18b672edff81ee2d3f9d6516465d8c/src/google/protobuf/descriptor.proto#L55 that is effectively sent to plugins, and then a plugin doesn't need to be denoted as a "service plugin" etc. I think the idea is you want the ability to directly affect struct tags on the code that is normally generated, but this can get dangerous I would think.

In any case, just adding information on structs defined in a module would be super useful.

peter-edge:
I think basically, why can't we just make Module (and all fields, recursively) https://github.com/thriftrw/thriftrw-go/blob/dev/compile/module.go a thrift struct, and send this to the plugin? Or could we? This might be a dupe of https://github.com/thriftrw/thriftrw-go/issues/231

peter-edge:
Actually I think I mean https://github.com/thriftrw/thriftrw-go/blob/dev/ast/program.go

abhinav:
@peter-edge plugins that have access to all the information from the Thrift
files is something we can eventually support. Please create a separate issue
for that. Note that depending on project priorities, we may not be able to get
to that soon.

PR #282, opened 2016-12-14
Expand plugin api

peter-edge:
This is in no way ready to merge, does not have additional testing, and frankly isn't the best code, but I wanted to put this up there just to start the conversation.

This expands the plugin API to contain type information for structs, exceptions, unions, and enums. Structs and enums are stored in a new map on GenerateServiceRequest, and IDs are stored within a Type, and TypeReferences now contain a Type.

The information on the Struct and Enum type is not complete (annotations, default values, etc). Optimally, I'd like to version api.thrift and rethink Type completely, but this was done to be backwards compatible. Struct needs to be stored by ID in case of a recursive struct definition, and Enum is stored by ID for consistency.

The idea here is that with this info, we could then next move to the TAGGER plugin feature. But again, this code needs to be rethought.

So...thoughts?


peter-edge:
Note that this PR only will add structs/enums to the Structs and Enums maps if they are requests, responses, or exceptions on service methods, or are fields of such requests, responses, or exceptions.

abhinav:
Thanks for the PR! I think this functionality should be exposed in a new
plugin feature. I can't think of a good name for the feature. "Generator",
maybe? Implying that it's a full generator rather than a service or tag
generator.

peter-edge:
I just added what I really want to do to `plugin/newapi/newapi.thrift` It's minimally commented, and some stuff still needs to be thought out (mostly how to identify the names of modules and the "base package", especially in a language indpendent way), but it's mostly self-explanatory.

This is the premise: What is now considered `GenerateService` isn't really service specific at all. The current API assumes what data is actually needed for a service plugin, when I think we should expose the thrift definition as much as possible for any plugin type. The return value is just a list of file paths to data, which isn't service specific either. Something I would think of as service specific would be along the lines of "generate a method interceptor", where a plugin could add a generated method interceptor for requests and responses. A go struct tagging plugin might also be a case, where we say "add or remove these tags from struct fields in the generated golang code".

What I've added is a generic way to represent the thrift definition, which is passed to every plugin request, regardless of feature. What is now `GenerateService` is just `Generate`, and I also added what I think the definition could be for a `GoStructTagger` plugin request and response. The thrift definition also does not do anything golang-specific - we don't know what plugins want to do, so we should just be exposing what the actual thrift definition is. When it comes to functionality that helps golang-specific plugins, I think we should expose this in a `plugins/helpers` package, ie the conversion from thrift type to golang type, whether or not a field is a pointer, etc...this should all be in a helper package that plugins can include.

I would say that if this was a direction you were interested in, we could:

1. Add this as the "new" api, and not touch the existing api code, first just the `Generate` plugin feature, then the `GoStructTagger` feature.
2. Add a message that is printed to stderr for existing api users that the existing api will eventually be deprecated (ie yarpc can push off moving to this).
3. Eventually remove the existing api code, so that we don't have to edit the plugin api in two places.

Thoughts? `plugins/newapi/newapi.thrift` is just a prototype I wrote up, probably needs some work (and I added comments where I have some questions that I need to answer/get answered).

https://github.com/thriftrw/thriftrw-go/issues/230

PR #285, opened 2017-02-03
ast: Add Visitor and Walk to walk the AST

abhinav:
This adds a Visitor interface and a Walk function which walks the AST using
the visitor. This is very similar to the standard library's go/ast Visitor
interface and Walk functions except we provide a Walker object which gives
access to the list of ancestors of a node.

- [x] Implementation
- [x] Tests
    - [x] Annotation
    - [x] BaseType
    - [x] Constant
    - [x] ConstantBoolean
    - [x] ConstantDouble
    - [x] ConstantInteger
    - [x] ConstantList
    - [x] ConstantMap
    - [x] ConstantMapItem
    - [x] ConstantReference
    - [x] ConstantString
    - [x] Enum
    - [x] EnumItem
    - [x] Field
    - [x] Function
    - [x] Include
    - [x] ListType
    - [x] MapType
    - [x] Namespace
    - [x] Program
    - [x] Service
    - [x] SetType
    - [x] Struct
    - [x] TypeReference
    - [x] Typedef


ast/visitor.go
```diff
@@ -0,0 +1,64 @@
...
+type VisitorFunc func(Walker, Node)
+
+// Visit the given node and its descendants.
+func (f VisitorFunc) Visit(w Walker, n Node) Visitor {
+	f(w, n)
+	return f
+}
+
+// CombineVisitors merges the given visitors into a single Visitor.
+func CombineVisitors(visitors ...Visitor) Visitor {
```
prashantv: nit: maybe `MultiVisitor` to match the naming style of `io.MultiWriter`?

ast/walk.go
```diff
@@ -0,0 +1,128 @@
...
+
+// nodeStack of nodes visited in the order they were visited
+type nodeStack []Node
+
+func (ss nodeStack) Ancestors() []Node {
+	if len(ss) == 0 {
+		return nil
+	}
+
+	ancestors := make([]Node, len(ss))
```
prashantv: Do you think having ancestors be in most recent order going to be very helpful? I'd be OK with it being top-down order, but if we do want to keep most recent order, we should probably just store them in that order to begin with? Or maybe have `nodeStack` be a struct and keep a lazily initialized reversed list?

ast/walk.go
```diff
@@ -0,0 +1,128 @@
...
+	for i, n := range ss {
+		ancestors[len(ss)-1-i] = n
+	}
+	return ancestors
+}
+
+func walk(stack nodeStack, v Visitor, node Node) {
+	if v = v.Visit(stack, node); v == nil {
```
prashantv: Since we're not scoping `v` to the `if`, inlining doesn't buy us much, so might be clearer to expand
```go
v = v.Visit(..)
if v == nil {
  return
}
```

ast/walk.go
```diff
@@ -0,0 +1,128 @@
...
+	case SetType:
+		walk(stack, v, n.ValueType)
+		walkAnnotations(stack, v, n.Annotations)
+	case *Struct:
+		walkFields(stack, v, n.Fields)
+		walkAnnotations(stack, v, n.Annotations)
+	case *Typedef:
+		walk(stack, v, n.Type)
+		walkAnnotations(stack, v, n.Annotations)
+	}
```
prashantv: can you put a `default` case with a `panic`? And then we can have randomly generated programs (or maybe just any of the test thrift files) run through `Walk` to ensure we didn't miss anything

ast/walk.go
```diff
@@ -0,0 +1,128 @@
...
+func walk(stack nodeStack, v Visitor, node Node) {
+	if v = v.Visit(stack, node); v == nil {
+		return
+	}
+
+	// After visiting the node, if we still have a non-nil visitor, we need to
+	// visit its children recursively.
+
+	stack = append(stack, node)
+	switch n := node.(type) {
```
prashantv: Is it worth moving this logic out of this method and into the `Node` interface? 

I think it would require refactoring things a little,
```go
type internalVisitor interface {
  visit(nodeStack, Node)
}

type Node interface {
  walk(v Visitor)
}

func (c *Constant) walk(s nodeStack, v internalVisitor) {
  v.visit(s, c.Type)
  v.visit(s, c.Value)
}

type walker struct{
  userVisitor Visitor
}

func (w *walker) visit(s nodeStack, n Node) {
  v := w.userVisitor.Visit(s, n)
  if v == nil {
    return
  }

  s = append(s, n)
  v.walk(s, v)
}

```

This way the type system will ensure that `Walk` handles every type, and the implementation for walking each type is part of the type's methods.

ast/walk.go
abhinav: It doesn't have to be in this order but I like the idea that you're zooming
out as you're walking further up the list. I don't mind storing the list
either way.

Do you want to store it in the same order to avoid copying? I want to make the
copy regardless of how it is stored so that callers of `Ancestors()` cannot
actually modify the list. Any modifications made to it while visiting a node
affect the siblings of that node.

ast/walk.go
abhinav: I considered doing that. A disadvantage of that is that the common behavior
around walker state management now has to be repeated in every node. Every
node has to have behavior that visits it, ensures that the returned visitor is
non-nil before visiting its children, remembering to append itself to the node
stack. Having everything in a single function makes this state management
significantly cleaner at the cost of slightly reduced type safety.

The nil check and the append to stack are done exactly once to visit the
"current" node and everything after that is only to visit node children.

ast/walk.go
abhinav: This wasn't previously matching all Node types; only those with children. But
it's probably safer to match everything and panic otherwise. Updating.

ast/walk.go
prashantv: In my above code example, `walker` would only be written once, and handles the `nil` visitor logic, appending etc. Each type only needs to implement the `walk` method, which calls `v.visit` on all "children" of that node. The same `walker` is passed regardless of the node type.

ast/walk.go
prashantv: Yeah I was hoping to avoid the copy -- I really wish there was a way to pass immutable slices. Let's keep this as-is, we can optimize later if needed.

ast/walk.go
abhinav: I made this change but instead of `internalVistior`, etc. I just added the following function to the Node interface:

    forEachChild(func(Node))

Based on that, `walker` as suggested above is quite straightforward to implement and the node implementations don't need to know much about how `ast.Walk` works.

abhinav:
Tests for all node types have been added. This is no longer WIP.

CC @prashantv @peter-edge

ast/visitor.go
```diff
@@ -0,0 +1,68 @@
...
+// THE SOFTWARE.
+
+package ast
+
+// Walker provides acccess to information about the state of the AST walker.
+type Walker interface {
+	// Ancestors returns a copy of a slice containing all the ancestor nodes
+	// of the current node. The first node in the slice is the immediate
+	// parent of the current node, the node after that its parent, and so on.
+	Ancestors() []Node
```
peter-edge: Is this a golang thing? I would have thought to call this `Children()`.

ast/walk.go
```diff
@@ -0,0 +1,71 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package ast
+
+// Walk walks the AST depth-first with the given visitor, starting at the
+// given node. The visitor's Visit function should return a non-nil visitor if
+// it wants to visit the children of the node it was called with.
+func Walk(v Visitor, n Node) {
+	// walk(nil, v, n)
```
peter-edge: remove

ast/mock_visitor_test.go
```diff
@@ -0,0 +1,39 @@
+// Automatically generated by MockGen. DO NOT EDIT!
```
peter-edge: How was this generated? can it be added to some script/Makefile target?

ast/walk.go
```diff
@@ -0,0 +1,71 @@
...
+		return
+	}
+
+	w.v = w.v.Visit(w.s, n)
+	if w.v == nil {
+		return
+	}
+
+	w.s = append(w.s, n)
+	n.forEachChild(w.Walk)
```
prashantv: There's a combination of a couple of things that we may want to rethink:
 - the `w.s = append(w.s, n)` is a little subtle, relying on the fact that we are using a value for passing around the `walker`, and that the slice modifications are always appends. If we keep this, we should probably add a comment, since making this `w *walker` would break.
 - we're rebinding `w.Walk` for each node, costing an extra allocating.

I think passing the `nodeStack` explicitly will allow us to save some allocations, and be more explicit.

ast/visitor.go
abhinav: Children would be the opposite direction. This returns the stack of parent nodes of the node being visited, not the child nodes.

ast/walk.go
abhinav: derp

ast/mock_visitor_test.go
abhinav: There's a `go:generate` declaration in ast/visitor.go. This repository doesn't
use a YARPC-like generate.sh.

peter-edge:
For my side, I'm fine with this

ast/walk.go
```diff
@@ -0,0 +1,71 @@
...
+
+	// Note that visitorWrapper is passed by value so we're operating on a
+	// copy of visitorWrapper.
+	w.v = w.v.Visit(ss, n)
+	if w.v == nil {
+		return
+	}
+
+	ss = append(ss, n)
+	n.visitChildren(ss, w)
```
prashantv: since `visitorWrapper` is a value, converting to a `visitor` (an interface) still costs an allocation.

should we have the `visitChildren` just refer `visitorWrapper` directly?

abhinav: I considered that but decided against it because I thought we might want to
call visitChildren with other implementations, but turns out we can always do
that later if necessary since this is internal.

PR #286, opened 2017-02-04
gen: Add Equals() to all datatypes

harshuber:
This adds a Equals function generator for all types which knows how to compare the equality of two Thrift types and their Value representations.

### DIFF
--- a/gen/collision_test.go
+++ b/gen/collision_test.go
@@ -101,6 +101,331 @@ func TestStruct(t *testing.T) {
 	}
 }
 
+func TestStructEquals(t *testing.T) {
+	tests := []struct {
+		desc     string
+		lhs, rhs interface{}
+		want     bool
+	}{
+		{
+			"StructCollision equal",
+			&tc.StructCollision{
+				CollisionField: true,
+			},
+			&tc.StructCollision{
+				CollisionField: true,
+			},
+			true,
+		},
+		{
+			"StructCollision unequal",
+			&tc.StructCollision{
+				CollisionField:  true,
+				CollisionField2: "true",
+			},
+			&tc.StructCollision{
+				CollisionField:  true,
+				CollisionField2: "false",
+			},
+			false,
+		},
+		{
+			"Empty StructCollision equal",
+			&tc.StructCollision{},
+			&tc.StructCollision{},
+			true,
+		},
+		{
+			"Empty StructCollision unequal",
+			&tc.StructCollision{},
+			&tc.StructCollision{
+				CollisionField: true,
+			},
+			false,
+		},
+		{
+			"StructCollision2 (struct_collision) equal",
+			&tc.StructCollision2{
+				CollisionField:  true,
+				CollisionField2: "such trueness",
+			},
+			&tc.StructCollision2{
+				CollisionField:  true,
+				CollisionField2: "such trueness",
+			},
+			true,
+		},
+		{
+			"StructCollision2 (struct_collision) unequal",
+			&tc.StructCollision2{
+				CollisionField:  false,
+				CollisionField2: "such trueness",
+			},
+			&tc.StructCollision2{
+				CollisionField:  true,
+				CollisionField2: "such trueness",
+			},
+			false,
+		},
+		{
+			"Empty StructCollision2 (struct_collision) equal",
+			&tc.StructCollision2{},
+			&tc.StructCollision2{},
+			true,
+		},
+		{
+			"Empty StructCollision2 (struct_collision) unequal",
+			&tc.StructCollision2{},
+			&tc.StructCollision2{
+				CollisionField: true,
+			},
+			false,
+		},
+		{
+			"PrimitiveContainers equal",
+			&tc.PrimitiveContainers{
+				A: []string{"arbre", "fleur"},
+				B: map[string]struct{}{
+					"croissant": {},
+					"baguette":  {},
+				},
+				C: map[string]string{
+					"voiture": "bleue",
+					"cammion": "rouge",
+				},
+			},
+			&tc.PrimitiveContainers{
+				A: []string{"arbre", "fleur"},
+				B: map[string]struct{}{
+					"croissant": {},
+					"baguette":  {},
+				},
+				C: map[string]string{
+					"voiture": "bleue",
+					"cammion": "rouge",
+				},
+			},
+			true,
+		},
+		{
+			"PrimitiveContainers unequal",
+			&tc.PrimitiveContainers{
+				A: []string{"arbre", "fleur"},
+				B: map[string]struct{}{
+					"croissant": {},
+					"baguette":  {},
+				},
+				C: map[string]string{
+					"voiture": "bleue",
+					"cammion": "rouge",
+				},
+			},
+			&tc.PrimitiveContainers{
+				A: []string{"arbre", "fleur"},
+				B: map[string]struct{}{
+					"croissant": {}, // No baguette
+				},
+				C: map[string]string{
+					"voiture": "bleue",
+					"cammion": "rouge",
+				},
+			},
+			false,
+		},
+		{
+			"Empty PrimitiveContainers equal",
+			&tc.PrimitiveContainers{},
+			&tc.PrimitiveContainers{},
+			true,
+		},
+		{
+			"Empty PrimitiveContainers unequal",
+			&tc.PrimitiveContainers{},
+			&tc.PrimitiveContainers{
+				A: []string{"arbre", "fleur"},
+			},
+			false,
+		},
+		{
+			"StructConstant (struct_constant) equal",
+			tc.StructConstant,
+			&tc.StructCollision2{
+				CollisionField2: "false indeed",
+			},
+			true,
+		},
+		{
+			"StructConstant (struct_constant) unequal",
+			tc.StructConstant,
+			&tc.StructCollision2{
+				CollisionField2: "true indeed",
+			},
+			false,
+		},
+		{
+			"UnionCollision .CollisionField equal",
+			&tc.UnionCollision{
+				CollisionField: ptr.Bool(true),
+			},
+			&tc.UnionCollision{
+				CollisionField: ptr.Bool(true),
+			},
+			true,
+		},
+		{
+			"UnionCollision .collision_field2 equal",
+			&tc.UnionCollision{
+				CollisionField2: ptr.String("so true"),
+			},
+			&tc.UnionCollision{
+				CollisionField2: ptr.String("so true"),
+			},
+			true,
+		},
+		{
+			"UnionCollision .CollisionField unequal",
+			&tc.UnionCollision{
+				CollisionField: ptr.Bool(true),
+			},
+			&tc.UnionCollision{
+				CollisionField: ptr.Bool(false),
+			},
+			false,
+		},
+		{
+			"UnionCollision .collision_field2 unequal",
+			&tc.UnionCollision{
+				CollisionField2: ptr.String("so true"),
+			},
+			&tc.UnionCollision{
+				CollisionField2: ptr.String("not so true"),
+			},
+			false,
+		},
+		{
+			"UnionCollision equal",
+			&tc.UnionCollision{
+				CollisionField:  ptr.Bool(true),
+				CollisionField2: ptr.String("so true"),
+			},
+			&tc.UnionCollision{
+				CollisionField:  ptr.Bool(true),
+				CollisionField2: ptr.String("so true"),
+			},
+			true,
+		},
+		{
+			"UnionCollision unequal",
+			&tc.UnionCollision{
+				CollisionField:  ptr.Bool(true),
+				CollisionField2: ptr.String("so true"),
+			},
+			&tc.UnionCollision{
+				CollisionField:  ptr.Bool(true),
+				CollisionField2: ptr.String("not so true"),
+			},
+			false,
+		},
+		{
+			"Empty UnionCollision equal",
+			&tc.UnionCollision{},
+			&tc.UnionCollision{},
+			true,
+		},
+		{
+			"Empty UnionCollision unequal",
+			&tc.UnionCollision{},
+			&tc.UnionCollision{
+				CollisionField: ptr.Bool(false),
+			},
+			false,
+		},
+		{
+			"UnionCollision2 (union_collision) .CollisionField equal",
+			&tc.UnionCollision2{
+				CollisionField: &[]bool{true}[0],
+			},
+			&tc.UnionCollision2{
+				CollisionField: &[]bool{true}[0],
+			},
+			true,
+		},
+		{
+			"UnionCollision (union_collision) .collision_field2 equal",
+			&tc.UnionCollision2{
+				CollisionField2: ptr.String("true indeed"),
+			},
+			&tc.UnionCollision2{
+				CollisionField2: ptr.String("true indeed"),
+			},
+			true,
+		},
+		{
+			"UnionCollision2 (union_collision) .CollisionField unequal",
+			&tc.UnionCollision2{
+				CollisionField: &[]bool{false}[0],
+			},
+			&tc.UnionCollision2{
+				CollisionField: &[]bool{true}[0],
+			},
+			false,
+		},
+		{
+			"UnionCollision (union_collision) .collision_field2 unequal",
+			&tc.UnionCollision2{
+				CollisionField2: ptr.String("true indeed"),
+			},
+			&tc.UnionCollision2{
+				CollisionField2: ptr.String("false indeed"),
+			},
+			false,
+		},
+		{
+			"UnionCollision (union_collision) equal",
+			&tc.UnionCollision2{
+				CollisionField:  &[]bool{false}[0],
+				CollisionField2: ptr.String("true indeed"),
+			},
+			&tc.UnionCollision2{
+				CollisionField:  &[]bool{false}[0],
+				CollisionField2: ptr.String("true indeed"),
+			},
+			true,
+		},
+		{
+			"UnionCollision (union_collision) unequal",
+			&tc.UnionCollision2{
+				CollisionField:  &[]bool{true}[0],
+				CollisionField2: ptr.String("true indeed"),
+			},
+			&tc.UnionCollision2{
+				CollisionField:  &[]bool{false}[0],
+				CollisionField2: ptr.String("true indeed"),
+			},
+			false,
+		},
+		{
+			"Empty UnionCollision (union_collision) equal",
+			&tc.UnionCollision2{},
+			&tc.UnionCollision2{},
+			true,
+		},
+		{
+			"Empty UnionCollision (union_collision) unequal",
+			&tc.UnionCollision{},
+			&tc.UnionCollision{
+				CollisionField: &[]bool{true}[0],
+			},
+			false,
+		},
+	}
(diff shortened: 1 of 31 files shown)


gen/field.go
```diff
@@ -352,3 +357,38 @@ func (f fieldGroupGenerator) String(g Generator) error {
...
+				<if .Required>
+					if !(<equals .Type $lhsField $rhsField>) {
+						return false
+					}
+				<else>
+					<if not (isPrimitiveType .Type)>
+						if (<$lhsField> == nil && <$rhsField> != nil) ||
+								(<$lhsField> != nil && <$rhsField> == nil) {
+							return false
+						} else if <$lhsField> != nil && <$rhsField> != nil {
```
abhinav: you can do this check in equalsPtr

gen/generator.go
```diff
@@ -342,6 +345,14 @@ func (g *generator) recordGenDeclNames(ignoreConflicts bool, d *ast.GenDecl) err
 // itself is not a reference type.
 //
 // 	<typeReferencePtr $someType>
+//
+//  equals(TypeSpec, lhs, rhs): Compares lhs and rhs of given TypeSpec for equality.
+//
+//  <equals $someType $lhs $rhs>
```
abhinav: clarify that an expression of type bool is returned

gen/list.go
```diff
@@ -27,6 +27,7 @@ import "go.uber.org/thriftrw/compile"
 type listGenerator struct {
 	hasReaders
 	hasLazyLists
+	hasEquals
```
abhinav: So the whole `has*` pattern is a leftover from before `EnsureDeclared` existed. You can now just use that and drop this.

gen/list.go
```diff
@@ -147,3 +148,49 @@ func (l *listGenerator) Reader(g Generator, spec *compile.ListSpec) (string, err
 
 	return name, wrapGenerateError(spec.ThriftName(), err)
 }
+
+// Equals generates a function to compare lists of the given type
+//
+// func $name(lhs, rhs $listType) bool {
+//      ...
+// }
```
abhinav: Code examples should be formatted like so:

```
//<space><tab>func ...
```

gen/list.go
```diff
@@ -147,3 +148,49 @@ func (l *listGenerator) Reader(g Generator, spec *compile.ListSpec) (string, err
...
+	}
+
+	err := g.DeclareFromTemplate(
+		`
+            <$listType := typeReference .Spec>
+
+			<$lhs := newVar "lhs">
+			<$rhs := newVar "rhs">
+			<$i := newVar "i">
+			<$lv := newVar "lv">
```
abhinav: inconsistent indentation. make sure your editor is using tabs here.
enable visible whitespace to verify this.

gen/list.go
```diff
@@ -147,3 +148,49 @@ func (l *listGenerator) Reader(g Generator, spec *compile.ListSpec) (string, err
...
+		return name, nil
+	}
+
+	err := g.DeclareFromTemplate(
+		`
+            <$listType := typeReference .Spec>
+
+			<$lhs := newVar "lhs">
+			<$rhs := newVar "rhs">
+			<$i := newVar "i">
```
abhinav: don't newvar everything at the top. do it as close to where it is first used

gen/map.go
```diff
@@ -190,6 +191,83 @@ func (m *mapGenerator) Reader(g Generator, spec *compile.MapSpec) (string, error
...
+
+            <$lhs := newVar "lhs">
+            <$rhs := newVar "rhs">
+            <$i := newVar "i">
+            <$j := newVar "j">
+            <$lk := newVar "lk">
+            <$lv := newVar "lv">
+            <$rk := newVar "rk">
+            <$rv := newVar "rv">
+            <$ok := newVar "ok">
```
abhinav: same as above. newVar as close to where the value is first used

gen/map.go
```diff
@@ -190,6 +191,83 @@ func (m *mapGenerator) Reader(g Generator, spec *compile.MapSpec) (string, error
...
+					return false
+				}
+
+				<if isHashable .Spec.KeySpec>
+					for <$lk>, <$lv> := range <$lhs> {
+						<$rv>, <$ok> := <$rhs>[<$lk>]
+						if !<$ok> {
+							return false
+						}
+						if !(<equals .Spec.ValueSpec $lv $rv>) {
```
abhinav: Consider breaking out some of the logic into functions here. For example,

```
<if isHashable .Spec.KeySpec>
  ...
<else>
  <equalsForUnhashableKey ...>
<end>
```

Where you inject `equalsUnhashable` into *just* this template by using TemplateFunc, implementing the function itself as a method on `mapGenerator`

```
g.DeclareFromTemplate(
  ...,
  TemplateFunc("equalsForUnhashableKey", m.equalsForUnhashableKey)
)
```

```
func (m *mapGenerator) equalsForUnhashableKey(g Generator, ...) (string, error) {
  // ...
}
```

gen/map.go
```diff
@@ -190,6 +191,83 @@ func (m *mapGenerator) Reader(g Generator, spec *compile.MapSpec) (string, error
...
+						<$lv> := <$i>.Value
+						<$ok> := false
+						for _, <$j> := range <$rhs> {
+							<$rk> := <$j>.Key
+							<$rv> := <$j>.Value
+							if !(<equals .Spec.KeySpec $lk $rk>) {
+								continue
+							}
+							if !(<equals .Spec.ValueSpec $lv $rv>) {
+								continue
```
abhinav: can't we just return false if the key matches but the value doesn't? caveat being that behavior is undefined if the map contract (only one instance of each key) isn't satisfied.

gen/set.go
```diff
@@ -150,3 +151,72 @@ func (s *setGenerator) Reader(g Generator, spec *compile.SetSpec) (string, error
...
+					return false
+				}
+
+				// if the values in the set are hashable they can be used
+				// as keys in a map.
+				<if isHashable .Spec.ValueSpec>
+					<$o> := make(<$setType>, len(<$lhs>))
+					for <$x> := range <$lhs> {
+						<$o>[<$x>] = struct{}{}
+					}
```
abhinav: why are we making a copy of lhs? if the lengths of the two sets are same, all you have to check is that everything in rhs is in lhs.

gen/typedef_test.go
```diff
@@ -23,74 +23,87 @@ package gen
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
```
abhinav: import group ordering we follow is:

- standard library
- local library imports
- external library imports

gen/field.go
```diff
@@ -30,6 +30,7 @@ var reservedIdentifiers = map[string]struct{}{
 	"ToWire":   {},
 	"FromWire": {},
 	"String":   {},
+	"Equals":   {},
```
harshuber: Can you please point me where similar checks are done for toWire/fromWire. I can add code for equals too there.

gen/equals.go
```diff
@@ -0,0 +1,110 @@
...
+					// Make sure that both the pointers are non nil.
+					if <$lhs> != nil && <$rhs> != nil {
+						// Call Equals method after dereferencing the pointers
+						<$x> := *<$lhs>
+						<$y> := *<$rhs>
+						return <equals .Spec $x $y>
+					} else if <$lhs> == nil && <$rhs> == nil {
+						return true
+					} else {
+						return false
```
harshuber: The function helps to handle the case for enums, since it would generate a function that takes two Enum pointers and calls underlying equals method for enums after dereferencing. Else, it would try to call 
*<$lhs>.Equals(*<$rhs>) which causes "invalid indirect of ptr" error.

gen/field.go
abhinav: I don't think we have existing tests for this. Just create a struct somewhere
with a `equals` field in one of the testdata Thrift files.

gen/equals.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	"fmt"
+
+	"go.uber.org/thriftrw/compile"
+)
+
+// EqualsGenerator is responsible for generating code that knows how
+// to compare the equality of two Thrift types and their Value representations.
+type EqualsGenerator struct {
```
abhinav: Make this unexported. No reason to expose it.

APIs in `gen/` are supposed to be unstable anyway.

gen/equals.go
```diff
@@ -0,0 +1,108 @@
...
+				// Make sure that both the pointers are non nil.
+				if <$lhs> != nil && <$rhs> != nil {
+					// Call Equals method after dereferencing the pointers
+					<$x> := *<$lhs>
+					<$y> := *<$rhs>
+					return <equals .Spec $x $y>
+				} else if <$lhs> == nil && <$rhs> == nil {
+					return true
+				} else {
+					return false
```
abhinav: nit: Collapse this into

```
if  <$lhs> != nil && <$rhs> != nil {
  // ..
}
return <$lhs> == nil && <$rhs> == nil
```

gen/equals.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	equalsPtrFunc := fmt.Sprintf("_%v_EqualsPtr", spec.ThriftName())
+	err := g.EnsureDeclared(
+		`
+			<$type := typeReference .Spec>
+			<$lhs := newVar "lhs">
+			<$rhs := newVar "rhs">
+			<$x := newVar "x">
+			<$y := newVar "y">
```
abhinav: Declare x and y as close to where you first use them as possible

abhinav: Also, based on the usage, you can probably just drop the newVar calls:

```
<$x := printf "*%s" $lhs>
<$y := printf "*%s" $rhs>
```

gen/equals.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+}
+
+// EqualsPtr is the same as Equals expect `lhs` and `rhs` are expected to be a
+// reference to a value of the given type.
+func (e *EqualsGenerator) EqualsPtr(g Generator, spec compile.TypeSpec, lhs, rhs string) (string, error) {
+	if !isPrimitiveType(spec) {
+		// Everything else is a reference type that has a Equals method on it.
+		return g.TextTemplate(
+			`((<.LHS> == nil && <.RHS> == nil) || (<.LHS> != nil && <.RHS> != nil && <equals .Spec .LHS .RHS>))`,
```
abhinav: Why can't the primitive version be the same instead of relying on the EqualsPtr function?

You can do the exact same thing as here with the only difference being that you do `*<.LHS>` and `*<.RHS>` instead. 

```
struct {
    LHS, LHSDeref string
    RHS, RHSDeref string
    ...
}{
    ...
    LHSDeref: "*" + lhs,
    RHSDeref: "*" + rhs,
}
```

gen/map.go
```diff
@@ -190,6 +190,101 @@ func (m *mapGenerator) Reader(g Generator, spec *compile.MapSpec) (string, error
...
+			<$rhs := newVar "rhs">
+			func <.Name>(<$lhs>, <$rhs> <$mapType>) bool {
+				if len(<$lhs>) != len(<$rhs>) {
+					return false
+				}
+
+				<$lk := newVar "lk">
+				<$lv := newVar "lv">
+				<$rv := newVar "rv">
+				<$ok := newVar "ok">
```
abhinav: nit: You can actually move these into the `<if ..>` block.

That will also remove them once they drop out of scope.

gen/map.go
```diff
@@ -190,6 +190,101 @@ func (m *mapGenerator) Reader(g Generator, spec *compile.MapSpec) (string, error
...
+				<if isHashable .Spec.KeySpec>
+					for <$lk>, <$lv> := range <$lhs> {
+						<$rv>, <$ok> := <$rhs>[<$lk>]
+						if !<$ok> {
+							return false
```
abhinav: Given how significantly the bodies differ in the hashable versus unhashble case, consider doing this check outside the template and simply calling a different function.

```
if !isHashable(spec) {
  return m.EqualsUnhashble(g, spec)
}

err := g.TextFromTemplate(
  ...
)
```

gen/map.go
```diff
@@ -190,6 +190,101 @@ func (m *mapGenerator) Reader(g Generator, spec *compile.MapSpec) (string, error
...
+				for _, <$j> := range <.RHS> {
+					<$rk> := <$j>.Key
+					<$rv> := <$j>.Value
+					if !<equals .Spec.KeySpec $lk $rk> {
+						continue
+					}
+
+					if !<equals .Spec.ValueSpec $lv $rv> {
+						// Caveat: Behavior is undefined if the map contract
+						// (only one instance of each key) isn't satisfied.
```
abhinav: Nit: Replace with

> Behavior is undefined if there are multiple entries with the same key.

gen/typedef.go
```diff
@@ -94,6 +94,16 @@ func typedef(g Generator, spec *compile.TypedefSpec) error {
...
 			<end>
 		}
+
+		<$lhs := newVar "lhs">
+		<$rhs := newVar "rhs">
+		func (<$lhs> <$typedefType>) Equals(<$rhs> <$typedefType>) bool {
+			<if isStructType .>
+				return (<typeReference .Target>)(<$lhs>).Equals((<typeReference .Target>)(<$rhs>))
+			<else>
+				return <equals .Target $lhs $rhs>
```
abhinav: Does the following not cover all cases?

```
<$l> := (<typeReference .Target>)(<$lhs>)
<$r> := (<typeReference .Target>)(<$rhs>)
return <equals .Target $l $r>
```

gen/equals.go
harshuber: This won't work for Enums as it would generate something like
```
*<$lhs>.Equals(*<$rhs>) which causes "invalid indirect of ptr" error.
```

gen/field.go
harshuber: Discussed offline with @abhinav. This would get rejected at the parser level itself.

harshuber:
Thanks for taking a look. Pushed another commit addressing your comments.

gen/container_test.go
```diff
@@ -352,6 +427,13 @@ func TestCrazyTown(t *testing.T) {
 					{4, 5, 6},
 				},
 			},
+			tc.ContainersOfContainers{
+				ListOfLists: [][]int32{
+					{1, 2, 3},
+					{4, 5, 6},
+					{7, 8, 9},
+				},
```
abhinav: The tests for Container and Typedefs are collating the round trip and Equals
tests. Can we pull them into their own `Test*` functions? The tests are hard
to read as it is; adding a second struct that is just "not equal" to the
struct being parsed is weird. You can leave the "x is equal to x" assert there
as a sanity check, though.

harshuber: That makes sense. Pushed another commit separating equals tests.

gen/equals.go
```diff
@@ -0,0 +1,105 @@
...
+	case *compile.SetSpec:
+		equals, err := e.setG.Equals(g, s)
+		return fmt.Sprintf("%s(%s, %s)", equals, lhs, rhs), err
+	default:
+		// Custom defined type
+		return fmt.Sprintf("%s.Equals(%s)", lhs, rhs), nil
+	}
+}
+
+// EqualsPtr is the same as Equals expect `lhs` and `rhs` are expected to be a
```
bombela: s/expect/except/

gen/equals.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	equalsPtrFunc := fmt.Sprintf("_%v_EqualsPtr", spec.ThriftName())
+	err := g.EnsureDeclared(
+		`
+			<$type := typeReference .Spec>
+			<$lhs := newVar "lhs">
+			<$rhs := newVar "rhs">
+			func <.Name>(<$lhs>, <$rhs> *<$type>) bool {
+				// Make sure that both the pointers are non nil.
```
bombela: s/the//

gen/map.go
```diff
@@ -190,6 +190,111 @@ func (m *mapGenerator) Reader(g Generator, spec *compile.MapSpec) (string, error
...
+				}
+
+				<$i := newVar "i">
+				<$j := newVar "j">
+				<$lk := newVar "lk">
+				<$lv := newVar "lv">
+				<$rk := newVar "rk">
+				<$rv := newVar "rv">
+				<$ok := newVar "ok">
+				for _, <$i> := range <$lhs> {
```
bombela: This code is O(n^2). It might be possible to go O(n log n) by sorting lhs and rhs before lopping over both at the same time.

gen/set.go
```diff
@@ -150,3 +150,64 @@ func (s *setGenerator) Reader(g Generator, spec *compile.SetSpec) (string, error
...
+				<$y := newVar "y">
+				<$ok := newVar "ok">
+				<if isHashable .Spec.ValueSpec>
+					for <$x> := range <$rhs> {
+						if _, <$ok> := <$lhs>[<$x>]; !<$ok> {
+							return false
+						}
+					}
+				<else>
+					// Note if values are not hashable then this is O(n^2) in time complexity.
```
bombela: As mentioned above, it might be possible to go O(n log n) by sorting lhs and rhs prior. A list equality check in O(n) will finish the job.

gen/map.go
abhinav: Sorting arbitrary objects might be out of scope for this PR. If we ever add a
`LessThan` function, we can optimize these operations.

gen/collision_test.go
```diff
@@ -122,6 +447,34 @@ func TestWithDefault(t *testing.T) {
...
+			CollisionField: true,
+		},
+	}
+
+	tests := []struct {
+		lhs, rhs tc.WithDefault
+		want     bool
+	}{
+		{a, a, true},
+		{b, b, true},
```
prashantv: just to complete the remaining combinations can you add `c, c` and `b, c` to the test tables

gen/equals.go
```diff
@@ -0,0 +1,105 @@
...
+	mapG  mapGenerator
+	setG  setGenerator
+	listG listGenerator
+}
+
+// Equals generates a string comparing rhs to the given lhs.
+// Equals generates an expression of type bool.
+func (e *equalsGenerator) Equals(g Generator, spec compile.TypeSpec, lhs, rhs string) (string, error) {
+	if isPrimitiveType(spec) {
+		if _, isEnum := spec.(*compile.EnumSpec); !isEnum {
```
prashantv: What about typedefs?

gen/testdata/collision/types.go
```diff
@@ -32,6 +32,12 @@ func (v *LittlePotatoe) FromWire(w wire.Value) error {
 	return err
 }
 
+func (lhs LittlePotatoe) Equals(rhs LittlePotatoe) bool {
+	l := (int64)(lhs)
+	r := (int64)(rhs)
+	return (l == r)
```
prashantv: I don't think we need this cast here, you can just do `return (l == r)`. Any reason why we do the conversion to an `int64`?

harshuber: It is to take care of structs as typedef.

gen/equals.go
harshuber: isPrimitiveType() currently returns false for typedefs. But the comment above isPrimitiveType() says typdefs are priitive types.

prashantv: Cool, let's fix those separately then

PR #289, opened 2017-02-10
gen: Fix helper naming conflicts

### LINKED ISSUE #287: codegen: Imported types with duplicate names cause conflicts
```
// common.thrift
typedef string UUID
```

```
// myservice.thrift

include "./common.thrift"

struct UUID {
  1: required i64 high
  2: required i64 low
}

struct Foo {
  1: required UUID localUUID
  2: required common.UUID importedUUID
}
```

This will fail with,

```
Failed to generate code: failed to generate code for ".../myservice.thrift": failed to generate code for "Foo": template: thriftrw:27:22: executing "thriftrw" at <fromWire .Type $valu...>: error calling fromWire: failed to generate code for "UUID": name "_UUID_Read" is already taken
```

And if we swap the other UUID with `typedef string UUID`, it generates code that doesn't compile:

```
myservice/types.go:63: cannot assign UUID to v.ImportedUUID (type common.UUID) in multiple assignment
```

Because it generates a single `UUID_Read` function:

```
func _UUID_Read(w wire.Value) (UUID, error) {
	var x UUID
	err := x.FromWire(w)
	return x, err
}
```

abhinav:
This fixes naming conflicts for helper functions and types like
`*_ValueList`, `*_MapItemList`, `*_Read`, `*_ptr`, and `*_Equals` for
the case where an imported type shares a name with a type defined in
the current Thrift file.

Failing test cases were added for all instances.

Depends on #288

Resolves #287


gen/mangle.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func newMangler() *mangler {
+	return &mangler{
+		names: make(map[string]map[string]string),
+		taken: make(map[string]struct{}),
+	}
+}
+
+func (m *mangler) MangleType(spec compile.TypeSpec) string {
```
prashantv: Should we add unit tests that test this method explicitly?

gen/mangle.go
```diff
@@ -0,0 +1,85 @@
...
+		return goCase(spec.ThriftName())
+	}
+
+	namesForFile, ok := m.names[thriftFile]
+	if !ok {
+		namesForFile = make(map[string]string)
+		m.names[thriftFile] = namesForFile
+	}
+
+	name, ok := namesForFile[spec.ThriftName()]
```
prashantv: nit: can inline into `if` below

abhinav:
Ready to merge if no objections

prashantv:
👍 looks good

PR #290, opened 2017-02-10
Handle the nil case in generated String() methods

dmcaulay:
I was seeing panics when I tried to log nil values with [zap](https://github.com/uber-go/zap). I'll add a `nil` check there to make sure other libraries don't introduce panics, but this seemed like a good safeguard to have in the generated structs.

Here's a [link](https://play.golang.org/p/zcXKrbDSwn) to some code that demonstrates the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ v1.1.0 (unreleased)
 -   AST: All type references and complex constant values now record the line
     numbers on which they were specified.
 -   AST: Added a Node interface to unify different AST object types.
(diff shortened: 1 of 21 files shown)


abhinav:
Please add this to the changelog

dmcaulay:
i don't have write access so feel free to land when ready 😄

prashantv:
Should we add some unit tests as well?

dmcaulay:
added a unit test

abhinav:
Thanks!

PR #294, opened 2017-03-20
WIP: Linter prototype

abhinav:
This implements a prototype Thrift linter based on the AST walking work
done in #285.

Linter rules walk the AST and post messages to the Lint Catcher.
SimpleRule exposes a simplified interface in front of that for basic
linters that don't need control over the AST traversal. These can simply
return true/false on a specific kind of Node and provide a
`text/template` compatible template for the message.

This is currently just a prototype and needs a lot more work. The
intention is to incubate it in thriftrw and move it to its own
thriftrw/thriftlint repo afterwards.


peter-edge:
@abhinav let me know when you want a review on this

cmd/thriftlint/lint.go
```diff
@@ -0,0 +1,193 @@
+package main
```
prashantv: what do you think of moving most of the logic into a library instead (APIs do not have to be guaranteed, but allows tools like `yab` to run the same lint rules)

abhinav: Moving much of the logic into a library is the intention. It'll make it easier to build Phab/whatever integration.

cmd/thriftlint/lint.go
```diff
@@ -0,0 +1,193 @@
...
+
+// Varying levels of issue severity.
+const (
+	Advice Severity = iota
+	Warning
+	Error
+)
+
+// Lint is an issue found in a Thrift file.
+type Lint struct {
```
peter-edge: idk what i think about this name, maybe `Issue`?

cmd/thriftlint/lint.go
```diff
@@ -0,0 +1,193 @@
...
+	// was found.
+	Line int
+
+	// Message is a plain-text message to the user indicating what the problem
+	// was.
+	Message string
+}
+
+// Catcher receives lint messages from different linter rules.
+type Catcher interface {
```
kriskowal: Trap :P

cmd/thriftlint/lint.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// Severity of messages posted by this rule.
+	Severity Severity
+
+	// A function in the form,
+	//
+	// 	func(w ast.Walker, n N) bool
+	//
+	// Where N is ast.Node or any type that implements ast.Node.
+	Inspect interface{}
```
kriskowal: such reflection. have you considered just writing each inspect function as an optional field?

abhinav: That will result in a very large number of Inspect functions :(

abhinav:
Abandoning PR.

PR #299, opened 2017-04-13
Embedding IDLs in the generated code

bombela:
All IDLs used for code generation are embedded as constant strings. This will
obviously bloat the binary on disk. But this wont take resident RAM.

A new structure must be shared by all generated packages for the
representation of a ThriftModule. A ThriftModule references included
ThriftModules. This means you get the full IDL tree from any entry point.

Motivation: YARPC services use ThriftRW for code generation when using Thrift
IDLs. It came up few times that we need access to the IDL sources used during
the build. Looking up files in a repo is inconvenient nor a guarantee of
correctness. By embedding IDL files in the generated code. And thus in the
binary itself, we guarantee correct introspection.


breerly:
@bombela thanks for the thorough description - that explains what's happening, and a little bit of why. Can you add the motivation for the change as well? That's pretty important to understanding where this goes.

bombela:
I naively assumed we all know why we want this.

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ Releases
 v1.2.0 (unreleased)
 -------------------
 
--   No changes yet.
+-   The Thrift IDL is now embedded withing the generated package.
```
abhinav: Mention the variable name under which it's available

gen/embedidl.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"crypto/sha1"
+	"encoding/hex"
+	"sort"
+
+	"go.uber.org/thriftrw/compile"
+)
+
+// EmbedIDL generate Go code with a full copy of the IDL embeded.
+func EmbedIDL(g Generator, i thriftPackageImporter, m *compile.Module) error {
```
abhinav: make this private because it depends on private types (`thriftPackageImporter`)

gen/embedidl.go
```diff
@@ -0,0 +1,86 @@
...
+	pkg, err := i.Package(m.ThriftPath)
+	if err != nil {
+		return wrapGenerateError("idl embeding", err)
```
abhinav: embedding

gen/embedidl.go
```diff
@@ -0,0 +1,86 @@
...
+		Package:  pkg,
+		FilePath: packageRelPath,
+		SHA1:     hex.EncodeToString(hash[:]),
+		Includes: includes,
+		Raw:      m.Raw,
+	}
+	err = g.DeclareFromTemplate(`
+		<$idl := import "go.uber.org/thriftrw/idl">
+
+		// ThriftModule represents the IDL file used to generate this package.
```
abhinav: FYI because of how the codegen works right now, this comment won't be included in the output.

gen/embedidl.go
```diff
@@ -0,0 +1,86 @@
...
+		FilePath string
+		SHA1     string
+		Includes []string
+		Raw      []byte
+	}{
+		Name:     m.Name,
+		Package:  pkg,
+		FilePath: packageRelPath,
+		SHA1:     hex.EncodeToString(hash[:]),
+		Includes: includes,
```
abhinav: Optional: This looks like the exact same shape as the ThriftModule type. Why not just use that as `data`?

gen/embedidl.go
```diff
@@ -0,0 +1,86 @@
...
+		Raw:      m.Raw,
+	}
+	err = g.DeclareFromTemplate(`
+		<$idl := import "go.uber.org/thriftrw/idl">
```
abhinav: The IDL package is really the entry point to the parser. I wouldn't use that
here.  I don't see any other obvious package to hold this type either so
consider creating a new one.

gen/embedidl.go
```diff
@@ -0,0 +1,86 @@
...
+
+		// ThriftModule represents the IDL file used to generate this package.
+		var ThriftModule = &<$idl>.ThriftModule {
+			Name: "<.Name>",
+			Package: "<.Package>",
+			FilePath: "<.FilePath>",
+			SHA1: "<.SHA1>",
+			Includes: []*<$idl>.ThriftModule {<range .Includes>
+					<.>.ThriftModule, <end>
```
abhinav: Consider,

```
<if .Includes>
  Includes: ...
<end>
```

That way `Includes` is left as `nil` if there are no includes.

gen/embedidl_test.go
```diff
@@ -0,0 +1,77 @@
...
+	rawIDL, sha1, err := loadIDL("enums.thrift")
+	if assert.NoError(t, err) {
+		assert.Equal(t, rawIDL, tm.Raw)
+		assert.Equal(t, sha1, tm.SHA1)
+	}
+}
+
+func TestIDLEmbeddingStructs(t *testing.T) {
+	tm := ts.ThriftModule
+	assert.Equal(t, "structs", tm.Name)
```
abhinav: table tests?

idl/thriftmodule.go
```diff
@@ -0,0 +1,32 @@
...
+package idl
+
+// ThriftModule is used by the generated code to expose the embedded IDL
+// source.
+type ThriftModule struct {
+	Name     string
+	Package  string
+	FilePath string
+	SHA1     string
+	Includes []*ThriftModule
```
abhinav: Document the fields please. Like, what's the SHA1 of. Is `Package` the Go package?

main.go
```diff
@@ -57,6 +57,7 @@ type genOptions struct {
 	NoTypes           bool `long:"no-types" description:"Do not generate code for types, implies --no-service-helpers."`
 	NoConstants       bool `long:"no-constants" description:"Do not generate code for const declarations."`
 	NoServiceHelpers  bool `long:"no-service-helpers" description:"Do not generate service helpers."`
+	NoIDLEmbedding    bool `long:"no-idl-embedding" description:"Do not embed IDLs into the generated code."`
```
abhinav: How about `--no-embed-idl`?

gen/embedidl.go
bombela: Yeah I know, but the day it goes trough it will be there :)

gen/embedidl.go
bombela: Includes is not identical because here its a simple slice of string. Its it possible to reuse ThriftModule and instantiate half-baked instances for the include package name. But not sure if best idea.

gen/embedidl.go
bombela: thriftrw/thriftreflect it is!

gen/embedidl.go
```diff
@@ -0,0 +1,88 @@
...
+		Raw:      m.Raw,
+	}
+	err = g.DeclareFromTemplate(`
+		<$idl := import "go.uber.org/thriftrw/thriftreflect">
+
+		// ThriftModule represents the IDL file used to generate this package.
+		var ThriftModule = &<$idl>.ThriftModule {
+			Name: "<.Name>",
+			Package: "<.Package>",
+			FilePath: "<.FilePath>",
```
abhinav: should we `printf "%q"` these too? Although this wouldn't usually happen, it is possible for the file path to have a quote in it.

bombela: damn, that's a very good point!

PR #302, opened 2017-04-26
Makefile: ensure make generate always works

Raynos:
This allows for running `make generate` after a flesh
clone of the thriftrw git repository without having
system dependencies.

However `make generate` does use `go get` to install
system dependencies for you...

r: @abhinav @kriskowal


kriskowal:
Tell me more about flesh clones. I’m considering various extreme sports.

Makefile
```diff
@@ -42,15 +42,29 @@ LINT_EXCLUDES := $(GENERATED_GO_FILES) $(LINT_EXCLUDES_EXTRAS)
...
 
 .PHONY: build
 build:
 	go build -i $(BUILD_FLAGS)
 
+.PHONY: install-ragel
+install-ragel:
+	ls ./vendor/ragel 1>/dev/null 2>/dev/null || ( \
+		( curl https://www.colm.net/files/ragel/ragel-6.9.tar.gz | \
+		tar -xz ) && mv ./ragel-6.9 ./vendor/ragel \
```
abhinav: I would let `make` decide if you need to build ragel instead of having your own check to see if it exists.

```
$(RAGEL_PATH)/bin/ragel:
    mkdir -p $(RAGEL_PATH)
    curl ...
```

You can compress down the curl, tar and mv to:

```
curl ... | tar -C ./vendor/ragel -xz --strip-components 1
```

The `-C` tells it to change to this directory before extracting and the strip-components tells it to drop the first component from the path of every file in the tar -- the "ragel-6.9" directory.

Makefile
```diff
@@ -42,15 +42,29 @@ LINT_EXCLUDES := $(GENERATED_GO_FILES) $(LINT_EXCLUDES_EXTRAS)
...
+	)
+	cd ./vendor/ragel ; (ls ./bin/ragel 2>/dev/null || \
+		(./configure --prefix=$(RAGEL_PATH) && make && make install) )
```
abhinav: `make && make install` can just be `make install`

Makefile
```diff
@@ -42,15 +42,29 @@ LINT_EXCLUDES := $(GENERATED_GO_FILES) $(LINT_EXCLUDES_EXTRAS)
...
+
 .PHONY: generate
-generate:
-	go build -i -tags=thriftrw.disableVersionCheck
-	PATH=$$(pwd):$$PATH go generate $$(glide nv)
+generate: install-ragel
+	go get -u github.com/golang/mock/mockgen
+	go get -u golang.org/x/tools/cmd/stringer
+	go get -u golang.org/x/tools/cmd/goyacc
```
abhinav: sure

Makefile
```diff
@@ -42,15 +42,29 @@ LINT_EXCLUDES := $(GENERATED_GO_FILES) $(LINT_EXCLUDES_EXTRAS)
...
+	go build \
+		-i -tags=thriftrw.disableVersionCheck
```
abhinav: why two lines for this one?

Makefile
```diff
@@ -42,15 +42,29 @@ LINT_EXCLUDES := $(GENERATED_GO_FILES) $(LINT_EXCLUDES_EXTRAS)
...
+	PATH=$$(pwd):$$PATH:$$PWD/vendor/ragel/bin go generate $$(glide nv)
```
abhinav: Replace the new one with `$(RAGEL_PATH)/bin`?

Makefile
Raynos: done.

Makefile
Raynos: done.

Makefile
Raynos: Oops. fixed.

Makefile
Raynos: fixed.

PR #304, opened 2017-04-26
Make `make lint` pass.

Raynos:
The linting command was broken in go1.7

This fixes it to pass.

r: @abhinav @kriskowal

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 WANT_VERSION = $(shell grep '^v[0-9]' CHANGELOG.md | head -n1 | cut -d' ' -f1)
 
 # Minor versions of Go for which the lint check should be run.
-LINTABLE_MINOR_VERSIONS := 7
+LINTABLE_MINOR_VERSIONS := 8
 
 # Paths besides auto-detected generated files that should be excluded from
 # lint results. If generated code uses '//line' directives with a different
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -62,7 +65,13 @@ ifdef SHOULD_LINT
 	@[ ! -s "$(FMT_LOG)" ] || (echo "gofmt failed:" | cat - $(FMT_LOG) && false)
 
 	$(eval VET_LOG := $(shell mktemp -t govet.XXXXX))
-	@go vet $(PACKAGES) 2>&1 | grep -v '^exit status' | $(FILTER_LINT) > $(VET_LOG) || true
+	@$(foreach pkg, $(PACKAGES), \
+		go tool vet $(VET_RULES) $(pkg) 2>&1 | \
+			grep -v '^exit status' | \
+			$(FILTER_LINT) | \
+			grep -v 'something"` not compatible with reflect.StructTag.Get:' \
```
abhinav: Is this last one from your other PR? can you defer adding this until that PR and we can evaluate whether there's a better way to do this.

Raynos: This is not from the other PR.

There is a multiline lint error msg from ./vendor/ `go tool vet` seems to go into vendor :(

Raynos:
lgtm.

PR #305, opened 2017-04-26
Add support for go.tag annotation

Raynos:
This PR adds initial support for go.tag annotation.
With the go.tag annotation you can now do :

```
struct Rename {
    1: required string Default (go.tag = "json:\"default\"")
    2: required string camelCase (go.tag = "json:\"snake_case\"")
}
```

A `go.tag` annotation must be a valid `reflect.StructTag`
value. We currently only pass the "json" struct tag and copy
it into the generated code.

If you add any other struct tags we do not copy them into the
generated go code.

In this PR i've also added support for generating a 
`required` option in the json struct tag which is supported
by [`easyjson`][1].

This first few commits in this PR are a duplicate of

r: @abhinav @kriskowal

  [1]: https://github.com/mailru/easyjson


gen/field.go
```diff
@@ -100,12 +103,36 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
-				return fmt.Sprintf("`json:\"%s,omitempty\"`", f.Name)
+				jsonTags += ",omitempty"
+			}
+
+			if f.Required {
+				jsonTags += ",required"
+			}
+
+			goTag := f.Annotations["go.tag"]
+			if goTag != "" {
```
bombela: Since `goTag` is only used withing the conditional's cope, you can do this:

`if goTag := string.Trim(f.Annotations["go.tag"]); goTag != ""`

I've also threw a string.Trim there. I am not sure if the parser stage already cleans that up. its easy enough to add a test for it though.

gen/field.go
```diff
@@ -100,12 +103,36 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+				structTag := reflect.StructTag(goTag)
```
bombela: instead of all this dance to find out the json name but dropping any other annotations. What about using a dedicated thrift annotation dedicated to override the json field name? `json.name` for example.

gen/field.go
```diff
@@ -100,12 +103,36 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+				if jsonTag != "" {
+					segments := strings.Split(jsonTag, ",")
+
+					// Assign jsonFieldName to be the name in json tag.
+					jsonFieldName = segments[0]
+				}
+			}
+
+			// If the field name is "-" then it means omit, add no tags
+			if jsonFieldName == "-" {
```
bombela: so if `go.tag="json: \"-\""`, you then omit the jsontags emitted by thriftrw. This doesnt come out as intuitive. You are re-purposing a go tag value inside a thrift annotation. And thriftrw is expected to always define the go json tag. Why would you want to ever omit it?

gen/field.go
Raynos: I think the idea is to add full support for `go.tag` ; I can add all the other information in this PR from `go.tag` if we want .

The json annotation needs merging with the default json annotation that thriftrw generates.

gen/field.go
Raynos: There are use cases to having a field in a thriftrw struct not serialized / deserialized into JSON.

For example

```
struct RequestBody {
    1: required string userUuid (go.tag = "json:\"-\"" , zanzibar.http.ref = "headers.x-uuid")
    2: required string name 
}
```

We have use cases for specifying that certain fields in the request body are serialized to headers instead. This means we generate our own client that serializes a field into a header and then calls json.Marshal() to serialize other fields into the body. For this to work correctly we need to mark a thrift field as not being json serialized.

gen/field.go
Raynos: good idea.

gen/field.go
```diff
@@ -100,12 +103,36 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
 			// optional collections because omitempty doesn't differentiate
 			// between nil and empty collections.
 
+			jsonTags := ""
```
Raynos: makes sense.

Raynos:
The `*_easyjson.go` file gets generated by easyjson.

I'll add a step to the Makefile to regenerate that file instead of generating it once.

gen/field.go
```diff
@@ -22,6 +22,9 @@ package gen
 
 import (
 	"fmt"
+	"strings"
+
```
abhinav: group the standard library imports together

gen/field.go
```diff
@@ -100,12 +103,36 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+			}
+
+			// If the field name is "-" then it means omit, add no tags
+			if jsonFieldName == "-" {
+				jsonTags = ""
 			}
 
-			return fmt.Sprintf("`json:%q`", f.Name)
 			// TODO(abg): Take go.tag and js.name annotations into account
+			return fmt.Sprintf("`json:\"%s%s\"`", jsonFieldName, jsonTags)
```
abhinav: This is incomplete. If I do,

```
1: required string something (go.tag = "yaml:\"lol\"")
```

This won't actually add that tag because you're not including `goTag`.

gen/field.go
```diff
@@ -100,12 +103,36 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
 			// between nil and empty collections.
 
+			jsonTags := ""
+			jsonFieldName := f.Name
+
 			if (isStructType(f.Type) || isPrimitiveType(f.Type)) && !f.Required {
-				return fmt.Sprintf("`json:\"%s,omitempty\"`", f.Name)
+				jsonTags += ",omitempty"
+			}
+
```
abhinav: This function is complicated enough at this point to pull it into a separate method.

    TemplateFunc("tag", f.generateTag)

gen/field.go
```diff
@@ -100,12 +103,36 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+			if f.Required {
+				jsonTags += ",required"
+			}
+
+			goTag := f.Annotations["go.tag"]
+			if goTag != "" {
+				structTag := reflect.StructTag(goTag)
+				jsonTag := structTag.Get("json")
+				if jsonTag != "" {
```
abhinav: For the JSON options, let's build a structure from the information in the JSON tag when we're doing all this work.

```
type jsonTag struct {
   FieldName string
  Options map[string]struct{}
}
```

gen/field.go
abhinav: FYI thrift supports single quotes for string literals too. So you can use

    go.tag = 'json:"foo"'

and avoid all this escaping

bombela: I see. Thank you for the explanation Raynos.

abhinav:
Superseded by #315

ChuntaoLu:
@abhinav Can we get this branch back? https://github.com/thriftrw/thriftrw-go/pull/315 does not support "required" json tag.

cc @Raynos

abhinav:
Hi @ChuntaoLu @Raynos,

I have restored the branch to fix your build but I don't intend on keeping it
around for long. Why are you relying on a development branch of this library?
We make no guarantees for development branches. They may be created, deleted,
and rewritten at the will of the maintainers. This branch **will** be deleted
again in the near future.

Please switch to the most recent stable release. If there's a feature you want
on that branch, work with us to get it merged into master.

As for that feature: @alextbok, the author of #315, tried to reach out to
@Raynos to get more details about it but was unable to, so the feature was
dropped. Now that #315 has been reviewed and merged, if you still need the
feature, I recommend implementing it on top of master, submitting a new PR,
and getting it merged into master.

Raynos:
I planned to get this branch updated but forgot, ill open a new PR for required support.

ChuntaoLu:
Thank you @abhinav

PR #307, opened 2017-05-03
Use go.uber.org/multierr

abhinav:
This drops our custom MultiError implementation in favor of
go.uber.org/multierr.


glide.yaml
```diff
@@ -12,3 +12,5 @@ import:
   - go/ast/astutil
 - package: github.com/jessevdk/go-flags
 - package: github.com/anmitsu/go-shlex
+- package: go.uber.org/multierr
+  version: ~0.2.0
```
breerly: Yikes - taking a hard dep on a pre 1.0 lib, in a 1.0 lib...

Doesn't this mean we need to 1.0 multierr now?

@abhinav @prashantv lets jam in person - I don't think I'm mistaken.

abhinav: Not necessarily. This is never exposed directly to users. As long as we're pinning to it, we're fine. Worst case, we actually check in the vendored code.

breerly: kk, sgtm

PR #308, opened 2017-05-04
gen/enum: Generate <Name>_Values function

### LINKED ISSUE #134: EnumSpec: Support reverse mapping
EnumSpec should provide a reverse mapping from enum item value to enum item. Behavior will be undefined for multiple enums with the same value. Alternatively, it can map to a list of items with matching values.

recht:
This generates a `<Name>_Values` function for all enums which returns
all known values for that enum.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ v1.3.0 (unreleased)
 -   Enums now implement the `encoding.TextUnmarshaler` interface. This allows
     retrieving enum values by name and integrates better with other encoding
     formats.
+-   Enums now have a `Get<EnumName>Values()` function which returns all known
+    values for that enum.
 
 
 v1.2.0 (2017-04-17)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ v1.3.0 (unreleased)
 -   Enums now implement the `encoding.TextUnmarshaler` interface. This allows
     retrieving enum values by name and integrates better with other encoding
     formats.
--   Enums now have a `Get<EnumName>Values()` function which returns all known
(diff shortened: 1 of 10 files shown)


abhinav:
Hey, thanks for implementing this. Two things:

-  I'm not sure about introducing a new top-level function. We've generally maintained a policy of avoiding reserving names in the top-level namespace unless absolutely necessary. Are you getting much value out of this function?
- Can you rename FromText to UnmarshalText and change it to accept a `[]byte` so that it conforms to [`encoding.TextUnmarshaler`](https://golang.org/pkg/encoding/#TextUnmarshaler)? That'll give us support for YAML and some other encodings for free.

recht:
As for the first question: currently there's no way of listing valid values, which makes it quite hard to do anything data driven. I'm all for not having too many top level symbols, but not being able to list values is pretty annoying, and adding it as a func on the type also seems pretty weird to me...

As for renaming: it makes sense, I'll fix that.

recht:
Any update on this?

abhinav:
Hey, @recht. Since the UnmarshalText change is independent of the GetValues
change, I pulled it into its own PR. Can you give #309 a look and verify
everything is in order?

For the `GetValues` change, I buy the need for such a function. We have a
workaround for the namespace issue: ThriftRW ensures that the code generated
for user-defined types does not use underscores even if the user-defined type
had underscores in its name in the Thrift file. This effectively reserves the
namespace of things with underscores for ThriftRW. So we can name these
functions `<EnumName>_Values()` and not worry about naming conflicts at all.
I've made the required changes for that to this PR. Please review this change
and we will merge it.

abhinav:
LGTM. @bombela @prashantv can one of you have a look when you get a chance? Thanks.

Issue #293, opened 2017-03-16
Add Go.Tag annotation support


mjcowan:
Please add go.tag annotation support for fields.

It would be great if all tags could be converted to go tags i.e

```
struct Foo {
   1: optional string baz (tag1 = 'foobar')
}
```

Generates

```
type Foo struct {
   baz *string `tag1="foobar"`
}
```

Raynos:
Partially implemented by #305

alextbok:
Implemented via https://github.com/thriftrw/thriftrw-go/pull/315 with the caveat that only `go.tag` annotations are converted to go tags

PR #315, opened 2017-07-07
Support go.tag in thrift struct field annotations

alextbok:
Resolves https://github.com/thriftrw/thriftrw-go/issues/293


gen/field.go
```diff
@@ -95,22 +104,79 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+func annotationsToRawTag(t compile.Annotations) string {
+	var a []string
+	for k, v := range t {
+		a = append(a, compileTag(k, v))
+	}
+	return strings.Join(a, goTagDelimiter)
+}
+
+// compileTag constructs a single go tag from the tag name and value
+func compileTag(k, v string) string {
```
abhinav: Naming: This isn't really compilation. You're building a string representation of the tag from the key-value pair.

You don't really need this function. You can inline the logic in annotationsToRawTag. The other usage of this becomes unnecessary once you switch to `map[string]string` up there.

gen/field.go
```diff
@@ -95,22 +104,79 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+	return fmt.Sprintf("`%s`", strings.Join(tags, goTagDelimiter))
+}
+
+// annotationsFromRawTag constructs an annotations map from a raw tag string
+// `json:"foo,required" bson:"bar"` -> {"json": "foo,required", "bson": "bar"}
+func annotationsFromRawTag(a string) compile.Annotations {
+	ts := map[string]string{}
+	for _, t := range strings.Split(a, goTagDelimiter) {
+		k := strings.Split(t, goTagSep)[0]
+		ts[k] = reflect.StructTag(t).Get(k)
```
abhinav: This isn't really a reversible operation. `compile.Annotations` represents the annotations from the Thrift file. Not all Thrift annotations map to Go tags (in fact, most of them don't), and not all Go tags map back to annotations. To me it sounds like you're trying to convert a struct tag to a `map[string]string` Just do that instead.

gen/field.go
```diff
@@ -22,10 +22,19 @@ package gen
...
+	"strings"
 
 	"go.uber.org/thriftrw/compile"
 )
 
+const (
+	goTagName      = "go.tag"
+	goTagDelimiter = " "
+	goTagSep       = ":"
+	jsonTagName    = "json"
```
abhinav: I would drop the `goTagDelimiter` and `goTagSep` constants. They hurt readability.

gen/field.go
```diff
@@ -95,22 +104,79 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
-			// optional collections because omitempty doesn't differentiate
-			// between nil and empty collections.
-
-			if (isStructType(f.Type) || isPrimitiveType(f.Type)) && !f.Required {
-				return fmt.Sprintf("`json:\"%s,omitempty\"`", f.Name)
-			}
-
-			return fmt.Sprintf("`json:%q`", f.Name)
-			// TODO(abg): Take go.tag and js.name annotations into account
+			return generateTags(f)
```
abhinav: generateTags has the same signature as the function calling it. You can do,

```
TemplateFunc("tag", generateTags)
```

gen/field.go
```diff
@@ -95,22 +104,79 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
 		TemplateFunc("declFieldName", f.declFieldName),
 	)
 }
 
+// generateTags parses the annotation on the thrift field and creates the resulting go tag
+func generateTags(f *compile.FieldSpec) string {
+	// start with default json tag
+	tags := []string{
+		compileTag(jsonTagName, compileJSONTagValue(f, f.Name)),
+	}
```
abhinav: When compiling the list of tags that will be added, use a `map[string]string` instead of this. It'll ensure no duplicates.

gen/field.go
```diff
@@ -95,22 +104,79 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+
+	// process go tags and overwrite json tag if specified in thrift annotation
+	if goAnnotation := f.ThriftAnnotations()[goTagName]; goAnnotation != "" {
```
abhinav: nit: `f.ThriftAnnotations()` => `f.Annotations`

gen/field.go
abhinav: Plus perhaps document the other constants.

gen/field.go
```diff
@@ -95,22 +104,79 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+		if jt := reflect.StructTag(goAnnotation).Get(jsonTagName); jt != "" {
+			// truncate current tags to remove current json tag and compile the one defined in thrift
```
abhinav: The truncation would be unnecessary if you used a map.

gen/field.go
```diff
@@ -95,22 +104,79 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+			goTags := annotationsFromRawTag(goAnnotation)
+			goTags[jsonTagName] = compileJSONTagValue(f, jt)
+			tags, f.Annotations[goTagName] = tags[1:], annotationsToRawTag(goTags)
```
abhinav: Don't mutate the FieldSpec. Codegen logic must not change the compiled representation of the Thrift file.

gen/field.go
```diff
@@ -95,22 +104,79 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+	// TODO (abg): Take js.name annotation into account
+	return fmt.Sprintf("`%s`", strings.Join(tags, goTagDelimiter))
+}
+
+// annotationsFromRawTag constructs an annotations map from a raw tag string
+// `json:"foo,required" bson:"bar"` -> {"json": "foo,required", "bson": "bar"}
+func annotationsFromRawTag(a string) compile.Annotations {
+	ts := map[string]string{}
+	for _, t := range strings.Split(a, goTagDelimiter) {
+		k := strings.Split(t, goTagSep)[0]
```
abhinav: This will break for`foo:"bar baz"`. It's valid for tag values to have spaces in them.

You can easily fix this by assuming that if a section doesn't have `:`, it's part of a value with a space in it.

gen/struct_test.go
```diff
@@ -979,10 +994,45 @@ func TestStructJSON(t *testing.T) {
...
+	gt := &ts.GoTags{
+		Foo:    "shouldomit",
+		FooBar: "shouldnotomit",
+	}
+	b, err := json.Marshal(gt)
+	assert.NoError(t, err, "should marshal")
+	assert.Equal(t, b, []byte(`{"foobar":"shouldnotomit"}`))
+
+	foo, _ := reflect.TypeOf(gt).Elem().FieldByName("Foo")
+	assert.Equal(t, "json:\"-\" foo:\"bar\"", string(foo.Tag))
```
abhinav: You can use the ` string literal form to avoid escaping

    `json:"-" foo:"bar"`

gen/field.go
```diff
@@ -95,22 +104,79 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+	var a []string
+	for k, v := range t {
+		a = append(a, compileTag(k, v))
+	}
+	return strings.Join(a, goTagDelimiter)
+}
+
+// compileTag constructs a single go tag from the tag name and value
+func compileTag(k, v string) string {
+	return fmt.Sprintf("%s%s\"%s\"", k, goTagSep, v)
```
abhinav: Also, you can use the `%q` format specifier to do the escaping for you.

    fmt.Sprintf("%s:%q", k, v)

gen/field.go
alextbok: ah duh, thanks for catching :)

gen/field.go
alextbok: neat, didn't know this

gen/field.go
```diff
@@ -94,23 +105,88 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
-		}),
+		TemplateFunc("tag", generateTags),
 		TemplateFunc("declFieldName", f.declFieldName),
 	)
 }
 
+// generateTags parses the annotation on the thrift field and creates the resulting go tag
+func generateTags(f *compile.FieldSpec) string {
+	// start with default json tag
+	tk := []string{jsonTagName} // list of tag keys to preserve order
```
abhinav: You don't need to go through all this trouble to preserve ordering. Tag ordering doesn't matter since it's basically a key-value pair. We should just make sure it's deterministic by sorting the map keys before generating the Go tag string.

gen/field.go
```diff
@@ -94,23 +105,88 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+// it also returns a list of keys to preserve tag order
+func mapFromRawTag(a string) (map[string]string, []string) {
+	ts, tk := map[string]string{}, []string{}
+	split := strings.Split(a, " ")
+	for i, t := range split {
+		if t[len(t)-1] != '"' {
+			split[i+1] = strings.Join(split[i:i+2], " ")
+			continue
+		}
+		k := strings.Split(t, ":")[0]
```
abhinav: This is perhaps not the safest way to implement this. I'm guessing there are corner cases you or I haven't thought of. Escaped quotes (`\"` come to mind).

Unless we're willing to replicate the logic from reflect.StructTag, we should maybe use https://github.com/fatih/structtag for this. That library seems more battle tested because it's used for vim-go's struct tag integration.

Note from the readme,

> Please vendor the library as it might change in future versions.

So you'll want to put it in vendor/ and check it in. To do this cleanly, `glide get` the library as usual, remove vendor/ from .gitignore, and then in vendor/ add a .gitignore with the following contents:

```
# We will explicitly whitelist vendored libraries that we want to check in.
/*

!/github.com/fatih/structtag
```

gen/field.go
```diff
@@ -94,23 +105,88 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+	var a []string
+	for _, k := range tk {
+		a = append(a, fmt.Sprintf("%s:%q", k, t[k]))
+	}
+	return fmt.Sprintf("`%s`", strings.Join(a, " "))
+}
+
+// compileJSONTagValue constructs the json tag value
+// it inspects the thrift field spec and sets the tag value accordingly
+func compileJSONTagValue(f *compile.FieldSpec, segments ...string) string {
```
abhinav: nit: you're not really using the varargs here. Just accept `name` as a single argument. Plus you lose some safety because the code below assumes `len(segments) > 0`.

gen/field.go
alextbok: yea I thought it'd be nice to preserve the author's order. I also think it helps readibility to see the json tag first since it will always be there. Other tags will be on some fields and not others so the tags wouldn't look nicely left-aligned since they could all have different starting tags

gen/field.go
alextbok: ah yea I'm not using them anymore. I'll remove

gen/field.go
```diff
@@ -94,23 +107,66 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
-			// TODO(abg): Take go.tag and js.name annotations into account
-		}),
+		TemplateFunc("tag", generateTags),
 		TemplateFunc("declFieldName", f.declFieldName),
 	)
 }
 
+// generateTags parses the annotation on the thrift field and creates the resulting go tag
+func generateTags(f *compile.FieldSpec) (string, error) {
+	ts := []string{}
```
abhinav: nit: Use `var` form for empty slices.

https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

gen/field.go
```diff
@@ -94,23 +107,66 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+
+	jsonTag := compileJSONTag(f, f.Name)
+
+	// process go tags and overwrite json tag if specified in thrift annotation
+	if goAnnotation := f.Annotations[goTagKey]; goAnnotation != "" {
+		tags, err := structtag.Parse(goAnnotation)
+		if err != nil {
+			return "", errors.New(fmt.Sprintf("failed to parse tag %q: %v", goAnnotation, err))
```
abhinav: nit: Use fmt.Errorf

    return "", fmt.Errorf("failed to parse tag %q: %v", goAnnotation, err)

gen/field.go
```diff
@@ -94,23 +107,66 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+	}
+
+	ts = append(ts, jsonTag)
+
+	sort.Slice(ts, func(i, j int) bool {
+		return ts[i] < ts[j]
+	})
+
+	// TODO (abg): Take js.name annotation into account
+	return fmt.Sprintf("`%s`", strings.Join(ts, " ")), nil
```
abhinav: Optional: Some of this logic can further be simplified by using `*structtag.Tags` directly.

```go
tags, err := structtag.Parse("")  // no tags
if err != nil { .. }

if err := tags.Set(compileJSONTag(..)); err != nil { .. }
```

You won't need to have special handling for overwriting the json tag because
`tags.Set` will take care of stomping on it if it already exists.

And on this line, you can simply do,

```
fmt.Sprintf("`%s`", tags.String())
```

gen/field.go
```diff
@@ -94,23 +107,66 @@ func (f fieldGroupGenerator) DefineStruct(g Generator) error {
...
+		Options: opts,
+	}
+
+	if name == "-" {
+		// If the field name is "-" then it means omit, add no tags
+		return t.String()
+	}
+
+	if (isStructType(f.Type) || isPrimitiveType(f.Type)) && !f.Required {
+		t.Options = append(t.Options, "omitempty")
```
abhinav: We should probably ensure it doesn't already have omitempty with [Tag.HasOption](http://godoc.org/github.com/fatih/structtag#Tag.HasOption).

Plus perhaps a test where the `go.tag` annotation does include `omitempty`.

gen/field.go
alextbok: good suggestion; updated

PR #316, opened 2017-07-24
main.go: exit non-zero if flag parsing fails

drafnel:
Currently, if an error is generated while parsing flags, an error
message is emitted by the go-flags package, but we then return a nil
error from do() causing the main function to exit with a zero status.

Let's take over the responsibility of emitting error messages from
go-flags and then propogate any errors that are produced during parsing
back up to the main function.  This means we need to handle ErrHelp
ourselves.

### DIFF
--- a/main.go
+++ b/main.go
@@ -77,12 +77,15 @@ func do() (err error) {
 
 	var opts options
 
-	parser := flags.NewParser(&opts, flags.Default)
+	parser := flags.NewParser(&opts, flags.Default & ^flags.PrintErrors)
 	parser.Usage = "[OPTIONS] FILE"
 
 	args, err := parser.Parse()
-	if err != nil {
-		return nil // message already printed by go-flags
+	if ferr, ok := err.(*flags.Error); ok && ferr.Type == flags.ErrHelp {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -77,12 +77,15 @@ func do() (err error) {
 
 	var opts options
 
-	parser := flags.NewParser(&opts, flags.Default)
+	parser := flags.NewParser(&opts, flags.Default & ^flags.PrintErrors)
```
abhinav: Hah, I would just change it to

    flags.HelpFlag | flags.PassDoubleDash

drafnel: I don't feel strongly about it, but I tend to do things this way when I want "whatever the default happens to be, except some specific set", that way if the default changes, this code doesn't need to change.  I also expect it to be easier to review since the reviewer doesn't need to lookup what the default set is and then verify that the newly specified set is equivalent to the default - <undesired-flags>.  But if you feel strongly about it, I can change it.  Let me know.

abhinav:
Thanks!

breerly:
@drafnel thanks for the contrib!

PR #317, opened 2017-07-25
Generated code should conform to standard

abhinav:
This changes code generated by ThriftRW Go to conform to
https://golang.org/s/generatedcode.

The golden_test verifies that all generated files match the standard.


breerly:
Oh wow, i didn't know about the generated code spec - super super awesome.

PR #318, opened 2017-08-01
ast: Record docstrings

### LINKED ISSUE #319: AST: Support parsing docstrings
We should parse docstrings and make them available in the AST.

abhinav:
This adds support for reading docstrings from Thrift IDLs.

We currently parse docstrings specified for,

- [x] Parse docstrings for all entities.
  - [x] Top-level definitions
  - [x] Enum items
  - [x] Struct fields
  - [x] Service functions
  - [x] Function parameters
  - [x] Function exceptions
- [x] Clean up docstrings. We need to remove the leading spaces and `*` from
      each line in the docstring based on the indentation, etc. That is,
      
          /**
           * foo
           * bar
           */
           
    Should be interpreted as,
      
          foo
          bar
      
    Not,
      
          * foo
          * bar
- [x] Tests

Follow up tasks:

- [ ] Include docstrings in generated code

Resolves #319

### DIFF
--- a/idl/internal/docstring.go
+++ b/idl/internal/docstring.go
@@ -44,18 +44,11 @@ func ParseDocstring(s string) string {
 	// At this point, we need to strip the leading "*" and " *" from every
 	// line and dedent again.
 	for i, l := range lines {
-		if len(l) == 0 {
-			continue
-		}
-
-		if l[0] == '*' {
+		switch {
+		case len(l) > 0 && l[0] == '*':
 			lines[i] = l[1:]
-			continue
-		}
-
-		if len(l) > 1 && l[:2] == " *" {
+		case len(l) > 1 && l[:2] == " *":
 			lines[i] = l[2:]
-			continue
 		}
 	}
 
--- a/idl/internal/docstring.go
+++ b/idl/internal/docstring.go
@@ -84,9 +84,23 @@ func dropTrailingEmptyLines(lines []string) []string {
 	return lines
 }
 
-// skipFirstIfUnindented tells dedent to not consider the first line for the
-// prefix if it's not already indented. This is needed because our docstringcs
-// will start at "/**" without any leading whitespace.
+// Takes a series of lines that have been indented and removes the indentation
+// of the first line from all lines.
+//
+// So,
+//
+//  "    foo"
+//  "      bar"
+//
+// Becomes,
+//
+//  "foo"
+//  "  bar"
+//
+// If the first line was not indentend, skipFirstIfUnindented controls whether
+// we strip the indentation of the second line from all lines instead. This
+// flag is needed because our docstringcs will start at "/**" without any
+// leading whitespace.
 func dedent(lines []string, skipFirstIfUnindented bool) {
 	if len(lines) == 0 {
 		return
--- a/idl/internal/docstring.go
+++ b/idl/internal/docstring.go
@@ -23,20 +23,17 @@ func CleanupDocstring(s string) string {
 	lines := strings.Split(s, "\n")
 	dedent(lines, true /* skipFirstIfUnindented */)
 
-	switch len(lines) {
-	case 0:
+	if len(lines) == 0 {
 		return s
-	case 1:
-		// /** foo */
-		s := lines[0]
-		s = strings.TrimPrefix(s, _docstringOpen)
-		s = strings.TrimSuffix(s, _docstringClose)
-		return strings.TrimSpace(s)
 	}
 
-	if len(lines) > 0 {
-		lines[0] = strings.TrimPrefix(lines[0], _docstringOpen)
-		lines[len(lines)-1] = strings.TrimSuffix(lines[len(lines)-1], _docstringClose)
+	// Strip comment markers from start and end.
+	lines[0] = strings.TrimPrefix(lines[0], _docstringOpen)
+	lines[len(lines)-1] = strings.TrimSuffix(lines[len(lines)-1], _docstringClose)
+
+	if len(lines) == 1 {
+		// Single-line doc block like, /** foo */
+		return strings.TrimSpace(lines[0])
 	}
 
 	lines = dropLeadingEmptyLines(lines)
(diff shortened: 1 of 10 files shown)


abhinav:
This diff is ready to review.

idl/internal/docstring.go
```diff
@@ -0,0 +1,138 @@
...
+// CleanupDocstring takes a docstring in the form,
+//
+//  /**
+//   * foo bar
+//   */
+//
+// And returns,
+//
+//  foo bar
+func CleanupDocstring(s string) string {
```
breerly: Cleanup feels weird here, suggest `ParseDocstring` instead.

prashantv: Or maybe `FormatDocstring`?

idl/internal/docstring.go
```diff
@@ -0,0 +1,138 @@
...
+	lines := strings.Split(s, "\n")
+	dedent(lines, true /* skipFirstIfUnindented */)
+
+	switch len(lines) {
+	case 0:
+		return s
```
prashantv: don't think this case can happen,
https://play.golang.org/p/fswgcibGYQ

idl/internal/docstring.go
```diff
@@ -0,0 +1,138 @@
...
+	case 1:
+		// /** foo */
+		s := lines[0]
+		s = strings.TrimPrefix(s, _docstringOpen)
+		s = strings.TrimSuffix(s, _docstringClose)
+		return strings.TrimSpace(s)
+	}
+
+	if len(lines) > 0 {
```
prashantv: We must have `len(lines) > 0` if we got to this point

idl/internal/docstring.go
```diff
@@ -0,0 +1,138 @@
...
+		lines[0] = strings.TrimPrefix(lines[0], _docstringOpen)
+		lines[len(lines)-1] = strings.TrimSuffix(lines[len(lines)-1], _docstringClose)
+	}
```
prashantv: this is basically the same logic that the `case 1` also has, trim the `_docstringOpen` and `_docstringClose` from the first and last line. Maybe the logic should look like:
```go
lines := strings.Split(...)
last := len(lines)-1

// Trim the /** and */ from the first and last lines.
lines[0] = strings.TrimPrefix(lines[0], _docstringOpen)
lines[last] = strings.TrimSuffix(lines[last], _docstringClose)

if len(lines) == 1 {
  return strings.TrimSpace(lines[0])
}

lines = dropLeadingEmptyLines(lines)
[...]
```

idl/internal/docstring_test.go
```diff
@@ -0,0 +1,118 @@
...
+		},
+		{
+			give: []string{
+				"/**",
+				"	    * foo does stuff",
+				"	    */",
+			},
+			want: []string{"foo does stuff"},
+		},
+	}
```
prashantv: can you add a test case where the lines inside are missing the ` * `? e.g.,
```
give : []string{
  "/**",
  "no prefix",
  "for lines",
  "*/",
},
```

and maybe another case where we have a mix of lines with the prefix and without

idl/internal/docstring.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+		lines = lines[:i]
+	}
+	return lines
+}
+
+// skipFirstIfUnindented tells dedent to not consider the first line for the
+// prefix if it's not already indented. This is needed because our docstringcs
+// will start at "/**" without any leading whitespace.
+func dedent(lines []string, skipFirstIfUnindented bool) {
```
prashantv: don't really understand what this function is supposed to do, can you put a short description in the comment above

idl/internal/docstring.go
abhinav: I like Parse more than Format here. This isn't formatting a docstring to be printed but parsing what the actual text from a docstring.

idl/internal/docstring.go
abhinav: That's probably true but I would like to keep the check in for the sake of defensive programming.

idl/internal/docstring.go
abhinav: Good catch with this and the other comment. A bunch of code got moved around but I forgot to remove the checks that were now redundant.

idl/internal/docstring.go
abhinav: Added. My original thought was that dedent would be self-explanatory as the reverse of indent except this function became really complicated.

abhinav:
Updated. The goal is to make a best-effort attempt of extracting the paragraph inside the docstring.

In,

```
struct X {
  /**
   * foo
   * bar
   */
  1: string foo
}
```

The string we get for the docstring of the field is,

```
/**
   * foo
   * bar
   */
```

We need to extract everything on the right side of the `*`s, preserve
indentation, and ideally, handle cases where the star is missing, indentation
is off, etc.

billf:
lgtm

idl/internal/docstring.go
```diff
@@ -0,0 +1,144 @@
...
+
+		// Dedent only if the first non-space character appears at or after
+		// the prefix.
+		if nonSpace >= len(prefix) {
+			lines[i] = s[len(prefix):]
+		}
+	}
+}
+
+func notIsSpace(r rune) bool {
```
kriskowal: pedantic: isNotSpace

idl/internal/docstring.go
```diff
@@ -0,0 +1,144 @@
...
+// Becomes,
+//
+//  "foo"
+//  "  bar"
+//
+// If the first line was not indentend, skipFirstIfUnindented controls whether
+// we strip the indentation of the second line from all lines instead. This
+// flag is needed because our docstringcs will start at "/**" without any
+// leading whitespace.
+func dedent(lines []string, skipFirstIfUnindented bool) {
```
kriskowal: Pythonic precedent notwithstanding, pedantically, the right word is unindent. https://english.stackexchange.com/questions/149511/what-is-the-opposite-word-to-indent-outdent-or-unindent

idl/internal/docstring.go
abhinav: I was going for `not(unicode.IsSpace)` but I see your point.

PR #322, opened 2017-08-14
Add support for required easyjson annotation

Raynos:
It's useful for the go structs generated by thriftrw to support
the `required` go tag annotation that easyjson uses to generate
"validating" json parser.

The third party library [easyjson](https://github.com/mailru/easyjson)
supports the notion of [`required tag`](https://github.com/mailru/easyjson/blob/master/tests/data.go#L667-L670)

This allows a user of thriftrw that uses JSON encoding ( instead of thrift
binary encoding ) to be able to validate that all required fields exist
when doing the parsing stage.

r: @abg


gen/field.go
```diff
@@ -161,6 +161,10 @@ func compileJSONTag(f *compile.FieldSpec, name string, opts ...string) *structta
 		t.Options = append(t.Options, "omitempty")
 	}
 
+	if f.Required {
+		t.Options = append(t.Options, "required")
```
abhinav: Will this continue to work if I was doing, `(go.tag = "json:\"required\"")`? Can you add a test for this?

(You may want to check `HasOption` like the previous block.)

Raynos: yeah will test / verify.

Raynos:
```
 type Rename struct {
-       Default   string `json:"default,required"`
+       Default   string `json:"default,required,required"`
        CamelCase string `json:"snake_case,required"`
 }
```

lol.

Raynos:
Updated with more tests.

abhinav:
FYI @alextbok in case you wanted a look.

PR #324, opened 2017-08-25
gen: Fix bug for typedefed defaults

abhinav:
We have a bug where if a field with a default value is a typedef of a
primitive type, the generated code will be equivalent to,

    type Timestamp int64

    var value *Timestamp = ptr.Int64(42)

This will fail because you cannot assign a `*int64` to a `*Timestamp`.

This commit fixes this issue by generating a `_Timestamp_ptr` function
similar to one generated for other types.


gen/testdata/thrift/typedefs.thrift
```diff
@@ -18,6 +18,10 @@ struct Event {
     2: optional Timestamp time  // optional typedef
 }
 
+struct DefaultPrimitiveTypedef {
+    1: optional State state = "hello"
```
abhinav: This would previously generated code that wouldn't compile

akshayjshah:
lgtm

PR #325, opened 2017-08-26
gen: Generate getters for optional primitives

### LINKED ISSUE #295: Generate methods to get the value or the zero value for a field
In some of our code we don't need to distinguish between optional fields that are left out and the zero value for the type. It'd be nice if there were getters or something so we didn't need to define functions like this:

```go
func stringEmptyOrVal(x *string) string {
	if x == nil {
		return ""
	}
	return *x
}

func float64EmptyOrVal(x *float64) float64 {
	if x == nil {
		return 0
	}
	return *x
}

func int64EmptyOrVal(x *int64) int64 {
	if x == nil {
		return 0
	}
	return *x
}
```

abhinav:
The majority use case for accessing values of optional primitives is to
dereference it as-is or use the default/zero value if not.

This modifies code generation to include accessor functions for optional
primitive fields which return non-pointer versions of those fields.

So instead of,

    var foo string
    if s.Foo != nil {
        foo = *s.Foo
    }
    doSomething(foo)

Users can do,

    doSomething(s.GetFoo())

This should also ease migration for users coming from Apache Thrift.

Resolves #295

Depends on #324


gen/field.go
```diff
@@ -440,3 +444,24 @@ func (f fieldGroupGenerator) Equals(g Generator) error {
...
+func (f fieldGroupGenerator) PrimitiveAccessors(g Generator) error {
+	return g.DeclareFromTemplate(
+		`
+		<$v := newVar "v">
+		<$o := newVar "o">
+		<$name := .Name>
+		<range .Fields>
+			<$fname := goName .>
+			<if and (not .Required) (isPrimitiveType .Type)>
+			func (<$v> *<$name>) Get<$fname>() (<$o> <typeReference .Type>) {
```
akshayjshah: nice - named return makes the generated code simpler.

gen/field.go
```diff
@@ -440,3 +444,24 @@ func (f fieldGroupGenerator) Equals(g Generator) error {
...
+
+func (f fieldGroupGenerator) PrimitiveAccessors(g Generator) error {
+	return g.DeclareFromTemplate(
+		`
+		<$v := newVar "v">
+		<$o := newVar "o">
+		<$name := .Name>
+		<range .Fields>
+			<$fname := goName .>
+			<if and (not .Required) (isPrimitiveType .Type)>
```
akshayjshah: Does generating the same type of accessors for non-primitive types introduce much extra complexity?

abhinav: No, but you don't gain anything by doing it. Non-primitive types are always passed around as pointers so the accessor will simply return the value of the field. `x.Foo` will be no different from `x.GetFoo()`, so users have no reason to use `GetFoo`.

abhinav:
TODO: Test for the case,

```
struct Foo {
  1: optional string bar
  2: optional string get_bar
}
```

(We'll need to use `go.name` to override the Go field name for `get_bar`.)

gen/field.go
akshayjshah: Ah, makes sense.

gen/testdata/thrift/collision.thrift
```diff
@@ -50,3 +50,13 @@ union union_collision {
...
 }
+
+struct AccessorNoConflict {
+    1: optional string getname
+    2: optional string get_name
+}
+
+struct AccessorConflict {
+    1: optional string name
+    2: optional string get_name (go.name = "GetName2")
```
kriskowal: This is potentially a breaking change, right?

abhinav: Yeah, this is potentially breaking but for unlikely conflicts, we've taken the policy of suggesting "use `go.name` in case of conflict" rather than cutting a 2.0.

From a quick look at all IDLs in our system, I don't see anything this would actually cause conflicts for. When people use "getFoo" in the request struct, "foo" is usually an optional field in the response struct.

PR #331, opened 2017-09-26
gen/json: Enable omitempty for optional collections

### LINKED ISSUE #330: Add `omitempty` json tag for optional list and map fields
As per `encoding/json` documentation: https://golang.org/pkg/encoding/json/#Marshal
> The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string.

If an thrift struct definition has an optional list or map:

```
struct OptionalListField {
    1: optional list<Item> items
    2: optional map<Key, Value> dictionary
}
```

then thriftrw should add the `omitempty` json tag:

```
type OptionalListField struct {
	Items       []Item         `json:"items,omitempty"`
	Dictionary  map[Key]Value  `json:"dictionary,omitempty"`
}
```

ravirajj:
This changes code generation to enable the `omitempty` JSON option for
optional collections to reduce `null` noise in the JSON output.

So,

    struct OptionalListField {
        1: optional list<Item> items
        2: optional map<Key, Value> dictionary
    }

Becomes,

    type OptionalListField struct {
      Items       []Item         `json:"items,omitempty"`
      Dictionary  map[Key]Value  `json:"dictionary,omitempty"`
    }

Encoding `OptionalListField{}` will now produce, `{}` instead of,

    {"items": null, "dictionary": null}

Resolves #330

### DIFF
--- a/compile/container.go
+++ b/compile/container.go
@@ -93,14 +93,11 @@ func (m *MapSpec) TypeCode() wire.Type {
 }
 
 // ForEachTypeReference for MapSpec
-func (m *MapSpec) ForEachTypeReference(f func(TypeSpec) error) (err error) {
-	if err = f(m.KeySpec); err != nil {
+func (m *MapSpec) ForEachTypeReference(f func(TypeSpec) error) error {
+	if err := f(m.KeySpec); err != nil {
 		return err
 	}
-	if err = f(m.ValueSpec); err != nil {
-		return err
-	}
-	return err
+	return f(m.ValueSpec)
 }
 
 // ThriftAnnotations returns all associated annotations.
--- a/compile/container.go
+++ b/compile/container.go
@@ -93,14 +93,14 @@ func (m *MapSpec) TypeCode() wire.Type {
 }
 
 // ForEachTypeReference for MapSpec
-func (m *MapSpec) ForEachTypeReference(f func(TypeSpec) error) error {
-	if err := f(m.KeySpec); err != nil {
+func (m *MapSpec) ForEachTypeReference(f func(TypeSpec) error) (err error) {
+	if err = f(m.KeySpec); err != nil {
 		return err
 	}
-	if err := f(m.ValueSpec); err != nil {
+	if err = f(m.ValueSpec); err != nil {
 		return err
 	}
-	return nil
+	return err
 }
 
 // ThriftAnnotations returns all associated annotations.
--- a/gen/field.go
+++ b/gen/field.go
@@ -70,32 +70,28 @@ func (f fieldGroupGenerator) checkReservedIdentifier(name string) error {
 	return nil
 }
 
-func (f fieldGroupGenerator) Generate(g Generator) (err error) {
-	if err = f.DefineStruct(g); err != nil {
+func (f fieldGroupGenerator) Generate(g Generator) error {
+	if err := f.DefineStruct(g); err != nil {
 		return err
 	}
 
-	if err = f.ToWire(g); err != nil {
+	if err := f.ToWire(g); err != nil {
 		return err
 	}
 
-	if err = f.FromWire(g); err != nil {
+	if err := f.FromWire(g); err != nil {
 		return err
 	}
 
-	if err = f.String(g); err != nil {
+	if err := f.String(g); err != nil {
 		return err
 	}
 
-	if err = f.Equals(g); err != nil {
+	if err := f.Equals(g); err != nil {
 		return err
 	}
 
-	if err = f.PrimitiveAccessors(g); err != nil {
-		return err
-	}
-
-	return err
+	return f.PrimitiveAccessors(g)
 }
 
 func (f fieldGroupGenerator) DefineStruct(g Generator) error {
--- a/gen/field.go
+++ b/gen/field.go
@@ -147,23 +147,24 @@ func generateTags(f *compile.FieldSpec) (string, error) {
 }
 
 func compileJSONTag(f *compile.FieldSpec, name string, opts ...string) *structtag.Tag {
-	// We want to add omitempty if the field is an optional struct or
-	// primitive to reduce "null" noise. We won't add omitempty for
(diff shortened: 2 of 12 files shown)


compile/container.go
```diff
@@ -93,14 +93,14 @@ func (m *MapSpec) TypeCode() wire.Type {
 }
 
 // ForEachTypeReference for MapSpec
-func (m *MapSpec) ForEachTypeReference(f func(TypeSpec) error) error {
-	if err := f(m.KeySpec); err != nil {
+func (m *MapSpec) ForEachTypeReference(f func(TypeSpec) error) (err error) {
```
kriskowal: The named return is inconsistent with the style below. We rarely use named returns, but when you do, the `return` statement receives no arguments. This also appears to be a style change unrelated to the PR.

gen/field.go
```diff
@@ -147,23 +147,24 @@ func generateTags(f *compile.FieldSpec) (string, error) {
...
 	}
 
+	// If the field name is "-" then it means omit, add no tags
 	if name == "-" {
-		// If the field name is "-" then it means omit, add no tags
 		return t
 	}
 
-	if (isStructType(f.Type) || isPrimitiveType(f.Type)) && !f.Required && !t.HasOption("omitempty") {
+	// We want to add the "omitempty" json tag if the field is an 'optional" to
```
kriskowal: Nit.

Either:

    is optional

Or

    is an "optional".

Also, mismatch single and double quote.

gen/field.go
```diff
@@ -147,23 +147,24 @@ func generateTags(f *compile.FieldSpec) (string, error) {
...
+	// reduce "null" noise. The "omitempty" json tag specifies that the field
```
kriskowal: capitalize JSON

gen/struct_test.go
```diff
@@ -877,112 +877,131 @@ func TestStructWithDefaults(t *testing.T) {
 
 func TestStructJSON(t *testing.T) {
 	tests := []struct {
-		v interface{}
-		j string
+		v, w interface{}
+		j    string
```
kriskowal: `v` is obviously short for `value`. `j` is also obviously for `JSON`. What does `w` mean? It might be useful to extract the three cases that make use of the `w` field from the table test, stand-alone. Using positional fields for a {k, v} duple is pretty clean.

gen/struct_test.go
```diff
@@ -992,10 +1011,13 @@ func TestStructJSON(t *testing.T) {
 			assert.Equal(t, tt.j, string(encoded))
 		}
 
-		v := reflect.New(reflect.TypeOf(tt.v).Elem()).Interface()
+		u := tt.v
+		if tt.w != nil {
+			u = tt.w
+		}
+		v := reflect.New(reflect.TypeOf(u).Elem()).Interface()
```
kriskowal: Also might be better to use full names here, since they are drifting from being abbreviations of meaningful names. `v` in scope no longer means value, but “type of value”. `w` appears to be an alternate value for purposes of inferring the type from a more complete object.

This would probably all be a lot more clear with the three relevant cases extracted to a clearer test with meaningful names.

gen/struct_test.go
```diff
@@ -877,112 +877,131 @@ func TestStructWithDefaults(t *testing.T) {
...
 			},
-			`{"name":"foo","contact":{"emailAddress":""}}`,
+			j: `{"name":"foo","contact":{"emailAddress":""}}`,
 		},
-		{&tu.EmptyUnion{}, "{}"},
-		{&tu.Document{Pdf: td.PDF("hello")}, `{"pdf":"aGVsbG8="}`},
-		{&tu.Document{Pdf: td.PDF{}}, `{"pdf":""}`},
+		{v: &tu.EmptyUnion{}, j: "{}"},
+		{v: &tu.Document{Pdf: td.PDF("hello")}, j: `{"pdf":"aGVsbG8="}`},
+		{v: &tu.Document{Pdf: td.PDF{}}, w: &tu.Document{Pdf: nil}, j: `{}`},
```
kriskowal: Does the previously written test case continue to pass? Might be better to just replace, remove, or move.

abhinav:
Thanks for the PR! Please address @kriskowal's comments and drop styling changes that are not relevant to this PR. Also, please rebase this change on dev. I've already updated the merge base for the PR to dev.

compile/container.go
ravirajj: This change was made to fix the lint error.  I'll update the style.

gen/field.go
ravirajj: Done.

gen/field.go
ravirajj: Done.

ravirajj:
- Rebased on `dev`.
- Added separate tests for optional empty list/map struct fields.
- `TestStructJSON` already covers optional nil and non-empty list/map struct fields.

compile/container.go
```diff
@@ -97,10 +97,7 @@ func (m *MapSpec) ForEachTypeReference(f func(TypeSpec) error) error {
 	if err := f(m.KeySpec); err != nil {
 		return err
 	}
-	if err := f(m.ValueSpec); err != nil {
-		return err
-	}
-	return nil
+	return f(m.ValueSpec)
```
kriskowal: Again with the unrelated style changes ;)

kriskowal:
Please get a stamp from @abhinav as well. We require two stamps for software networking projects.

compile/container.go
ravirajj: I had to make these changes to fix the lint errors!

abhinav: This is a lint error in Go 1.8 but not in 1.9. I changed Travis on this repo
to run for Go 1.9 only. This should no longer be a problem.

abhinav:
Merged. Thanks for the change!

PR #336, opened 2017-10-03
gen: Expose Ptr() method on enums

### LINKED ISSUE #335: Types: Map enums to enum thrift pointers
It would be nice if there were nice ways to map thrift enums to thrift enum pointers.

For example, if I have

```
enum ActionType {
    ENUM_ONE  = 1
    ENUM_TWO = 2 
    ENUM_THREE = 3
}
```

It would be nice if thrift-rw created a method called

```
func (a ActionType) Ptr() *ActionType {
    // Implementation here
}
```
Related: #231

hlin117:
This adds a `Ptr()` method to generated enum types which returns a pointer to
the enum value.

For example, given,

```
// action_type.thrift
enum ActionType {
    ENUM_ONE  = 1
    ENUM_TWO = 2 
    ENUM_THREE = 3
}
```

```
// main.go
import pkg "path.to.generated.types"

func DoSomething(aPtr *ActionType) {
    ...
}

func main() {
    DoSomething(&pkg.ActionTypeEnumOne)
}
```

The above will fail because `pkg.ActionTypeEnumOne` is a `const`. Go does not
allow taking pointers to `const` values.

With this PR, we can do,

```
// main.go
import pkg "path.to.generated.code"

func DoSomething(aPtr *ActionType) {
    ...
}

func main() {
    DoSomething(pkg.ActionTypeEnumOne.Ptr())
}
```

Resolves #335

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Releases
 v1.9.0 (unreleased)
 -------------------
 
--   No changes yet.
+-   Generates enum types now include a `Ptr()` method.
 
 
 v1.8.0 (2017-09-29)
--- a/gen/enum.go
+++ b/gen/enum.go
@@ -107,8 +107,8 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 			}
 		}
 
-		// ToPtr translates <$enumName> into its pointer representation.
-		func (<$v> <$enumName>) ToPtr() *<$enumName> {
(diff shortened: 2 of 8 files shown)


abhinav:
Made the following changes:

- Renamed the method to `Ptr()`
- Rebased on dev
- Updated the changelog

hlin117:
Woo! I'm happy I can contribute to this project. Thanks for the support @abhinav @breerly !

PR #338, opened 2017-10-27
 Enable marshal/unmarshal methods for i64 codegen

oliviazhang0809:
enable marshal and unmarshal for json annotation for Date/Long for i64


kriskowal:
Ignore above. It looks to me like this does in fact unmarshal {high, low} and number, detecting the format with the `{` or `"` prefix. 👍

compile/typedef.go
```diff
@@ -53,13 +55,17 @@ func compileTypedef(file string, src *ast.Typedef) (*TypedefSpec, error) {
 			Reason: err,
 		}
 	}
+	isLong := annotations["json.type"] == "Long"
+	isTimestamp := annotations["json.type"] == "Date"
```
abhinav: This isn't quite right. The annotation is not supported on the `typedef`
declaration but on the `i64`.

It's,

    typedef i64 (json.type = "Long") Timestamp

Not,

    typedef i64 Timestamp (json.type = "Long")

https://github.com/thriftrw/thriftrw-node#i64

So you should be looking at the annotations on the [I64Spec].

  [I64Spec]: https://godoc.org/go.uber.org/thriftrw/compile#I64Spec

Also, in the compile step for the I64Spec, you'll want to validate that only
valid values are provided for `json.type`. That's probably where you want to
associate this information with the I64Spec as a new `JSONType` field of type
I64JSONType`.

gen/testdata/thrift/structs.thrift
```diff
@@ -85,6 +85,15 @@ struct Graph {
 //////////////////////////////////////////////////////////////////////////////
 // Nested structs (Optional)
 
+typedef i64 Timestamp (json.type = "Date")
```
abhinav: This one is not allowed

compile/typedef.go
```diff
@@ -53,13 +55,17 @@ func compileTypedef(file string, src *ast.Typedef) (*TypedefSpec, error) {
...
 
 	return &TypedefSpec{
 		Name:        src.Name,
 		File:        file,
 		Target:      typ,
 		Annotations: annotations,
 		Doc:         src.Doc,
+		IsLong:      isLong,
```
AlexAPB: Instead of adding multiple booleans (which may result in future booleans being snuck into here too) could we use an enum instead? This would also ensure/make clear that isLong and isTimestamp are mutually exclusive.

AlexAPB: thoughts @abhinav ?

compile/typedef.go
Raynos: I double checked our internal usage; it looks like it's all `typedef i64 (json.type = ...)`

oliviazhang0809: @abhinav ^^

Raynos: That means that abhinav is correct.

We should not read annotations on the typedef directly; we should only read it based on the `Target` annotations in the `Link()` method.

abhinav: Clarification to @Raynos suggestion: I would rather have this information tracked at I64Spec compilation time. See my original comment for the general direction this should take.

abhinav:
Hello, the implementation needs to change significantly before this change can be merged. Please see https://github.com/thriftrw/thriftrw-go/pull/338#discussion_r147537397 and https://github.com/thriftrw/thriftrw-go/pull/338#discussion_r147537920 for suggestions on how this change should be structured.

abhinav:
Okay, so this change is a lot more complicated than this.

First, *anywhere* in the Thrift file where `i64` is referenced, it can have a
`json.type` annotation. This includes typedefs, structs, unions, and function
arguments. This is what thriftrw-node supports right now.

    typedef i64 (json.type = "Long") Long
    
    struct Foo {
        1: optional i64 (json.type = "Date") date
    }

When ThriftRW is compiling the Thrift file, it builds an I64Spec for *each*
such reference. So the code that compiles I64Specs can interpret the
`json.type` annotation and store this information on the I64Spec object. This
is also where you will validate that the value for `json.type` is recognized
and supported. Doing it like this covers all cases where i64 could have an
annotation.

Additionally, we need a library component in ThriftRW itself which represents
the different ways in which `i64` can be serialized. You will probably declare
the following two types,

    type Timestamp i64
    type Date i64

Each with its own `MarshalJSON` and `UnmarshalJSON` methods for that
`json.type`. This will be used later.

Codegen must now be augmented to include MarshalJSON and UnmarshalJSON for
**all** user-defined types. So StructSpec and TypedefSpec need to be updated
to generate MarshalJSON and UnmarshalJSON (EnumSpec already has the methods).

-   For typedefs,
    -   if this is a user-defined type, cast and delegate to the underlying
        MarshalJSON/UnmarshalJSON implementation.
    -   for Thrift-native types, rely on `json.Marshal`
    -   for i64 with a valid `json.type`, cast to Timestamp/Long (implemented
        above) and delegate to that MarshalJSON/UnmarshalJSON implementation.
-   For structs, we have two options:
    -   Implement MarshalJSON and UnmarshalJSON ourselves. This seems very
        risky and error prone.
    -   Build an alternative internal representation of the struct and call
        `json.Marshal/Unmarshal` on that. So, given,
        
            type User struct {
                Name string
                Email *string
                JoinDate *i64  // annotated with json.type = "Date"
            }
            
        We will generate,
        
            type _json_User struct {
                Name string
                Email *string
                JoinDate *thriftrw.Date
            }
        
        Then `User.MarshalJSON` maps the `User` to `_json_User` and calls
        `json.Marshal` on that. Similarly, `User.UnmarshalJSON` calls
        `json.Unmarshal` on a `_json_User` and maps the fields back to this
        `User` object.

Now I understand that this complicates the task significantly more than
previously anticipated, but this is the most straightforward way I see of
implementing this change correctly. Supporting just typedefs won't be enough.
I just checked; your team owns many Thrift files with the `js.type` annotation
on struct fields, and teams outside yours own even a larger number of such
Thrift files.

Given the complexity of this change, it may also be worth exploring other
options. Obvious options are,

-   You're already using easyjson IIRC. Perhaps a plugin that makes easyjson
    behave differently based on annotations could be written.
-   Ensure that all your Thrift files are using typedefs for `i64`s
    (completely backwards compatible), and generate `MarshalJSON` and
    `UnmarshalJSON` for *just* those typedefs from a plugin. It's a matter of
    placing a Go file in the same directory as the generated code.

CC @Raynos @kriskowal

abhinav:
Closing as discussed with @Raynos.

Raynos:
For reference this was implemented in https://github.com/uber/zanzibar/blob/master/scripts/resolve_i64/main.go

PR #339, opened 2017-12-01
Envelope agnostic binary protocol

kriskowal:
This adds a DecodeRequest method to the Binary protocol, making it capable of
automatically inferring whether a request buffer contains an envelope and
decoding the body regardless.

It goes on to add a RequestProtocol interface to extend the exisiting Protocol
interface, including DecodeRequest.


protocol/binary.go
```diff
@@ -61,3 +61,30 @@ func (binaryProtocol) DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error) {
 	e, err := reader.ReadEnveloped()
 	return e, err
 }
+
+func (b binaryProtocol) DecodeRequest(r io.ReaderAt) (wire.Value, error) {
```
breerly: This probably deserves a nice description an potentially a background into why this is useful/necessary :)

kriskowal: So it is done.

kriskowal:
I’ve added changes…did not modify the type of Binary, added EnvelopeAgnosticBinary instead. Added a return enum to DecodeRequest to indicate what kind of envelope was around the request so the RPC layer can respond in kind.

protocol/binary.go
```diff
@@ -61,3 +67,53 @@ func (binaryProtocol) DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error) {
...
+	// 0x80 is a valid preamble for a strict enveloped request.
+	if buf[0]&0x80 > 0 {
+		e, err := b.DecodeEnveloped(r)
+		if err != nil {
+			return wire.Value{}, NoMessage, err
+		}
+		return e.Value, VersionedMessage, nil
+	}
+
+	// All other patterns are either bare structs or invalid.
```
prashantv: rule 4 was,
```
	// 4. A message of length >1 starting with 0x00-0x0f can only be the
	// beginning of an un-enveloped field.
```

However, we're assuming here that anything where `buf[0] != 0` must be a non-enveloped struct. Isn't it possible it's an enveloped non-strict value with a long string as the name?

If we're ignoring that case, then it seems like we should very explicitly call it out.

protocol/binary.go
```diff
@@ -61,3 +67,53 @@ func (binaryProtocol) DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error) {
 	e, err := reader.ReadEnveloped()
 	return e, err
 }
+
+// DecodeRequest specializes Decode and replaces DecodeEnveloped for the
```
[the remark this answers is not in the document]
kriskowal: I’ve clarified that the request is a struct. Requests and responses are always implicit TStructs on the wire. There is nowhere in the protocol to specify otherwise.

protocol/binary_test.go
```diff
@@ -678,21 +696,32 @@ func TestBinaryEnvelopeErrors(t *testing.T) {
 
 	for _, tt := range tests {
 		reader := bytes.NewReader(tt.encoded)
-		_, err := Binary.DecodeEnveloped(reader)
+		_, err := EnvelopeAgnosticBinary.DecodeEnveloped(reader)
```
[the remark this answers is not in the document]
kriskowal: Need not, should not, fixed.

protocol/binary_test.go
```diff
@@ -776,6 +808,24 @@ func TestBinaryEnvelopeSuccessful(t *testing.T) {
...
+		if !assert.Equal(t, tt.want.Value, r, "%v: decoded request mismatch", tt.msg) {
+			continue
+		}
+
+		if !assert.Equal(t, tt.wantEnveloping, env, "%v: decoded request should have envelope") {
+			continue
+		}
+
+		// Verify a round trip, back from encode after decode.
+
```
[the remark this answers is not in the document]
kriskowal: I’d like to keep the blank. A comment without a blank I infer applies only to the immediate following block, a comment with a blank applies until the next comment with a following blank. In this case, this comment applies to the remaining two checks, which have a blank between them.

protocol/binary.go
```diff
@@ -61,3 +82,108 @@ func (binaryProtocol) DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error) {
...
+// specific purpose of decoding request structs that may or may not have an
+// envelope.
+// This allows a Thrift request handler to transparently accept requests
+// regardless of whether the caller submits an envelope.
+func (b binaryProtocol) DecodeRequest(r io.ReaderAt) (wire.Value, EnvelopeSpecificResponder, error) {
+	// Strip request envelopes if present.
+	// 1. A message of length 1 containing only 0x00 can only be a bare,
+	// un-enveloped empty struct.
+	// 2. A message of length >1 starting with 0x00 can only be enveloped
+	// without a version (non-strict)
```
prashantv: can you provide rationale here. e.g., why can't it be a valid struct (since field type starts a struct, and 0 is not a valid field type).

protocol/binary.go
```diff
@@ -61,3 +82,108 @@ func (binaryProtocol) DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error) {
...
+	// All other patterns are either bare structs or invalid.
+	val, err := b.Decode(r, wire.TStruct)
+	if err != nil {
+		return wire.Value{}, noEnvelopeResponder, err
+	}
+	return val, noEnvelopeResponder, nil
+}
+
+// noEnvelopeResponder responds to a request without an envelope.
+
```
prashantv: nit: comment should be right above the declaration.

protocol/binary.go
```diff
@@ -61,3 +82,108 @@ func (binaryProtocol) DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error) {
...
+type _noEnvelopeResponder struct{}
+
+func (_noEnvelopeResponder) EncodeResponse(v wire.Value, t wire.EnvelopeType, w io.Writer) error {
```
prashantv: should we go the other way (underscore prefix for the variable, and not the type, since that also matches our internal style guide)

protocol/binary.go
```diff
@@ -61,3 +82,108 @@ func (binaryProtocol) DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error) {
...
+		if err != nil {
+			return wire.Value{}, noEnvelopeResponder, err
+		}
+		return e.Value, &envelopeV1Responder{
+			Name:  e.Name,
+			SeqID: e.SeqID,
+		}, nil
+	}
+
+	// All other patterns are either bare structs or invalid.
```
prashantv: we should call out why this isn't filtering on just the `0x00-0x0f` which is what the rule called out. specifically that we're OK with assuming that we don't see large procedure fields?

protocol/binary.go
```diff
@@ -61,3 +82,108 @@ func (binaryProtocol) DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error) {
...
+
+type _noEnvelopeResponder struct{}
+
+func (_noEnvelopeResponder) EncodeResponse(v wire.Value, t wire.EnvelopeType, w io.Writer) error {
+	return Binary.Encode(v, w)
+}
+
+var noEnvelopeResponder EnvelopeSpecificResponder = &_noEnvelopeResponder{}
+
+// envelopeV0Responder responds to requests with a non-strict (unversioned) envelope.
```
prashantv: the comment should go right above the declaration without a blank line.

protocol/binary/envelope.go
```diff
@@ -52,6 +51,24 @@ func (bw *Writer) WriteEnveloped(e wire.Envelope) error {
 	return bw.WriteValue(e.Value)
 }
 
+// WriteLegacyEnveloped writes enveloped value using the non-strict envelope
+// (non-strict lacks an envelope version).
+func (bw *Writer) WriteLegacyEnveloped(e wire.Envelope) error {
```
prashantv: should we export this? can we just keep it private since we're pretty confident there isn't a good reason for users to use this?

protocol/binary_test.go
```diff
@@ -788,3 +821,219 @@ func TestBinaryEnvelopeSuccessful(t *testing.T) {
 		assert.Equal(t, tt.encoded, buf.Bytes(), "%v: reencoded bytes mismatch")
 	}
 }
+
+func TestReqRes(t *testing.T) {
+
```
prashantv: nit: delete blank

protocol/binary_test.go
```diff
@@ -788,3 +821,219 @@ func TestBinaryEnvelopeSuccessful(t *testing.T) {
...
+				vfield(1, vi16(42)),
+				vfield(2, vlist(wire.TBinary, vbinary("foo"), vbinary("bar"))),
+				vfield(3, vset(wire.TBinary, vbinary("baz"), vbinary("qux"))),
+			),
+			reqBytes: []byte{
+				0x06,       // type:1 = i16
+				0x00, 0x01, // id:2 = 1
+				0x00, 0x2a, // value = 42
+
+				0x0F,       // type:1 = list
```
prashantv: can we avoid manually encoding these bytes, and use the existing binary encoding? prefer to avoid re-testing the struct binary encoding here, which we already test elsewhere. the binary that is relevant here is the envelope, the rest should be `encode(vstruct(..))`

protocol/binary_test.go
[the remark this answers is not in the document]
breerly: Makes me wonder if those sections should be encapsulated into a single method/func... likely thinking about it too hard.

protocol/binary.go
```diff
@@ -21,17 +21,45 @@
...
+
+type errUnexpectedEnvelopeType wire.EnvelopeType
+
+func (e errUnexpectedEnvelopeType) Error() string {
+	return fmt.Sprintf("unexpected envelope type: %v", wire.EnvelopeType(e))
+}
+
 func init() {
 	Binary = binaryProtocol{}
+	EnvelopeAgnosticBinary = binaryProtocol{}
```
breerly: Great naming.

protocol/binary.go
```diff
@@ -21,17 +21,45 @@
...
 import (
+	"fmt"
 	"io"
 
 	"go.uber.org/thriftrw/protocol/binary"
 	"go.uber.org/thriftrw/wire"
 )
 
 // Binary implements the Thrift Binary Protocol.
+// Binary can be cast up to EnvelopeAgnosticProtocol to support DecodeRequest.
```
breerly: ty

protocol/binary.go
breerly: +1

protocol/binary_test.go
```diff
@@ -683,16 +684,27 @@ func TestBinaryEnvelopeErrors(t *testing.T) {
 			continue
 		}
 
+		// Also verify that DecodeRequest fails.
```
breerly: Thinking outloud - should these be in their own test?

protocol/binary_test.go
```diff
@@ -776,6 +791,24 @@ func TestBinaryEnvelopeSuccessful(t *testing.T) {
 			continue
 		}
 
+		// Also verify whether we can infer the presence of the envelope
```
breerly: Same thing - wondering if the envelope agnostic items should be in their own test.

protocol/binary_test.go
```diff
@@ -788,3 +821,219 @@ func TestBinaryEnvelopeSuccessful(t *testing.T) {
 		assert.Equal(t, tt.encoded, buf.Bytes(), "%v: reencoded bytes mismatch")
 	}
 }
+
+func TestReqRes(t *testing.T) {
```
breerly: Suggest renaming to describe the behavior being verified... perhaps `TestEnvelopeAgnosticism`...

protocol/binary/envelope.go
kriskowal: In this case, we’re using protocol/binary.Writer from the protocol package, so this does need to be public. We might have had the foresight to put this in the same package or an internal package, but sadly missed that opportunity.

protocol/protocol.go
```diff
@@ -44,3 +44,28 @@ type Protocol interface {
 	// Enveloped values are assumed to be TStructs.
 	DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error)
 }
+
+// EnvelopeAgnosticProtocol defines a specific way for a Thrift value to be
+// encoded or decoded, additionally being able to decode requests without prior
+// knowledge of whether the request is enveloped.
+type EnvelopeAgnosticProtocol interface {
```
breerly: Might be worth also adding a comment about upcasting Protocol to this interface here.

protocol/binary.go
```diff
@@ -61,3 +89,116 @@ func (binaryProtocol) DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error) {
...
+// envelope.
+// This allows a Thrift request handler to transparently accept requests
+// regardless of whether the caller submits an envelope.
+// The caller specifies the expected envelope type, one of OneWay or Unary, on
+// which the decoder asserts if the envelope is present.
+func (b binaryProtocol) DecodeRequest(et wire.EnvelopeType, r io.ReaderAt) (wire.Value, EnvelopeSpecificResponder, error) {
+	// Strip request envelopes if present.
+	// 1. A message of length 1 containing only 0x00 can only be a bare,
+	// un-enveloped empty struct.
+	// 2. A message of length >1 starting with 0x00 can only be enveloped
```
abhinav: I think the explanation should include the actual expected shapes of the
payloads.

```
Enveloped (strict):
    version:2 0x00 envType:1 name~4 id:4
    where version = 0x8001
Enveloped (non-strict):
    name~4 envType:1 id:4
Non-enveloped:
    fieldType:1 (fieldID:2 fieldData:? fieldType:1)*
    when fieldType = 0, no more information follows
```

Plus it can be more obvious.

```
A message can be,

1.  Enveloped (strict)
    
    The message must begin with 0x80 -- the first byte of the version number.

2.  Enveloped (non-strict)

    The message begins with the length of the method name. As long as the
    method name is not 16 MB long, the first byte of this message is 0x00. If
    the message contains at least 9 other bytes, it is enveloped using the
    non-strict format.

        4  bytes     message length (minimum = 0)
        1  byte      message type
        4  bytes     sequence ID
        1  byte      empty struct (0x00)
        --------
        10 bytes
        
3.  Unenveloped

    If the message begins with 0x00 but does not contain any more bytes, it's
    a bare, un-enveloped empty struct.

    If the message begins with any other byte, it's an unenveloped message or
    an invalid request.
```

protocol/binary.go
```diff
@@ -61,3 +89,116 @@ func (binaryProtocol) DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error) {
...
+	// without a version (non-strict)
+	// 3. A message of length >1 with MSB 0x80 set can only be enveloped with a
+	// version (strict)
+	// 4. A message of length >1 starting with 0x00-0x0f can only be the
+	// beginning of an un-enveloped field.
+	var buf [2]byte
+	// If we fail to read two bytes, the only possible valid value is the empty struct.
+	if _, err := r.ReadAt(buf[0:2], 0); err != nil {
```
abhinav: Instead of relying on `err != nil`, it may be preferable to rely on the first return value of `ReadAt`. (We're ignoring the error anyway.)

```lang=go
if n, _ := r.ReadAt(buf[:2], 0); n < 2 {
    ...
}
```

protocol/binary/envelope.go
abhinav: I don't feel strongly about this but as discussed offline,
WriteLegacyEnveloped isn't necessary because protocol implementations are
expected to be able to decode both, strict and non-strict envelopes regardless
of what they're sending.

Looking at the Apache Thrift Python library implementation, non-strict
envelopes are rejected only if the user explicitly opted into it. Strict
envelopes (which is what ThriftRW Go uses) are never rejected, even if the
user opted into writing non-strict envelopes.

protocol/binary.go
```diff
@@ -21,17 +21,90 @@
...
+//
+//  2.  Enveloped (non-strict)
+//
+//      The message begins with the length of the method name. As long as the
+//      method name is not 16 MB long, the first byte of this message is 0x00. If
+//      the message contains at least 9 other bytes, it is enveloped using the
+//      non-strict format.
+//
+//          4  bytes     method name length (minimum = 0)
+//			n  bytes	 method name
```
abhinav: alignment

protocol/binary.go
```diff
@@ -21,17 +21,90 @@
...
+//          --------
+//          >10 bytes
+//
+//  3.  Unenveloped
+//
+//      If the message begins with 0x00 but does not contain any more bytes, it's
+//      a bare, un-enveloped empty struct.
+//
+//      If the message begins with any other byte, it's an unenveloped message or
+//      an invalid request./
```
abhinav: trailing `/`

protocol/binary.go
```diff
@@ -61,3 +134,133 @@ func (binaryProtocol) DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error) {
...
+// struct starting with that field identifier. Valid field identifiers today
+// are in the range 0x00-0x0f. There is some chance that a future version of
+// the protocol will add more field types, but it is very unlikely that the
+// field type will flow into the MSB (128 type identifiers, starting with the
+// 15 valid types today).
+func (b binaryProtocol) DecodeRequest(et wire.EnvelopeType, r io.ReaderAt) (wire.Value, EnvelopeSpecificResponder, error) {
+	var buf [2]byte
+
+	// If we fail to read two bytes, the only possible valid value is the empty struct.
+	if _, err := r.ReadAt(buf[0:2], 0); err != nil {
```
abhinav: repeating nit because comment was hidden by GH:
Probably better to use the integer return value since there are other reasons
for `err` to be non-nil.

```go
if n, err := r.ReadAt(...); n == 1 {
    // Only empty struct possible.
} else if err != nil {
    // Read error occurred.
    return wire.Value{}, nil, err
}
```

protocol/binary.go
```diff
@@ -61,3 +134,133 @@ func (binaryProtocol) DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error) {
...
+		}, nil
+	}
+
+	// All other patterns are either bare structs or invalid.
+	// We delegate to the struct decoder to distinguish invalid type
+	// identifiers, outside the 0-15 range.
+	val, err := b.Decode(r, wire.TStruct)
+	if err != nil {
+		return wire.Value{}, _noEnvelopeResponder, err
+	}
```
abhinav: optional nitty nit: can simplify

    val, err := b.Decode(...)
    return val, _noEnvelopeResponder, err

protocol/binary/envelope.go
abhinav: I buy it.

protocol/binary_test.go
```diff
@@ -776,6 +791,24 @@ func TestBinaryEnvelopeSuccessful(t *testing.T) {
...
+		r, responder, err := EnvelopeAgnosticBinary.DecodeRequest(tt.want.Type, reader)
+		if !assert.NoError(t, err, "%v: failed to decode request with envelope", tt.msg) {
+			continue
+		}
+
+		if !assert.Equal(t, tt.want.Value, r, "%v: decoded request mismatch", tt.msg) {
+			continue
+		}
+
+		if !assert.True(t, tt.wantResponderType == reflect.TypeOf(responder), "%v: decoded request should have responder want %v got %T", tt.msg, tt.wantResponderType, responder) {
```
abhinav: optional: can `assert.Equal` on `reflect.Type`s

```go
assert.Equal(t, tt.wantResponderType, reflect.TypeOf(responder))
```

protocol/binary_test.go
```diff
@@ -788,3 +821,137 @@ func TestBinaryEnvelopeSuccessful(t *testing.T) {
...
 }
+
+func tbinary(v wire.Value) []byte {
+	buf := &bytes.Buffer{}
+	Binary.Encode(v, buf)
+	return buf.Bytes()
+}
+
+func TestReqRes(t *testing.T) {
+
```
abhinav: nit: delete empty line at start of function

I think I read somewhere that upcoming versions of golint will complain about this

protocol/binary_test.go
```diff
@@ -788,3 +821,137 @@ func TestBinaryEnvelopeSuccessful(t *testing.T) {
...
+		if !assert.True(t, reflect.TypeOf(reser) == tt.responderType, "%s: responder type mismatch", tt.msg) {
+			continue
+		}
+
+		if !assert.True(t, wire.ValuesAreEqual(tt.req, req), "%s: decoded request mismatch", tt.msg) {
+			continue
+		}
+
+		writer := bytes.NewBuffer(make([]byte, 0, 0))
```
abhinav: `new(bytes.Buffer)`?

protocol/binary.go
```diff
@@ -61,3 +134,133 @@ func (binaryProtocol) DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error) {
...
+func (b binaryProtocol) DecodeRequest(et wire.EnvelopeType, r io.ReaderAt) (wire.Value, EnvelopeSpecificResponder, error) {
+	var buf [2]byte
+
+	// If we fail to read two bytes, the only possible valid value is the empty struct.
+	if _, err := r.ReadAt(buf[0:2], 0); err != nil {
+		val, err := b.Decode(r, wire.TStruct)
+		if err != nil {
+			return wire.Value{}, _noEnvelopeResponder, err
+		}
+		return val, _noEnvelopeResponder, nil
```
prashantv: optional, can this be simplified to
```lang=go
val, err := b.Decode(r, wire.TStruct)
return val, _noEnvelopeResponder, err
```

protocol/binary.go
prashantv: `ReadAt` returns a non-nil error if the number of bytes read is `< len(buf)`:
> ReadAt reads len(p) bytes into p starting at offset off in the underlying input source. It returns the number of bytes read (0 <= n <= len(p)) and any error encountered.
>
>When ReadAt returns n < len(p), it returns a non-nil error explaining why more bytes were not returned. In this respect, ReadAt is stricter than Read.

protocol/binary/envelope.go
prashantv: The logic to detect the type of envelope (if any) seems very tied to the `binary` protocol -- should all of that logic live inside of the `binary` package? That way `WriteLegacyEnvelope` is not exported, but we still support it for this specific use case.

protocol/binary.go
abhinav: `ReadAt` could read zero bytes because of some IO error. That will be treated as a successful read of an empty struct right now.

protocol/binary_test.go
kriskowal: went with &bytes.Buffer{}

kriskowal:
@prashantv I agree that putting all the binary protocol bits in binary would be good, but I’m making the call to avoid a yak shave into restructuring the existing code.

Taking everything else ;)

protocol/binary.go
```diff
@@ -221,15 +221,16 @@ func (noEnvelopeResponder) EncodeResponse(v wire.Value, t wire.EnvelopeType, w i
 	return Binary.Encode(v, w)
 }
 
-var _noEnvelopeResponder EnvelopeSpecificResponder = &noEnvelopeResponder{}
+// NoEnvelopeResponder responds to a request without an envelope.
+var NoEnvelopeResponder EnvelopeSpecificResponder = &noEnvelopeResponder{}
```
abhinav: nit: as discussed offline, I think we should call the interface just Responder

PR #340, opened 2018-01-04
Remove version check

kriskowal:
This removes version checks from code generation, and no-ops the version check
library. Version checks put an undue burden on dependees, requiring code
regeneration after updating even to a backward-compatible release of thriftrw.


version/no_check.go
```diff
@@ -18,24 +18,10 @@
...
-// Rationale: It is possible that the old generated code is not compatible with
-// the new version of ThriftRW in subtle ways but still compiles successfully.
-// This function will ensure that the version mismatch is detected and help
-// avoid bugs that could be caused by this discrepancy.
-func CheckCompatWithGeneratedCodeAt(string, string) {
-	// This version does not do anything. It is used only when the
-	// disableVersionCheck is used. Use the following command to build a
-	// binary with this tag.
-	//
-	// 	go build -tags thriftrw.disableVersionCheck
```
abhinav: As a result of this, you can also delete the `-tags` business from,

https://github.com/thriftrw/thriftrw-go/blob/c6d9372002b8de6db0e9ee21549c2e2d83304889/Makefile#L65
https://github.com/thriftrw/thriftrw-go/blob/c6d9372002b8de6db0e9ee21549c2e2d83304889/gen/testdata/Makefile#L15
https://github.com/thriftrw/thriftrw-go/blob/c6d9372002b8de6db0e9ee21549c2e2d83304889/internal/plugin/gen.sh#L27

abhinav:
Let's get at least one other person from the team to also have a look. CC @prashantv @breerly

peter-edge:
+1000

peter-edge:
Can we remove the version number being included in the generated code as well?

abhinav:
We should keep the version number in the generated code. It helps to know which version of ThriftRW was used to generate some code when debugging issues.

version/no_check.go
```diff
@@ -18,24 +18,10 @@
...
-}
+// CheckCompatWithGeneratedCodeAt is a relic from when we attempted to verify
+// that generated code was installed with the vendored version of thriftrw.
+// It had a troubled past, forcing dependees to regenerate code whenever
```
prashantv: I think the version check makes sense in one direction: disallowing generated code from version N+1 to work with library version N. However, the other direction (which is what this comment calls out) doesn't really make sense: If you generate code at version N, then library versions N+1 (within the same major) should be compatible.

Is the issue here that we were doing the wrong check?

kriskowal: We are doing the wrong check, so every `glide up` must be followed by codegen. This is particularly bad because codegen tends to also update mocks, which in turn leak implementation details that would not otherwise constitute a breaking change.

The current check verifies that the code was not generated with a prior release, and at that, appears to have a bug since 1.9.0 does not satisfy 1.9.0.

This is a third-order yak shave. My inclination is to remove the check and introduce the correct check later.

prashantv: Sure I'm ok with that. If it's easy to flip the version check to do the opposite, I'd prefer that, but I'm OK with whatever approach is simplest for now.

PR #351, opened 2018-04-23
Add accessors for all struct fields

amckinney:
This removes the constraint for generating struct field accessors for primitive types only.

Previously, the following struct:
```
type Foo struct {
    Bar *int32
    Baz []string
}
```

Would generate a struct with a `GetBar` method, but would omit a `GetBaz` method because it was not a primitive type. Now, we would expect to see both of these methods.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- Add accessors for all struct fields.
+- Structs now include getter functions for all fields.
 
 ## [1.11.0] - 2018-03-27
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- Structs now include getter functions for all fields.
+- Structs now include getter functions for all fields. This
+  improves Apache Thrift compatibility.
 
 ## [1.11.0] - 2018-03-27
 ### Added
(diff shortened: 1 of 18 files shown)


CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- No changes yet.
+- Add accessors for all struct fields.
```
abhinav: nit:

    Structs now include getter functions for all fields.

gen/field.go
```diff
@@ -473,17 +473,23 @@ func (f fieldGroupGenerator) PrimitiveAccessors(g Generator) error {
 			<$fname := goName .>
 			<reserveFieldOrMethod $fname>
 			<reserveFieldOrMethod (printf "Get%v" $fname)>
-			<if and (not .Required) (isPrimitiveType .Type)>
 			// Get<$fname> returns the value of <$fname> if it is set or its
 			// zero value if it is unset.
```
abhinav: I know this is from before but the comment is not quite right since we return the default value too when it's unset.

Can we make the comment mention that the default will be returned if unset. Something like,

    // Get<$fname> returns the value of <$fname> if it is set or its
    // <if .Default>default<else>zero<end> value if it is unset.

amckinney: Sounds good, I'll update it here.

CHANGELOG.md
amckinney: 👍

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- No changes yet.
+- Structs now include getter functions for all fields.
```
blampe: Include a reason for the new feature -- for improved Apache compatibility?

amckinney:
Failure here:
```
go get -u -f github.com/golang/lint/golint
package golang.org/x/lint: unrecognized import path "golang.org/x/lint" (parse https://golang.org/x/lint?go-get=1: no go-import meta tags ())
```

PR #354, opened 2018-05-21
Add tool for listing imported thrift files

jacobgreenleaf:
Adds a tool that given a thrift file outputs the list of thrift files that the given file depends. This can be used for build tools that use file-based dependency rules.

### DIFF
--- a/cmd/thriftrw-list-deps/README.md
+++ b/cmd/thriftrw-list-deps/README.md
@@ -1,0 +1,17 @@
+# list-deps
+
+This tool can be used on a given Thrift file to output the list of Thrift files that the given file depends on.
+
+## Installation
+
+```bash
+$ go get go.uber.org/thriftrw/cmd/thriftrw-list-deps
+```
+
+## Usage
+
+```bash
+$ thriftrw-list-deps -relative=$(pwd) gen/testdata/thrift/structs.thrift
+gen/testdata/thrift/enums.thrift
+$
+```
--- a/cmd/thriftrw-list-deps/README.md
+++ b/cmd/thriftrw-list-deps/README.md
@@ -1,4 +1,4 @@
-# list-deps
+# thriftrw-list-deps
 
 This tool can be used on a given Thrift file to output the list of Thrift files that the given file depends on.
 
@@ -11,7 +11,7 @@ $ go get go.uber.org/thriftrw/cmd/thriftrw-list-deps
 ## Usage
 
 ```bash
-$ thriftrw-list-deps --relative=$(pwd) gen/testdata/thrift/structs.thrift
+$ thriftrw-list-deps --relative-to=$(pwd) gen/testdata/thrift/structs.thrift
 gen/testdata/thrift/enums.thrift
 $
 ```
--- a/cmd/thriftrw-list-deps/README.md
+++ b/cmd/thriftrw-list-deps/README.md
@@ -11,7 +11,7 @@ $ go get go.uber.org/thriftrw/cmd/thriftrw-list-deps
 ## Usage
 
 ```bash
-$ thriftrw-list-deps -relative=$(pwd) gen/testdata/thrift/structs.thrift
+$ thriftrw-list-deps --relative=$(pwd) gen/testdata/thrift/structs.thrift
 gen/testdata/thrift/enums.thrift
 $
 ```
--- a/cmd/thriftrw-list-deps/main.go
+++ b/cmd/thriftrw-list-deps/main.go
@@ -2,9 +2,7 @@ package main
 
 import (
 	"fmt"
-	"io"
 	"log"
-	"os"
 	"path/filepath"
 
 	"github.com/jessevdk/go-flags"
@@ -13,44 +11,65 @@ import (
 )
 
 type args struct {
-	ThriftFile string `positional-arg-name:"thrift-file"`
+	ThriftFile string `positional-arg-name:"file" description:"Path to the Thrift file"`
 }
 
 var opts struct {
-	Relative string `long:"relative" description:"Output paths will be relative to this directory"`
-	Args     args   `positional-args:"yes" required:"yes"`
+	RelativeTo string `long:"relative-to" description:"If specified, output paths will be relative to this directory"`
+	Args       args   `positional-args:"yes" required:"yes"`
 }
 
-func do(relative string, inputThriftFile string, output io.Writer) error {
+// listDependentThrifts lists the Thrift files that the given Thrift file depends on or includes, both directly (through imports) and
+// indirectly (through transitive imports).
+//
+// The returned file paths are absolute, unless relativeDir parameter is given, in which case the paths are relative to
+// the relativeDir.
+func listDependentThrifts(inputThriftFile string, relativeDir string) ([]string, error) {
+	dependentThriftFiles := make([]string, 0)
+
 	module, err := compile.Compile(inputThriftFile)
 	if err != nil {
-		return err
+		return nil, fmt.Errorf("could not compile %q: %v", inputThriftFile, err)
 	}
 
 	for _, included := range module.Includes {
 		outputFilePath := included.Module.ThriftPath
-		if relative != "" {
-			outputFilePath, err = filepath.Rel(relative, outputFilePath)
+
+		if relativeDir != "" {
+			outputFilePath, err = filepath.Rel(relativeDir, outputFilePath)
 			if err != nil {
-				return err
+				return nil, fmt.Errorf("cannot make %q relativeDir to %q: %v", outputFilePath, relativeDir, err)
 			}
 		}
-		fmt.Fprintf(output, "%s\n", outputFilePath)
+
+		dependentThriftFiles = append(dependentThriftFiles, outputFilePath)
+	}
+
+	return dependentThriftFiles, nil
+}
+
+func run() error {
+	if _, err := flags.Parse(&opts); err != nil {
+		return fmt.Errorf("error parsing arguments: %v", err)
+	}
+
+	inputThriftFile := opts.Args.ThriftFile
+
+	paths, err := listDependentThrifts(inputThriftFile, opts.RelativeTo)
+	if err != nil {
+		return fmt.Errorf("error listing deps of %q: %v", inputThriftFile, err)
+	}
+
+	for _, path := range paths {
+		fmt.Println(path)
 	}
 
 	return nil
 }
 
 func main() {
-	if _, err := flags.Parse(&opts); err != nil {
-		log.Fatalf("error parsing arguments: %s", err.Error())
-		os.Exit(1)
-	}
-
-	inputThriftFile := opts.Args.ThriftFile
-
-	if err := do(opts.Relative, inputThriftFile, os.Stdout); err != nil {
-		log.Fatalf("error listing deps of %s: %s", inputThriftFile, err.Error())
-		os.Exit(1)
+	log.SetFlags(0) // so that the error message isn't noisy
+	if err := run(); err != nil {
+		log.Fatalf("%v", err)
 	}
 }
--- a/cmd/thriftrw-list-deps/main.go
+++ b/cmd/thriftrw-list-deps/main.go
@@ -30,7 +30,7 @@ func listDependentThrifts(input string, relativeTo string) ([]string, error) {
 		return nil, fmt.Errorf("could not compile %q: %v", input, err)
 	}
 
-	module.Walk(func(mod *compile.Module) error {
+	err = module.Walk(func(mod *compile.Module) error {
 		output := mod.ThriftPath
 
 		// Do not return a self-referencing dependency
@@ -49,6 +49,10 @@ func listDependentThrifts(input string, relativeTo string) ([]string, error) {
 		deps = append(deps, output)
 		return nil
 	})
+
+	if err != nil {
+		return nil, err
+	}
 
 	return deps, nil
 }
--- a/cmd/thriftrw-list-deps/main.go
+++ b/cmd/thriftrw-list-deps/main.go
@@ -10,42 +10,47 @@ import (
 	"go.uber.org/thriftrw/compile"
 )
 
-type args struct {
-	ThriftFile string `positional-arg-name:"file" description:"Path to the Thrift file"`
-}
-
 var opts struct {
 	RelativeTo string `long:"relative-to" description:"If specified, output paths will be relative to this directory"`
-	Args       args   `positional-args:"yes" required:"yes"`
+	Args       struct {
+		ThriftFile string `positional-arg-name:"file" description:"Path to the Thrift file"`
+	} `positional-args:"yes" required:"yes"`
 }
 
 // listDependentThrifts lists the Thrift files that the given Thrift file depends on or includes, both directly (through imports) and
 // indirectly (through transitive imports).
 //
-// The returned file paths are absolute, unless relativeDir parameter is given, in which case the paths are relative to
-// the relativeDir.
-func listDependentThrifts(inputThriftFile string, relativeDir string) ([]string, error) {
-	dependentThriftFiles := make([]string, 0)
+// The returned file paths are absolute, unless relativeTo parameter is given, in which case the paths are relative to
+// the relativeTo directory.
+func listDependentThrifts(input string, relativeTo string) ([]string, error) {
+	var deps []string
 
-	module, err := compile.Compile(inputThriftFile)
+	module, err := compile.Compile(input)
 	if err != nil {
-		return nil, fmt.Errorf("could not compile %q: %v", inputThriftFile, err)
+		return nil, fmt.Errorf("could not compile %q: %v", input, err)
 	}
 
-	for _, included := range module.Includes {
-		outputFilePath := included.Module.ThriftPath
+	module.Walk(func(mod *compile.Module) error {
+		output := mod.ThriftPath
 
-		if relativeDir != "" {
-			outputFilePath, err = filepath.Rel(relativeDir, outputFilePath)
+		// Do not return a self-referencing dependency
+		if output == input {
+			return nil
+		}
+
+		if relativeTo != "" {
+			output, err = filepath.Rel(relativeTo, output)
 			if err != nil {
-				return nil, fmt.Errorf("cannot make %q relativeDir to %q: %v", outputFilePath, relativeDir, err)
+				return fmt.Errorf("%q depends on %q, which is not relative to %q; ensure that --relative-to"+
+					" is an an ancestor of %q: %v", input, output, relativeTo, output, err)
 			}
 		}
 
-		dependentThriftFiles = append(dependentThriftFiles, outputFilePath)
-	}
+		deps = append(deps, output)
+		return nil
+	})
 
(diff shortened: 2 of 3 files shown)


tools/walkimports/README.md
```diff
@@ -0,0 +1,11 @@
+# walkimports
+
+This tool can be used on a given thrift file to output the list of thrift files that the given file depends.
```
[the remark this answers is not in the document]
jacobgreenleaf: done

tools/walkimports/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Usage
+
+```
+thriftrw$ go run tools/walkimports/main.go -relative=$(pwd) gen/testdata/thrift/structs.thrift
```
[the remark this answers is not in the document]
jacobgreenleaf: Fixed first two comments.

I wanted an example that could be actually run to show what the tool does, and found the test data in `gen/testdata/thrift/structs.thrift` in this repo.

tools/walkimports/README.md
```diff
@@ -0,0 +1,11 @@
+# walkimports
```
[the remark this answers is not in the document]
jacobgreenleaf: I ended up using `thriftrw-list-deps`

tools/walkimports/main.go
```diff
@@ -0,0 +1,46 @@
+package main
```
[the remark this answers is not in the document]
jacobgreenleaf: Thanks, makes sense.

tools/walkimports/main.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"path/filepath"
+
+	"go.uber.org/thriftrw/compile"
+)
+
+var relative *string = flag.String("relative", "", "Output paths will be relative to this directory")
```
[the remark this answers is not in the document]
jacobgreenleaf: Done.

tools/walkimports/main.go
```diff
@@ -0,0 +1,46 @@
...
+			}
+		}
+		fmt.Fprintf(output, "%s\n", outputFilePath)
+	}
+
+	return nil
+}
+
+func main() {
+	flag.Parse()
```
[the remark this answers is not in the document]
jacobgreenleaf: Done.

tools/walkimports/main.go
```diff
@@ -0,0 +1,46 @@
...
+
+	inputThriftFile := os.Args[len(os.Args)-1]
```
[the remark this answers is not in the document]
jacobgreenleaf: Nice catch. I have made the first positional argument required, so the program now errors if it is not specified:

```
$ thriftrw-list-deps
the required argument `thrift-file` was not provided
2018/06/12 12:15:17 error parsing arguments: the required argument `thrift-file` was not provided
$
```

cmd/thriftrw-list-deps/main.go
```diff
@@ -0,0 +1,58 @@
...
+	_, err := flags.Parse(&opts)
+	if err != nil {
+		log.Fatalf("error parsing arguments: %s", err.Error())
+		os.Exit(1)
+	}
+
+	inputThriftFile := opts.Args.ThriftFile
+
+	err = do(opts.Relative, inputThriftFile, os.Stdout)
+	if err != nil {
```
amckinney: nit:
```
if err := do(opts.Relative, inputThriftFile, os.Stdout); err != nil { ... }
```

cmd/thriftrw-list-deps/main.go
```diff
@@ -0,0 +1,58 @@
...
+			}
+		}
+		fmt.Fprintf(output, "%s\n", outputFilePath)
+	}
+
+	return nil
+}
+
+func main() {
+	_, err := flags.Parse(&opts)
```
amckinney: nit:
```
if _, err := flags.Parse(&opts); err != nil { ... }
```

cmd/thriftrw-list-deps/main_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+func TestRecursiveImports(t *testing.T) {
+	outputBuf := new(bytes.Buffer)
+	cwd, err := os.Getwd()
+	require.NoError(t, err)
+	idlRoot := filepath.Join(cwd, "../../gen/testdata/thrift")
+	err = do(idlRoot, filepath.Join(idlRoot, "structs.thrift"), outputBuf)
+	require.NoError(t, err)
+	outputLines := strings.Split(outputBuf.String(), "\n")
+	// t.Logf("output lines: %+v", outputLines)
```
amckinney: Lingering log

cmd/thriftrw-list-deps/main_test.go
```diff
@@ -0,0 +1,44 @@
...
+	assert.Equal(t, 2, len(outputLines))
+	assert.Equal(t, "enums.thrift", outputLines[0])
+	assert.Equal(t, "", outputLines[1])
+
+	outputBuf.Reset()
```
amckinney: nit: Rather than a reset buffer / re-do, we could use sub-tests or table-driven tests.

i.e. something similar to: 
```
t.Run("default options", ...)
t.Run("with relative path", ...)
```

With that said, could we add a test that includes an output with multiple expected thrift files?

cmd/thriftrw-list-deps/main_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	err = do(filepath.Join(cwd, "../../"), filepath.Join(idlRoot, "structs.thrift"), outputBuf)
+	require.NoError(t, err)
+	outputLines = strings.Split(outputBuf.String(), "\n")
+	// t.Logf("output lines: %+v", outputLines)
```
amckinney: Lingering log

cmd/thriftrw-list-deps/README.md
```diff
@@ -0,0 +1,17 @@
...
+## Installation
+
+```bash
+$ go get go.uber.org/thriftrw/cmd/thriftrw-list-deps
+```
+
+## Usage
+
+```bash
+$ thriftrw-list-deps -relative=$(pwd) gen/testdata/thrift/structs.thrift
```
abhinav: nitty nit: `--relative`

cmd/thriftrw-list-deps/main.go
```diff
@@ -0,0 +1,56 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/jessevdk/go-flags"
+
+	"go.uber.org/thriftrw/compile"
+)
+
+type args struct {
+	ThriftFile string `positional-arg-name:"thrift-file"`
```
abhinav: nit: I would call it just "File" (and the positional arg "FILE").

And can we add a `description:"Path to the Thrift file"`?

cmd/thriftrw-list-deps/main.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+var opts struct {
+	Relative string `long:"relative" description:"Output paths will be relative to this directory"`
```
abhinav: nit:

    If specified, output paths will be relative to this directory.

abhinav: Also, can we call this `--relative-to`? So that usage looks like,

    thriftrw-list-deps --relative-to=idl idl/whatever/foo.thirft

cmd/thriftrw-list-deps/main.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		return err
+	}
+
+	for _, included := range module.Includes {
+		outputFilePath := included.Module.ThriftPath
+		if relative != "" {
+			outputFilePath, err = filepath.Rel(relative, outputFilePath)
+			if err != nil {
+				return err
```
abhinav: The error message should include more context.

    fmt.Errorf("cannot make %q relative to %q: %v", outputFilePath, relative, err)

cmd/thriftrw-list-deps/main.go
```diff
@@ -0,0 +1,56 @@
...
+	Args     args   `positional-args:"yes" required:"yes"`
+}
+
+func do(relative string, inputThriftFile string, output io.Writer) error {
+	module, err := compile.Compile(inputThriftFile)
+	if err != nil {
+		return err
+	}
+
+	for _, included := range module.Includes {
```
abhinav: This will only do direct dependencies. Don't we want to list every Thrift file, both direct and transitive?

You can use [`module.Walk`](https://godoc.org/go.uber.org/thriftrw/compile#Module.Walk) to do the work for you.

cmd/thriftrw-list-deps/main.go
```diff
@@ -0,0 +1,56 @@
...
+		outputFilePath := included.Module.ThriftPath
+		if relative != "" {
+			outputFilePath, err = filepath.Rel(relative, outputFilePath)
+			if err != nil {
+				return err
+			}
+		}
+		fmt.Fprintf(output, "%s\n", outputFilePath)
```
abhinav: From a testing perspective, it would be preferable for the function with the business logic to return a `[]string` so that tests can assert that.

cmd/thriftrw-list-deps/main.go
```diff
@@ -0,0 +1,56 @@
...
+
+func main() {
+	if _, err := flags.Parse(&opts); err != nil {
+		log.Fatalf("error parsing arguments: %s", err.Error())
+		os.Exit(1)
+	}
+
+	inputThriftFile := opts.Args.ThriftFile
+
+	if err := do(opts.Relative, inputThriftFile, os.Stdout); err != nil {
```
abhinav: We should try to have exactly one log.Fatal in the program. The way I like to organize it is,

```go
func main() {
  log.SetFlags(0)  // so that the error message isn't noisy
  if err := run(); err != nil {
    log.Fatalf("%+v", err)  // this exits with a non-zero code
  }
}

// run is the real "main", except it returns an error instead of calling log.Fatal
func run() error {
  var opts struct {
    // flags defined here
  }
  // parse flags, return err if non-nil.

  deps, err := listDependencies(opts.RelativeTo, opts.InputFile)
  if err != nil {
    ...
  }

  for _, dep := range deps {
    fmt.Println(dep)
  }
}

func listDependencies(...) ([]string, error) {
   // business logic
}
```

Then tests can be written against `listDependencies`.

cmd/thriftrw-list-deps/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+		require.NoError(t, err)
+		outputLines := strings.Split(outputBuf.String(), "\n")
+		t.Logf("output lines: %+v", outputLines)
+		assert.Equal(t, 2, len(outputLines))
+		assert.Equal(t, "gen/testdata/thrift/enums.thrift", outputLines[0])
+		assert.Equal(t, "", outputLines[1])
+	})
+}
+
+func ExampleUsage() {
```
abhinav: The way testable examples work, this won't show up in the docs at all. Something called `ExampleUsage` is expected to appear next to the documentation for the function "Usage". Since you made os.Stdout swappable, you could just make this a normal test.

```
buffer := new(bytes.Buffer)
do(..., buffer)
assert.Equal(t, "...", buffer.String())
```

cmd/thriftrw-list-deps/README.md
```diff
@@ -0,0 +1,17 @@
+# list-deps
```
abhinav: nit: thriftrw-list-deps now

cmd/thriftrw-list-deps/README.md
```diff
@@ -0,0 +1,17 @@
...
+## Installation
+
+```bash
+$ go get go.uber.org/thriftrw/cmd/thriftrw-list-deps
+```
+
+## Usage
+
+```bash
+$ thriftrw-list-deps --relative=$(pwd) gen/testdata/thrift/structs.thrift
```
abhinav: outdated: relative-to

cmd/thriftrw-list-deps/main.go
```diff
@@ -0,0 +1,75 @@
...
+	"go.uber.org/thriftrw/compile"
+)
+
+type args struct {
+	ThriftFile string `positional-arg-name:"file" description:"Path to the Thrift file"`
+}
+
+var opts struct {
+	RelativeTo string `long:"relative-to" description:"If specified, output paths will be relative to this directory"`
+	Args       args   `positional-args:"yes" required:"yes"`
```
abhinav: optional: Can embed `args` struct here if you would prefer that.

```
var opts struct {
}
```

cmd/thriftrw-list-deps/main.go
```diff
@@ -0,0 +1,75 @@
...
+		fmt.Println(path)
+	}
+
+	return nil
+}
+
+func main() {
+	log.SetFlags(0) // so that the error message isn't noisy
+	if err := run(); err != nil {
+		log.Fatalf("%v", err)
```
abhinav: nit: Equivalent to `log.Fatal(err)`

cmd/thriftrw-list-deps/main.go
```diff
@@ -0,0 +1,75 @@
...
+
+	return dependentThriftFiles, nil
+}
+
+func run() error {
+	if _, err := flags.Parse(&opts); err != nil {
+		return fmt.Errorf("error parsing arguments: %v", err)
+	}
+
+	inputThriftFile := opts.Args.ThriftFile
```
abhinav: nit: don't need verbose names. can do `thriftFile` or even just `file`.

cmd/thriftrw-list-deps/main.go
```diff
@@ -0,0 +1,75 @@
...
+	Args       args   `positional-args:"yes" required:"yes"`
+}
+
+// listDependentThrifts lists the Thrift files that the given Thrift file depends on or includes, both directly (through imports) and
+// indirectly (through transitive imports).
+//
+// The returned file paths are absolute, unless relativeDir parameter is given, in which case the paths are relative to
+// the relativeDir.
+func listDependentThrifts(inputThriftFile string, relativeDir string) ([]string, error) {
+	dependentThriftFiles := make([]string, 0)
```
abhinav: nit: use `var` form for empty slices.

https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

cmd/thriftrw-list-deps/main.go
```diff
@@ -0,0 +1,75 @@
...
+
+	module, err := compile.Compile(inputThriftFile)
+	if err != nil {
+		return nil, fmt.Errorf("could not compile %q: %v", inputThriftFile, err)
+	}
+
+	for _, included := range module.Includes {
```
abhinav: This won't cover transitive imports. You'll need to use [`module.Walk`](https://godoc.org/go.uber.org/thriftrw/compile#Module.Walk) for that.

```go
module.Walk(func(included *compile.Module) error {
    ...
})
```

cmd/thriftrw-list-deps/main.go
abhinav: Also, nit: Again, verbose variables not needed. Can use just `deps`.

cmd/thriftrw-list-deps/main.go
```diff
@@ -0,0 +1,75 @@
...
+		outputFilePath := included.Module.ThriftPath
+
+		if relativeDir != "" {
+			outputFilePath, err = filepath.Rel(relativeDir, outputFilePath)
+			if err != nil {
+				return nil, fmt.Errorf("cannot make %q relativeDir to %q: %v", outputFilePath, relativeDir, err)
```
abhinav: The error message here can be clearer as to what's actionable for the user.

    fmt.Errorf(
        "%q depends on %q, which is not relative to %q: "+
        "ensure that --relative-to is an ancestor of %q: %v",
        inputThriftFile, outputFilePath, relativeDir, outptuFilePath, err)

cmd/thriftrw-list-deps/main.go
```diff
@@ -0,0 +1,80 @@
...
+// the relativeTo directory.
+func listDependentThrifts(input string, relativeTo string) ([]string, error) {
+	var deps []string
+
+	module, err := compile.Compile(input)
+	if err != nil {
+		return nil, fmt.Errorf("could not compile %q: %v", input, err)
+	}
+
+	module.Walk(func(mod *compile.Module) error {
```
abhinav: This method returns the error from the provided function. That should be handled.

cmd/thriftrw-list-deps/main_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"os"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestThriftrwListDeps(t *testing.T) {
+	tmpDir, err := ioutil.TempDir("", "")
+	if err != nil {
+		t.Skipf("error creating temporary directory: %v", err)
```
abhinav: This should be a `require.NoError`. That will stop execution and mark the test
as failed. Otherwise, the `go test` will succeed and the test will be skipped.

cmd/thriftrw-list-deps/main_test.go
```diff
@@ -0,0 +1,86 @@
...
+			t.Skipf("error writing %s: %v", path, err)
+		}
+	}
+
+	t.Run("no dependencies", func(t *testing.T) {
+		outputLines, err := listDependentThrifts(filepath.Join(tmpDir, "test/a.thrift"), tmpDir)
+		require.NoError(t, err)
+		sort.Strings(outputLines)
+		t.Logf("output lines: %+v", outputLines)
+		assert.Equal(t, 0, len(outputLines))
```
abhinav: optional: can use `assert.Empty(t, outputLines)`

cmd/thriftrw-list-deps/main_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	for name, content := range exampleThrifts {
+		path := filepath.Join(tmpDir, name)
+		err = os.MkdirAll(filepath.Dir(path), 0755)
+		if err != nil {
+			t.Skipf("error mkdir %s: %v", path, err)
+		}
+		err = ioutil.WriteFile(path, []byte(content), 0644)
+		if err != nil {
+			t.Skipf("error writing %s: %v", path, err)
```
abhinav: require.NoError here

cmd/thriftrw-list-deps/main_test.go
```diff
@@ -0,0 +1,86 @@
...
+		assert.Equal(t, 1, len(outputLines))
+		assert.Equal(t, "test/a.thrift", outputLines[0])
+	})
+	t.Run("one dependency with relative path", func(t *testing.T) {
+		outputLines, err := listDependentThrifts(filepath.Join(tmpDir, "test/b.thrift"), filepath.Join(tmpDir, "test"))
+		require.NoError(t, err)
+		sort.Strings(outputLines)
+		t.Logf("output lines: %+v", outputLines)
+		assert.Equal(t, 1, len(outputLines))
+		assert.Equal(t, "a.thrift", outputLines[0])
```
abhinav: Can collapse length and value check.

```
assert.Equal(t, []string{"a.thrift"}, outputLines)
```

PR #355, opened 2018-06-20
gen/enum: Implement encoding.TextMarshaler

aRatnam12:
This makes it so that the MarshalYAML method is implemented by auto-generated enum types.

Currently, enums generated by thriftrw-go implement the UnmarshalText, MarshalJSON, and UnmarshalJSON interface methods. Since they don't implement MarshalText or MarshalYAML, the enum values get marshaled into YAML strings as `int32`s

### DIFF
--- a/gen/enum.go
+++ b/gen/enum.go
@@ -107,22 +107,22 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 			}
 		}
 
-		// MarshalYAML serializes <$enumName> to yaml
+		// MarshalText encodes <$enumName> to text
 		//
 		// If the enum value is recognized, its name is returned. Otherwise,
 		// its integer value is returned
 		//
-		// This implements the yaml Marshaler interface
-		func (<$v> <$enumName>) MarshalYAML() (interface{}, error) {
+		// This implements the text Marshaler interface
+		func (<$v> <$enumName>) MarshalText() ([]byte, error) {
 			<if len .Spec.Items ->
 				switch int32(<$v>) {
 				<range .UniqueItems ->
 					case <.Value>:
-						return (interface{})("<.Name>"), nil
+						return []byte("<.Name>"), nil
 				<end ->
 				}
 			<end ->
-			return (interface{})(<$v>), nil
+			return []byte(<$strconv>.FormatInt(int64(<$v>), 10)), nil
 		}
 
 		// Ptr returns a pointer to this enum value.
--- a/gen/enum.go
+++ b/gen/enum.go
@@ -107,12 +107,12 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 			}
 		}
 
-		// MarshalText encodes <$enumName> to text
+		// MarshalText encodes <$enumName> to text.
 		//
 		// If the enum value is recognized, its name is returned. Otherwise,
-		// its integer value is returned
+		// its integer value is returned.
 		//
-		// This implements the text Marshaler interface
+		// This implements the TextMarshaler interface.
 		func (<$v> <$enumName>) MarshalText() ([]byte, error) {
 			<if len .Spec.Items ->
 				switch int32(<$v>) {
--- a/gen/enum.go
+++ b/gen/enum.go
@@ -107,6 +107,24 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 			}
 		}
 
+		// MarshalYAML serializes <$enumName> to yaml
+		//
+		// If the enum value is recognized, its name is returned. Otherwise,
+		// its integer value is returned
+		//
+		// This implements the yaml Marshaler interface
+		func (<$v> <$enumName>) MarshalYAML() (interface{}, error) {
+			<if len .Spec.Items ->
+				switch int32(<$v>) {
+				<range .UniqueItems ->
+					case <.Value>:
+						return (interface{})("<.Name>"), nil
+				<end ->
+				}
+			<end ->
+			return (interface{})(<$v>), nil
+		}
+
 		// Ptr returns a pointer to this enum value.
 		func (<$v> <$enumName>) Ptr() *<$enumName> {
 			return &<$v>
--- a/gen/enum_test.go
+++ b/gen/enum_test.go
@@ -362,6 +362,16 @@ func TestTextMarshaler(t *testing.T) {
 	}
 }
 
+func TestTextMarshalerUnknownValue(t *testing.T) {
+	v := te.EnumDefault(42)
+	b, err := v.MarshalText()
+	if assert.NoError(t, err, "unable to marshall unknown value") {
+		assert.Equal(t, `42`, string(b), "text output doesn't match")
+		err = v.UnmarshalText(b)
+		assert.Error(t, err, "unmarshaling unknown enum value should return error")
+	}
+}
+
 func TestEnumAccessors(t *testing.T) {
 	t.Run("Records", func(t *testing.T) {
 		t.Run("set", func(t *testing.T) {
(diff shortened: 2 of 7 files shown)


abhinav:
Can you clarify what the use case is here? Types generated by ThriftRW Go are
meant to be marshalled to/from Thrift payloads, not JSON or YAML. There is no
guarantee that types generated by ThriftRW are round-trippable in JSON or YAML
between different implementations of ThriftRW or, as is the case here, in the
same implementation, because that is exactly the problem the Thrift encoding
is trying to solve. Passing ThriftRW shapes as JSON or YAML is a completely
unsupported path.

abhinav:
That said, if we really really need this, I would rather have MarshalText than
MarshalYAML. It'll Just Work work with the YAML package and it's not tied
inherently to YAML. It'll probably work with the TOML package too?

lukepur:
Hey @abhinav, thanks for the fast reply.

Our use case here is a bit unorthodox. We are authoring a service where the primary domain object contains a field which is the YAML representation of a configuration structure. We, rightly or wrongly, have chosen to leverage a thrift file to express the schema of this configuration object. This a) keeps our schema representation in line with standards and b) allows us to leverage the power of ThriftRW to generate the golang types automatically as we develop the schema.

One of the requirements of this configuration property however is that it is round-trip serializable to YAML. This is where the types and type methods generated by ThriftRW fall slightly short of our requirements. In particular, the marshaling of enums to YAML is using the `int32` values instead of the string representation of the symbols, and subsequently trying to unmarshal a previously marshaled struct is failing. Hence this PR to allow round-trip YAML marshaling of enums.

We completely understand that this use case may lie outside the support path of thriftrw-go, so if the functionality of this pull request is not useful (or worse, potentially problematic) then we will pursue another course. Another option we have considered is to simply use thriftrw to generate the types and then move them to a source-controlled directory and make our amendments there. This is slightly less "automatic" for us, but the infrequency of changes to the schema thrift file should render this an acceptable solution.

aRatnam12:
> I would rather have MarshalText than MarshalYAML
I'll let you respond to Luke's message first but if we do in fact go through with this, happy to switch it to MarshalText instead. We actually started off with that but switched it to the yaml marshaler just in case MarshalText could have other unintended consequences for current users of thriftrw.

abhinav:
@lukepur Thanks for the explanation. Makes sense. @aRatnam12 Having a
MarhsalText method sounds reasonable to me because the types already have an
UnmarshalText method. This keeps YAML out of the equation and so ThriftRW
still isn't making any commitments about YAML compatibility.

Thanks for the PR!

aRatnam12:
Sounds good to me! Removed anything yaml related and switched to MarshalText

gen/enum.go
```diff
@@ -107,6 +107,24 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 			}
 		}
 
+		// MarshalText encodes <$enumName> to text
```
abhinav: nit: End with `.` or this will be treated as a header.

gen/enum.go
```diff
@@ -107,6 +107,24 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+		//
+		// If the enum value is recognized, its name is returned. Otherwise,
+		// its integer value is returned
```
abhinav: nit: End with `.`

gen/enum.go
```diff
@@ -107,6 +107,24 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+		//
+		// This implements the text Marshaler interface
```
abhinav: nit: `s/text Marshaler/TextMarshaler/` and end with `.`

gen/enum_test.go
```diff
@@ -335,6 +335,28 @@ func TestUnmarshalTextReturnsErrorOnInvalidValue(t *testing.T) {
 	assert.Error(t, err)
 }
 
+func TestMarshalTextReturnsValue(t *testing.T) {
```
abhinav: nit: Just TestMarshalText?

gen/enum_test.go
```diff
@@ -335,6 +335,28 @@ func TestUnmarshalTextReturnsErrorOnInvalidValue(t *testing.T) {
...
+	tests := []struct {
+		title string
+		e     te.EnumDefault
+		text  string
```
abhinav: nit: for field naming for table tests, we've preferred to name values `give`
or `giveFoo`, `giveBar` for the input and `want` or `wantFoo`, `wantBar` for
the output. It clearly separates what the input of the test case is from the
expected output, especially when there are multiple.

gen/enum_test.go
```diff
@@ -335,6 +335,28 @@ func TestUnmarshalTextReturnsErrorOnInvalidValue(t *testing.T) {
...
+		{`2 <-> Baz`, te.EnumDefaultBaz, "Baz"},
+		{`42 <-> 42`, 42, "42"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.title, func(t *testing.T) {
+			b, err := tt.e.MarshalText()
+			if assert.NoError(t, err, "unable to marshal enum") {
+				assert.Equal(t, []byte(tt.text), b, "text output doesn't match")
+			}
```
abhinav: I think it would be valuable to verify that values are round-tripped correctly
here: call UnmarshalText on the result to verify that you get back what the
input was. (In which case this whole test can be renamed to
`TestTextMarshaler`.)

gen/enum_test.go
```diff
@@ -335,6 +335,28 @@ func TestUnmarshalTextReturnsErrorOnInvalidValue(t *testing.T) {
...
+		{`0 <-> Foo`, te.EnumDefaultFoo, "Foo"},
+		{`1 <-> Bar`, te.EnumDefaultBar, "Bar"},
+		{`2 <-> Baz`, te.EnumDefaultBaz, "Baz"},
+		{`42 <-> 42`, 42, "42"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.title, func(t *testing.T) {
+			b, err := tt.e.MarshalText()
+			if assert.NoError(t, err, "unable to marshal enum") {
```
abhinav: optional: can `require.NoError` and skip the if statement, especially if you add the UnmarshalText verification.

gen/enum_test.go
aRatnam12: Ah that's a nice convention!

gen/enum_test.go
aRatnam12: Thanks for the heads up. I'll leave it as is since the rest of the file seems to be using if statements as well

gen/enum_test.go
aRatnam12: Sounds good, Unmarshaling doesn't seem to support unknown values so I removed the 42 test.

aRatnam12:
Thanks for the review, I updated based on the comments. Also, any thoughts on my previous question about release timeline?

gen/enum_test.go
abhinav: I would add a separate test case for just unknown value marshaling while keeping this for known values.

abhinav:
Apologies, I didn't see the question. I'm happy to cut the release soon after
this lands. We avoid releasing on Fridays or after 4pm on any other weekday to
avoid outages during off hours, so we can put this out any Mon-Thu morning
after landing.

abhinav:
This LGTM besides needing an explicit test for MarshalText with unknown values (outside the table test).

aRatnam12:
Just added the test, and that sounds great thanks!

aRatnam12:
Thanks for reviewing this abhinav! It seems like I'm not able to merge this in myself, do you have some process in place for that?

abhinav:
Merged. I'll release this Monday morning.

aRatnam12:
Great, thanks a ton!

lukepur:
Yes, thanks @abhinav for the support and super-fast turnaround on this. It has helped us out a lot! 🥇

abhinav:
Released in v1.12

Issue #362, opened 2018-08-03
Support enum item renaming


cl1337:
### 1: What feature wanted:

for thrift files with reserved keyword: `function`, `end`, `native` .. etc
```thrift
enum TestEnum {
    function
    end
    native
}
```
we want to be able to separate naming from its string value in `UnmarshalText` with some labels:
```thrift
enum TestEnum {
    function1 (go.wire.label = "function")
    end1 (go.wire.label = "end")
    native (go.wire.label = "native")
}
```
and generate `type.go` with following `UnmarshalText` function 
```go
func (v *TestEnum) UnmarshalText(value []byte) error {
	switch string(value) {
	case "function":
		*v = TestEnumTypeFunction1
		return nil
	case "end":
		*v = TestEnumTypeEnd1
		return nil
        case "native":
		*v = TestEnumTypeNative1
		return nil 
	default:
		return fmt.Errorf("unknown enum value %q for %q", value, "TestEnum")
	}
}
```

### 2: why we need this

we're migrating one of our major nodejs service to go, because thriftrw node and thriftrw go have different opinion about what's a valid thrift file (e.g. keyword mentioned above), a lot of thrift files that are currently used on production cannot be migrate to go

we're not allowed to change what name was specified in thrift files, which will require a non-trivial and never end migration, we need to tweak thrift compile to make sure we can still use same name

all reserved keywords [here](https://github.com/thriftrw/thriftrw-go/blob/16ef2f392e3ca08ea35e44f95e9b5831ea2356dd/idl/internal/lex.rl#L106) 

### 3: proposed fix

Similar to https://github.com/thriftrw/thriftrw-go/issues/293, that separates concern of schema naming and actual data on the wire, we could instead of using `.Name` [here](https://github.com/thriftrw/thriftrw-go/blob/9ce01c65792c9e633eb9086985fde8f28470214f/gen/enum.go#L101), infer name from labels like `go.label` annotation

Happy to put PR out if this issue makes sense to you, also open to other good ideal to address this, thanks in advance

abhinav:
This sounds reasonable. The same annotation can be supported on struct fields to override their name in the struct's JSON representation and the Zap marshalers (#361).

kriskowal:
Reasonable.

cl1337:
https://github.com/thriftrw/thriftrw-go/pull/363

PR #363, opened 2018-08-03
gen/enum: Add `go.label` to override item name

cl1337:
This PR addresses https://github.com/thriftrw/thriftrw-go/issues/362


cl1337:
build failed due to 
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

should be fixed by restart, which I don't have permission to do so

gen/enum.go
```diff
@@ -26,6 +26,11 @@ import (
 	"go.uber.org/thriftrw/compile"
 )
 
+const (
+	// GoWireLabel overrides name on the wire
+	GoWireLabel = "go.wire.label"
```
kriskowal: Let’s call this just `"label"` or `"go.label"` since it affects the JSON and log representation of the in memory struct and not the TBinary (wire) protocol.

gen/internal/tests/thrift/enums.thrift
```diff
@@ -52,5 +52,12 @@ enum lowerCaseEnum {
     containing, lower_case, items
 }
 
+// EnumWithWireLabel use label name in serialization/deserialization
+enum EnumWithWireLabel {
+    username (go.wire.label = "surname"),
+    password (go.wire.label = "hashed_password"),
+    salt (go.wire.label = "")
```
kriskowal: Can we also test an annotation without a value, for whether the behavior is equivalent to empty string? `salt (label)` is valid Thrift IDL.

gen/enum_test.go
```diff
@@ -399,3 +399,27 @@ func TestEnumAccessors(t *testing.T) {
...
+			b, err := label.MarshalText()
+			assert.NoError(t, err)
+			assert.Equal(t, tt.wireValue, string(b))
+			assertRoundTrip(
+				t, &tt.enumItem,
+				wire.NewValueI32(int32(tt.enumItem)),
+				"test roundtrip "+tt.wireValue,
+			)
+		})
+	}
```
kriskowal: Can we also confirm through an integration test that this effects the desired behavior for JSON marshaling?

Can we also test the behavior of serializing an unrecognized enum case?

gen/internal/tests/thrift/enums.thrift
cl1337: good catch

gen/enum_test.go
cl1337: 👍

cl1337: just want to make sure I understand your comment correctly, are you suggesting something here like:
```go
...
    // marshal
    label := te.EnumWithWireLabel(te.EnumWithWireLabelUsername)
    b, err := json.Marshal(label)
    assert.Equal(t, "username", string(b))

    //  unmarlshal      
    var testLabel te.EnumWithWireLabel
    err = json.Unmarshal(b, testLabel)
    assert.Equal(t, testLabel, label)
... 
```

or you're suggesting a specific kind of integration tests? Thanks

kriskowal: This is fine.

gen/enum_test.go
```diff
@@ -399,3 +399,69 @@ func TestEnumAccessors(t *testing.T) {
...
+	for _, tt := range []struct {
+		value te.EnumWithLabel
+		rep   string
+	}{
+		{-1, "-1"},
+		{1 << 10, "1024"},
+	} {
+		invalidEnumItem := te.EnumWithLabel(tt.value)
+		b, err := json.Marshal(invalidEnumItem)
+		assert.NoError(t, err)
```
cl1337: this is default marshal behavior: if not matching any key, just marshal as string representation of int value, without raising error

gen/enum_test.go
```diff
@@ -399,3 +399,69 @@ func TestEnumAccessors(t *testing.T) {
...
+		wireValue string
+	}{
+		{te.EnumWithLabelUsername, "\"surname\""},
+		{te.EnumWithLabelPassword, "\"hashed_password\""},
+		{te.EnumWithLabelSalt, "\"salt\""},
+		{te.EnumWithLabelSugar, "\"sugar\""},
+	}
+	for _, tt := range tests {
+		t.Run("compare label:"+tt.wireValue, func(t *testing.T) {
+			// marshal
```
cl1337: added marshal / unmarshal behavior

gen/enum_test.go
```diff
@@ -399,3 +399,69 @@ func TestEnumAccessors(t *testing.T) {
...
+			assertRoundTrip(
+				t, &tt.enumItem,
+				wire.NewValueI32(int32(tt.enumItem)),
+				"test roundtrip "+tt.wireValue,
+			)
+		})
+	}
+}
+
+func TestEnumLabelInvalid(t *testing.T) {
```
cl1337: added test cases for invalid enum item marshal / unmarshal

cl1337:
thanks @kriskowal !

gen/enum.go
```diff
@@ -265,6 +275,34 @@ func enumItemName(enumName string, spec *compile.EnumItem) (string, error) {
 	return enumName + name, err
 }
 
+// enumItemLabelName returns the actual name used for serialization/deserialization
+// default to EnumItem.Name, override by the value of GoLabel
```
abhinav: nit:

    // enumItemLabelName returns the label we use for this enum item in the generated code.

(I'm trying to make it clear that the label actually has nothing to do with
serialization for the use cases that ThriftRW supports. Using ThriftRW objects
as JSON is technically unsupported; it just happens to work.)

gen/enum.go
```diff
@@ -265,6 +275,34 @@ func enumItemName(enumName string, spec *compile.EnumItem) (string, error) {
...
+func enumItemLabelName(spec *compile.EnumItem) string {
+	labelName := spec.Name
+	val, ok := spec.Annotations[GoLabel]
+	if ok && len(val) > 0 {
+		labelName = val
```
abhinav: nit: If you use the single assignment form for a map, it returns the zero
value for that type (which is an empty string in this case).

So this can be simplified to,

```go
if val := spec.Annotations[GoLabel]; len(val) > 0 {
  labelName = val
}
```

gen/enum.go
```diff
@@ -265,6 +275,34 @@ func enumItemName(enumName string, spec *compile.EnumItem) (string, error) {
...
+	}
+	return labelName
+}
+
+// validateEnumUniqueNames apply name label GoLabel and raise error if there's
+// duplicates in resolved enum item names
+func validateEnumUniqueNames(spec *compile.EnumSpec) error {
+	items := spec.Items
+	used := make(map[string]bool, len(items))
```
abhinav: When using a `map` as a set, we prefer to use `struct{}` instead of bool so
that you don't accidentally rely on the value of the boolean. Plus `struct{}`
has zero runtime cost.

```
used := make(map[string]struct{}, len(items))
...
used[itemName] = struct{}{}
```

gen/enum.go
```diff
@@ -265,6 +275,34 @@ func enumItemName(enumName string, spec *compile.EnumItem) (string, error) {
...
+	for _, i := range items {
+		itemName := enumItemLabelName(&i)
+		if _, isUsed := used[itemName]; isUsed {
+			return fmt.Errorf(
+				"duplicated item name %q found in enum %q",
```
abhinav: nit: If we find a conflict, it would be nice to know what it's with. The above
map could be changed to a `map[string]string`, mapping label to original item.

    map[string]EnumItem

Then this can be,

    if conflict, isUsed := used[itemName]; isUsed {
        return fmt.Errorf(
            "item %q with label %q conflicts with item %q in enum %q",
            i.Name, itemName, conflict.Name, spec.Name)
    }

gen/enum_test.go
```diff
@@ -399,3 +400,107 @@ func TestEnumAccessors(t *testing.T) {
...
 	})
 }
+
+func TestEnumLabelLegit(t *testing.T) {
+	tests := []struct {
+		enumItem  te.EnumWithLabel
+		wireValue string
+	}{
+		{te.EnumWithLabelUsername, "\"surname\""},
+		{te.EnumWithLabelPassword, "\"hashed_password\""},
```
abhinav: nit: can use backticks to avoid having to escape "

    `"surname"`
    `"hashed_password"`

gen/enum_test.go
```diff
@@ -399,3 +400,107 @@ func TestEnumAccessors(t *testing.T) {
 		})
 	})
 }
+
+func TestEnumLabelLegit(t *testing.T) {
+	tests := []struct {
+		enumItem  te.EnumWithLabel
+		wireValue string
```
abhinav: nit: jsonValue not wireValue

gen/enum_test.go
```diff
@@ -399,3 +400,107 @@ func TestEnumAccessors(t *testing.T) {
 		})
 	})
 }
+
+func TestEnumLabelLegit(t *testing.T) {
```
abhinav: nit: s/Legit/Valid

gen/enum_test.go
```diff
@@ -399,3 +400,107 @@ func TestEnumAccessors(t *testing.T) {
...
+	tests := []struct {
+		enumItem  te.EnumWithLabel
+		wireValue string
+	}{
+		{te.EnumWithLabelUsername, "\"surname\""},
+		{te.EnumWithLabelPassword, "\"hashed_password\""},
+		{te.EnumWithLabelSalt, "\"salt\""},
+		{te.EnumWithLabelSugar, "\"sugar\""},
```
abhinav: We should add a case for unknown enums here. (They're still valid.)

    {te.EnumWithLabel(42), "42"}

gen/enum_test.go
```diff
@@ -399,3 +400,107 @@ func TestEnumAccessors(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		t.Run("compare label:"+tt.wireValue, func(t *testing.T) {
```
abhinav: I would just do `t.Run(tt.enumItem.String()`

gen/enum_test.go
```diff
@@ -399,3 +400,107 @@ func TestEnumAccessors(t *testing.T) {
...
+			// marshal
+			label := te.EnumWithLabel(tt.enumItem)
+			b, err := json.Marshal(label)
+			assert.NoError(t, err)
```
abhinav: We should do require.NoError here so that the test doesn't proceed if it failed.

gen/enum_test.go
```diff
@@ -399,3 +400,107 @@ func TestEnumAccessors(t *testing.T) {
...
+			assert.Equal(t, tt.wireValue, string(b))
+
+			// unmarshal
+			var expectedLabel te.EnumWithLabel
+			err = json.Unmarshal(b, &expectedLabel)
```
abhinav: require.NoError

gen/enum_test.go
```diff
@@ -399,3 +400,107 @@ func TestEnumAccessors(t *testing.T) {
...
+			assert.Equal(t, expectedLabel, label)
```
abhinav: With assert.Equal, the call pattern is `assert.Equal(t, expected, got)`. Here, `label` is what we expect and `expectedLabel` is what we got (from deserializing). So this should be,

    assert.Equal(t, label, expectedLabel)

(We should fix naming.)

gen/enum_test.go
abhinav: Can we put the JSON behavior check in a subtest since this is also testing roundtripping.

```
t.Run("JSON round trip", func(t *testing.T) {
  ...
})
assertRoundTrip(...)
```

gen/enum_test.go
```diff
@@ -399,3 +400,107 @@ func TestEnumAccessors(t *testing.T) {
...
+		})
+	}
+}
+
+func TestEnumLabelInvalid(t *testing.T) {
+
+	for _, tt := range []struct {
+		value te.EnumWithLabel
+		rep   string
+	}{
```
abhinav: nit: for table tests, we prefer the following form.

```
tests := []struct{
  ...
}{
  ...
}

for _, tt := range tests {
  ..
}
```

Inlining the test cases with the range statement hurts readability IMO.

gen/enum_test.go
```diff
@@ -399,3 +400,107 @@ func TestEnumAccessors(t *testing.T) {
...
+		{-1, "-1"},
+		{1 << 10, "1024"},
```
abhinav: This isn't technically invalid. These are valid enum items. They're just unknown. You can fold these cases into the valid case above.

gen/enum_test.go
```diff
@@ -399,3 +400,107 @@ func TestEnumAccessors(t *testing.T) {
...
+		errVal []byte
+		errMsg string
+	}{
+		{
+			[]byte("some-random-str"),
+			"invalid character 's' looking for beginning of value",
+		},
+		{
+			[]byte("\"; drop table users;\""),
+			"unknown enum value \"; drop table users;\" for \"EnumWithLabel\"",
```
abhinav: nit: use backticks to avoid quoting "

gen/enum_test.go
```diff
@@ -399,3 +400,107 @@ func TestEnumAccessors(t *testing.T) {
...
+					compile.EnumItem{
+						Name:  "B",
+						Value: 1,
+						Annotations: map[string]string{
+							"go.label": "A",
+						},
+					},
+				},
+			},
+			"duplicated item name \"A\" found in enum \"duplicate item name\"",
```
abhinav: nit: backticks

gen/enum_test.go
```diff
@@ -399,3 +400,107 @@ func TestEnumAccessors(t *testing.T) {
...
+	} {
+		t.Run("invalid unmarshal: "+tt.errMsg, func(t *testing.T) {
+			var expectedLabel te.EnumWithLabel
+			err := json.Unmarshal(tt.errVal, &expectedLabel)
+			assert.Equal(t, err.Error(), tt.errMsg)
+		})
+	}
+}
+
+func TestGenInvalidEnumFailure(t *testing.T) {
```
abhinav: nit: TestEnumLabelConflict

gen/internal/tests/thrift/enums.thrift
```diff
@@ -52,5 +52,13 @@ enum lowerCaseEnum {
     containing, lower_case, items
 }
 
+// EnumWithLabel use label name in serialization/deserialization
+enum EnumWithLabel {
+    username (go.label = "surname"),
+    password (go.label = "hashed_password"),
+    salt (go.label = ""),
+    sugar (go.label)
```
abhinav: Can the items be upper-case (the labels can be lower-case)? That matches our
Thrift Style Guide.

    USERNAME (go.label = "surname")

And we should have one case where the label is a Thrift keyword since that's
the motivation behind this change.

gen/enum.go
cl1337: agreed, the serialization/deserialization is pretty misleading given ThriftRW's context here

gen/enum.go
cl1337: thanks! Pretty neat trick, and good to know about the zero runtime cost part :)
```go
func BenchmarkBool(b *testing.B) {
	mp := make(map[string]bool)
	for i := 0; i < b.N; i++ {
		if _, ok := mp[string(i)]; ok {
			mp[string(i)] = true
		}
	}
}

func BenchmarkStruct(b *testing.B) {
	mp := make(map[string]struct{})
	for i := 0; i < b.N; i++ {
		if _, ok := mp[string(i)]; ok {
			mp[string(i)] = struct{}{}
		}
	}
}
```
```
▶ go test -bench=Benchmark -count=1 -benchmem bench_test.go
goos: darwin
goarch: amd64
BenchmarkBool-8     	200000000	         7.00 ns/op	       0 B/op	       0 allocs/op
BenchmarkStruct-8   	200000000	         7.31 ns/op	       0 B/op	       0 allocs/op
PASS
```

gen/enum.go
cl1337: makes sense

gen/internal/tests/thrift/enums.thrift
cl1337: good advice, enum name and its label is actually case sensitive, I added upper/lower cases for item name and label, also added one reserved keyword label

CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- No changes yet.
+### Added
+- gen: Added `go.label` used in enum item to override enum
+  item name in codegen.
```
prashantv: i think we should be more explicit about what exactly `go.label` affects here. as someone without much context on this change, it's unclear what "item name in codegen" actually means. does it mean the name of the Go identifier? Does it mean the name in `String`? What about JSON/YAML?

gen/enum.go
```diff
@@ -265,6 +273,32 @@ func enumItemName(enumName string, spec *compile.EnumItem) (string, error) {
 	return enumName + name, err
 }
 
+// enumItemLabelName returns the label we use for this enum item in the generated code.
+func enumItemLabelName(spec *compile.EnumItem) string {
+	labelName := spec.Name
```
prashantv: optional: can be simplified:
```go
if val := spec.Annotations[GoLabel]; len(val) > 0 {
  return val
}
return spec.Name
```

gen/enum_test.go
```diff
@@ -25,11 +25,12 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
```
prashantv: think imports were grouped in this repo before -- can we keep the third party imports at the bottom?

gen/enum.go
```diff
@@ -21,11 +21,15 @@
 package gen
 
 import (
+	"fmt"
 	"strings"
 
 	"go.uber.org/thriftrw/compile"
 )
 
+// GoLabel is the annotation expected on enum items to override their label.
```
prashantv: similar to changelog, can we expand documentation here on what exactly this means and how this affects users relying on the JSON output from thriftrw-go

gen/enum_test.go
```diff
@@ -399,3 +400,97 @@ func TestEnumAccessors(t *testing.T) {
...
+					compile.EnumItem{
+						Name:  "B",
+						Value: 1,
+						Annotations: map[string]string{
+							"go.label": "A",
+						},
+					},
+				},
+			},
+			`item "B" with label "A" conflicts with item "A" in enum "duplicate item name"`,
```
prashantv: suggest an extra test where 2 different enums specify the same `go.label`

gen/enum_test.go
```diff
@@ -399,3 +400,97 @@ func TestEnumAccessors(t *testing.T) {
...
+	}
+}
+
+func TestEnumLabelInvalidUnmarshal(t *testing.T) {
+	tests := []struct {
+		errVal []byte
+		errMsg string
+	}{
+		{
+			[]byte("some-random-str"),
```
prashantv: this test isn't testing the `UnmarshalText` at all -- this is testing the `json.Unmarshal`. Not sure that's useful?

prashantv: we should have a test for the original enum name, and ensure that it's no longer recognized. e.g., `"USERNAME"`

CHANGELOG.md
cl1337: good point, what do you think of:
```
- gen: Added `go.label="<TAGGED_NAME>"` annotation for enum items. Corresponding items
  of the generated Go structs will be using <TAGGED_NAME> for JSON marshal/unmarshal
  instead of origin item name, this change does not apply to YAML
```

gen/enum_test.go
cl1337: https://github.com/thriftrw/thriftrw-go/blob/dev/gen/internal/tests/enums/types.go#L98
`json.Unmarshal` will call `UnmarshalText` in this test case.

I was testing against `UnmarshalText`, and was suggested to test `json.Unmarshal` in previous revision, let me know if we want to also add `UnmarshalText` here

gen/enum_test.go
cl1337: ops, ide linter didnt' capture this

gen/enum.go
cl1337: how about:
```
// GoLabel is the annotation expected on enum items to override their label.
// Enum items tagged with `go.label=<TAGGED_NAME>` will generate go struct
// using `TAGGED_NAME` for JSON Marshal/Unmarshal
```

CHANGELOG.md
prashantv: I think this change applies to all text outputs -- `String()`, JSON, YAML and anything else that uses `TextMarshaler`

gen/enum_test.go
prashantv: The issue is that this code never hits `UnmarshalText`.
https://play.golang.org/p/B1SIZYDGr_4

Using `json.Unmarshal` is fine, but the test case is incorrect since it probably should be `"some-random-str"` (e.g., the quotes should be part of the value passed to `Unmarshal`)

CHANGELOG.md
cl1337: I'll revise it to
```
- gen: Added `go.label="<TAGGED_NAME>"` annotation for enum items. 
  Corresponding items of the generated Go structs will be using 
  <TAGGED_NAME> for text marshal / unmarshal instead of its origin item
  name
```

gen/enum_test.go
cl1337: ah, good catch! Thanks for this comment, I'll fix the test cases here

cl1337: actually looked at it again I added that first test case intentionally, you are right it didn't hit `UnmarshalText` at all like the rest of test cases, I added it as an invalid case for json Unmarshal, but has nothing to do with this label PR (generic invalid case), I'll make it `"some-randome-str"` so it hits `unmarshalText`

gen/enum.go
prashantv: There's no go struct involved for an enum, so I think it's a little confusing. I think it would be better as:
```
// GoLabel allows overriding the text formatting of an enum.
// Enum items will use the annotation value when serialized as
// a string. This affects String(), as well as text marshalling, used by JSON/YAML.
```

gen/enum_test.go
```diff
@@ -399,3 +401,128 @@ func TestEnumAccessors(t *testing.T) {
...
+				wire.NewValueI32(int32(tt.enumItem)),
+				"test roundtrip "+tt.jsonValue,
+			)
+		})
+	}
+}
+
+func TestEnumLabelInvalidUnmarshal(t *testing.T) {
+	tests := []struct {
+		errVal []byte
```
prashantv: nit: since you are using strings for all the cases, make this `string`, and then convert to `[]byte` when calling `Unmarshal`

removes some noise from tests

gen/enum_test.go
```diff
@@ -399,3 +401,128 @@ func TestEnumAccessors(t *testing.T) {
...
+		{
+			[]byte(`"PASSWORD"`),
+			`unknown enum value "PASSWORD" for "EnumWithLabel"`,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(string(tt.errVal), func(t *testing.T) {
+			var expectedLabel te.EnumWithLabel
+			err := json.Unmarshal(tt.errVal, &expectedLabel)
+			assert.Equal(t, err.Error(), tt.errMsg)
```
prashantv: use `assert.EqualError`, otherwise, if it doesn't return an error, this will panic

gen/enum_test.go
```diff
@@ -399,3 +401,128 @@ func TestEnumAccessors(t *testing.T) {
...
+					},
+				},
+			},
+			`item "B" with label "C" conflicts with item "A" in enum "duplicate item name"`,
+		},
+	}
+	for _, tt := range testCases {
+		t.Run(tt.spec.Name, func(t *testing.T) {
+			err := enum(nil, &tt.spec)
+			assert.Error(t, err)
```
prashantv: use `assert.EqualError` instead of `assert.Error` + `assert.Equal`, which could panic if `err` is `nil`

gen/enum_test.go
cl1337: neat! Thanks for the advice

PR #366, opened 2018-08-06
Support fast Zap logging

mh-park:
(This supersedes #361.)

This adds a `zapcore.MarshalLogObject/Array` method to all generated structs and its underlying components, enums, and non-primitive typedefs, implementing Zap marshalers (i.e. [ObjectMarshaler](https://godoc.org/go.uber.org/zap/zapcore#ObjectMarshaler), [ArrayMarshaler](https://godoc.org/go.uber.org/zap/zapcore#ArrayMarshaler)). For typedefs of primitives, the logging is up to the user, simply by casting it down to the root type and using the respective `Add/Append...` method of the Zap encoder.


CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- No changes yet.
+- Added `zapcore.MarshalLogObject/Array` method to all generated structs 
+  and its underlying components, enums, and non-primitive typedefs, implementing 
+  Zap marshalers.
```
abhinav: Let's change this to,

```
### Added
- Generated types now implement zapcore.ObjectMarshaler or
  zapcore.ArrayMarshaler where appropriate. This should lead to much faster
  logging of these objects.
```

gen/field.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Uber Technologies, Inc.
+// Copyright (c) 2018 Uber Technologies, Inc
```
abhinav: ?

gen/field.go
```diff
@@ -461,6 +465,46 @@ func (f fieldGroupGenerator) Equals(g Generator) error {
...
+						<if (zapCanError .Type) ->
+							if err := <$encAdd>; err != nil {
+								return err
+							}
+						<- else ->
+							<$encAdd>
+						<- end ->
+					}
+				<- end>
+			<end>
```
abhinav: nit: use `<end ->` for to end the range so that if .Fields is empty, we don't get an empty line.

gen/field.go
```diff
@@ -461,6 +465,46 @@ func (f fieldGroupGenerator) Equals(g Generator) error {
...
+					<- $encAdd := printf "%s.Add%s(%q, %s)" $enc (zapEncoder .Type) .Name (zapMarshaler .Type $fval) ->
+					<if (zapCanError .Type) ->
+						if err := <$encAdd>; err != nil {
+							return err
+						}
+					<- else ->
+						<$encAdd>
+					<- end ->
+				<- else ->
+					<- $encAdd := printf "%s.Add%s(%q, %s)" $enc (zapEncoder .Type) .Name (zapMarshalerPtr .Type $fval) ->
```
abhinav: I think we can reduce some of the duplication here with a TemplateFunc:

```
TemplateFunc("fieldMarshaler", func(required bool, t compile.TypeSpec, fval string) (string, error) {
    if required {
        return zapMarshaler(g, t, fval)
    }
    return zapMarshalerPtr(g, t, fval)
})
```

Then the template becomes something like,

```
<- $encAdd := printf "%s.Add%s(%q, %s)" ... (fieldMarshaler .Required .Type $fval) ->
<- if .Required ->
    <if (zapCanError .Type) ->
        ...
    <- else ->
        <$encAdd>
    <- end _>
<- else ->
    ...
<end>
```

gen/field.go
```diff
@@ -461,6 +465,46 @@ func (f fieldGroupGenerator) Equals(g Generator) error {
...
+		<$zapcore := import "go.uber.org/zap/zapcore">
+		<$v := newVar "v">
+		<$enc := newVar "enc">
+
+		// MarshalLogObject implements zapcore.ObjectMarshaler, enabling
+		// fast logging of <.Name>.
+		func (<$v> *<.Name>) MarshalLogObject(<$enc> <$zapcore>.ObjectEncoder) error {
+			<range .Fields>
+				<- $fname := goName . ->
+				<- $fval := printf "%s.%s" $v $fname ->
```
abhinav: nit: we can inline `$fname` in the `$fval` declaration

gen/list.go
```diff
@@ -177,3 +181,45 @@ func (l *listGenerator) Equals(g Generator, spec *compile.ListSpec) (string, err
...
+	g Generator,
+	spec compile.TypeSpec,
+	root *compile.ListSpec,
+	fieldValue string,
+) (string, error) {
+	name := zapperName(g, spec)
+	if err := g.EnsureDeclared(
+		`
+			type <.Name> <typeReference .Type>
+			<$zapcore := import "go.uber.org/zap/zapcore">
```
abhinav: nit: put the import at the top of this template (almost like a real import)

gen/zap.go
```diff
@@ -0,0 +1,132 @@
...
+		return "Array"
+
+	// User-defined
+	case *compile.EnumSpec, *compile.StructSpec:
+		return "Object"
+	}
+	panic(root)
+}
+
+func (z *zapGenerator) zapMarshaler(g Generator, spec compile.TypeSpec, fieldValue string) (string, error) {
```
abhinav: Let's add some documentation to these methods explaining what they're returning/where they're meant to be used.

Like zapEncoder returns the name of the type in Zap as which the object will be logged (array, object, etc.).
zapMarshaler accepts [..] and returns an expression that [..].

And we should mention how they all fit together in the zapGenerator docs.

gen/zap.go
```diff
@@ -0,0 +1,132 @@
...
+	if isPrimitiveType(spec) {
+		return fieldValue, nil
+	}
+
+	switch t := root.(type) {
+	case *compile.BinarySpec:
+		// There is no AppendBinary for ArrayEncoder, so we opt for encoding it ourselves and
+		// appending it as a string. We also use AddString instead of AddBinary for ObjectEncoder
+		// for consistency.
+		base64 := g.Import("encoding/base64")
```
abhinav: nice

gen/list.go
```diff
@@ -177,3 +181,45 @@ func (l *listGenerator) Equals(g Generator, spec *compile.ListSpec) (string, err
...
 	return name, wrapGenerateError(spec.ThriftName(), err)
 }
+
+func (l *listGenerator) zapMarshaler(
+	g Generator,
+	spec compile.TypeSpec,
+	root *compile.ListSpec,
+	fieldValue string,
+) (string, error) {
+	name := zapperName(g, spec)
```
abhinav: Are you sure you need to bother with the original `spec`? The wrapper is
generated with the root type regardless. The typedef-of-list can be cast to
the wrapper of the list directly.

gen/map.go
```diff
@@ -281,3 +281,144 @@ func (m *mapGenerator) equalsUnhashable(g Generator, spec *compile.MapSpec) (str
...
+	spec compile.TypeSpec,
+	root *compile.MapSpec,
+	fieldValue string,
+) (string, error) {
+	name := zapperName(g, spec)
+	switch root.KeySpec.(type) {
+	case *compile.StringSpec:
+		// We are already at the root, so no need to check if it is a typedef of a
+		// string. For simplicity, we always cast the key to a string when logging this way.
+		if err := g.EnsureDeclared(
```
abhinav: You're dealing with the root `map[...]` type but the key could still be a typedef of a string so you still want inspect `RootTypeSpec(root.KeySpec)`.

```
typedef string Foo
typedef map<Foo, i32> Bar
```

gen/map.go
```diff
@@ -281,3 +281,144 @@ func (m *mapGenerator) equalsUnhashable(g Generator, spec *compile.MapSpec) (str
...
+
+func (m *mapGenerator) zapMarshaler(
+	g Generator,
+	spec compile.TypeSpec,
+	root *compile.MapSpec,
+	fieldValue string,
+) (string, error) {
+	name := zapperName(g, spec)
+	switch root.KeySpec.(type) {
+	case *compile.StringSpec:
```
abhinav: can we split these two cases out into two functions called from here? would help with readabilty a bit.

gen/map.go
```diff
@@ -281,3 +281,144 @@ func (m *mapGenerator) equalsUnhashable(g Generator, spec *compile.MapSpec) (str
...
+		// We are already at the root, so no need to check if it is a typedef of a
+		// string. For simplicity, we always cast the key to a string when logging this way.
+		if err := g.EnsureDeclared(
+			`
+				type <.Name> <typeReference .Type>
+				<$zapcore := import "go.uber.org/zap/zapcore">
```
abhinav: Imports at the top please

gen/map.go
```diff
@@ -281,3 +281,144 @@ func (m *mapGenerator) equalsUnhashable(g Generator, spec *compile.MapSpec) (str
...
+				<$v := newVar "v">
+				<$enc := newVar "enc">
+				// MarshalLogObject implements zapcore.ObjectMarshaler, enabling
+				// fast logging of <.Name>.
+				func (<$m> <.Name>) MarshalLogObject(<$enc> <$zapcore>.ObjectEncoder) error {
+					for <$k>, <$v> := range <$m> {
+						<- $encAdd := printf "%s.Add%s((string)(%s), %s)" $enc (zapEncoder .Type.ValueSpec) $k (zapMarshaler .Type.ValueSpec $v)>
+						<if (zapCanError .Type.ValueSpec) ->
+							if err := <$encAdd>; err != nil {
+								return err
```
abhinav: Damn, this is getting repetitive. I wonder if we can clean up this `err != nil` business.

Can we maybe do something like,

```
<zapEncodeBegin .Type.ValueSpec>
  <$enc>.Add<zapEncoder ..>(..., <zapMarshaler .Type.ValueSpec $v>)
<zapEncodeEnd .Type.ValueSpec>
```

Where zapEncodeBegin and zapEncodeEnd add the `if err != nil`, etc. if needed?

This will handle all these cases where we're doing `if zapCanErr` and
`$errAdd` again and again.

gen/typedef.go
```diff
@@ -103,6 +103,26 @@ func typedef(g Generator, spec *compile.TypedefSpec) error {
 				return <equals .Target $lhs $rhs>
 			<- end>
 		}
+
+		<if (eq (zapEncoder .Target) "Object") ->
```
abhinav: The template format supports template-level comments. Let's add a couple to explain what's going on here.

```
</* ... */>
```

gen/typedef.go
```diff
@@ -103,6 +103,26 @@ func typedef(g Generator, spec *compile.TypedefSpec) error {
...
+			<$zapcore := import "go.uber.org/zap/zapcore">
+			<$enc := newVar "enc">
+			func (<$v> <$typedefType>) MarshalLogObject(<$enc> <$zapcore>.ObjectEncoder) error {
+				<if isStructType . ->
+					return ((<typeReference .Target>)(<$v>)).MarshalLogObject(<$enc>)
```
abhinav: Isn't this case handled by `zapMarshaler . $v` too?

gen/zap.go
```diff
@@ -0,0 +1,132 @@
...
+	case *compile.BoolSpec, *compile.I8Spec, *compile.I16Spec, *compile.I32Spec,
+		*compile.I64Spec, *compile.DoubleSpec, *compile.StringSpec, *compile.BinarySpec:
+		return false
+
+	// Non-primitives
+	case *compile.MapSpec, *compile.SetSpec, *compile.ListSpec, *compile.EnumSpec,
+		*compile.StructSpec:
+		return true
+	}
+	panic(root)
```
abhinav: Let's drop the panic. You can add checks for the true cases since there are
fewer of those and return false by default for all other cases.

Alternatively, you can also just call `zapEncoder` on the type and return true
only if the output is "Object" or "Array".

(If you do the zapEncodeBegin/zapEncodeEnd business, you can probably drop
this whole function.)

mh-park: did you want to drop panic for the other functions as well?

abhinav: For the cases we're doing exhaustive checks, let's keep the panic (but with a better error message).

For cases like this where it's more of an explicit list of things that can error out, we can just do catch-all for the false case rather than panicking.

gen/map.go
```diff
@@ -281,3 +281,147 @@ func (m *mapGenerator) equalsUnhashable(g Generator, spec *compile.MapSpec) (str
...
+			<$m := newVar "m">
+			<$k := newVar "k">
+			<$v := newVar "v">
+			<$enc := newVar "enc">
+			// MarshalLogObject implements zapcore.ObjectMarshaler, enabling
+			// fast logging of <.Name>.
+			func (<$m> <.Name>) MarshalLogObject(<$enc> <$zapcore>.ObjectEncoder) error {
+				for <$k>, <$v> := range <$m> {
+					<zapEncodeBegin .Type.ValueSpec ->
+						<$enc>.Add<zapEncoder .Type.ValueSpec>((string)(<$k>), <zapMarshaler .Type.ValueSpec $v>)
```
akshayjshah: If `$k` is already a string, do we need the `(string)` here?

gen/enum.go
```diff
@@ -125,6 +125,23 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+		// MarshalLogObject implements zapcore.ObjectMarshaler, enabling
+		// fast logging of <$enumName>.
+		func (<$v> <$enumName>) MarshalLogObject(enc <$zapcore>.ObjectEncoder) error {
+			enc.AddInt32("value", int32(<$v>))
+			<if len .Spec.Items ->
+				switch int32(<$v>) {
+				<range .UniqueItems ->
+					case <.Value>:
+						enc.AddString("name", "<.Name>")
+				<end ->
```
akshayjshah: I assume that there's no `default` case in the switch because there's no safe default name, correct? If so, can we leave a comment in the generated code explaining that?

gen/list.go
```diff
@@ -177,3 +181,40 @@ func (l *listGenerator) Equals(g Generator, spec *compile.ListSpec) (string, err
 
 	return name, wrapGenerateError(spec.ThriftName(), err)
 }
+
+func (l *listGenerator) zapMarshaler(
```
abhinav: Let's document what we're logging things as.

    // Slices are logged as JSON arrays.

gen/map.go
```diff
@@ -281,3 +281,147 @@ func (m *mapGenerator) equalsUnhashable(g Generator, spec *compile.MapSpec) (str
 
 	return name, wrapGenerateError(spec.ThriftName(), err)
 }
+
+func (m *mapGenerator) zapMarshaler(
```
abhinav: We should document this one in detail because it has non-obvious behavior.

    // Maps are logged as objects if the key is a string or a typedef of a
    // string. If the key is not a string, maps are logged as arrays of
    // objects with a key and value.
    //
    //   map[string]int32{"foo": 1, "bar": 2}
    //   => {"foo": 1, "bar": 2}
    // 
    //   map[int32]string{1: "foo", 2: "bar"}
    //   => [{"key": 1, "value": "foo"}, {"key": 2, "value": "bar"}]

gen/list.go
```diff
@@ -177,3 +181,40 @@ func (l *listGenerator) Equals(g Generator, spec *compile.ListSpec) (string, err
...
+	g Generator,
+	root *compile.ListSpec,
```
abhinav: nit: s/root/spec

In this context, there's only the "spec", no "root".

gen/list.go
```diff
@@ -177,3 +181,40 @@ func (l *listGenerator) Equals(g Generator, spec *compile.ListSpec) (string, err
...
+			type <.Name> <typeReference .Type>
+			<$l := newVar "l">
+			<$v := newVar "v">
+			<$enc := newVar "enc">
+			// MarshalLogArray implements zapcore.ArrayMarshaler, enabling
+			// fast logging of <.Name>.
+			func (<$l> <.Name>) MarshalLogArray(<$enc> <$zapcore>.ArrayEncoder) error {
+				for _, <$v> := range <$l> {
+					<zapEncodeBegin .Type.ValueSpec ->
+						<$enc>.Append<zapEncoder .Type.ValueSpec>(<zapMarshaler .Type.ValueSpec $v>)
```
akshayjshah: @abhinav This is fine, but it has the minor downside of bailing out on the first error. This potentially leaves other errors lurking later in the array. How complex would it be to use `multierr` here and collect all the errors in one pass?

Completely fine if we leave this for a follow-up (or never do it).

gen/map.go
```diff
@@ -281,3 +281,147 @@ func (m *mapGenerator) equalsUnhashable(g Generator, spec *compile.MapSpec) (str
...
+			type <.Name> <typeReference .Type>
+			<$m := newVar "m">
+			<$k := newVar "k">
+			<$v := newVar "v">
+			<$i := newVar "i">
+			<$enc := newVar "enc">
+			// MarshalLogArray implements zapcore.ArrayMarshaler, enabling
+			// fast logging of <.Name>.
+			func (<$m> <.Name>) MarshalLogArray(<$enc> <$zapcore>.ArrayEncoder) error {
+				<- if isHashable .Type.KeySpec ->
```
akshayjshah: :sob: Thrift :sob:

gen/map.go
```diff
@@ -281,3 +281,147 @@ func (m *mapGenerator) equalsUnhashable(g Generator, spec *compile.MapSpec) (str
...
+			`, struct {
+			Name string
+			Type *compile.MapSpec
+		}{
+			Name: name,
+			Type: root,
+		},
+	); err != nil {
+		return "", err
+	}
```
abhinav: optional nit: can return err with the sprintf. There's no chance of that
panicking because name and fieldValue are strings (non-nillable).

```
err := g.EnsureDeclared(`
  ...
`, ...)
return fmt.Sprintf(...), err
```

gen/map.go
```diff
@@ -281,3 +281,147 @@ func (m *mapGenerator) equalsUnhashable(g Generator, spec *compile.MapSpec) (str
...
+				return nil
+			}
+			`, struct {
+			Name string
+			Type *compile.MapSpec
+		}{
+			Name: name,
+			Type: root,
+		},
+		TemplateFunc("zapMapItemMarshaler", curryGenerator(m.zapMapItemMarshaler, g)),
```
abhinav: Looking at TemplateFunc docs, you don't need to `curryGenerator`. That's done
automatically if the provided function accepts a Generator as its first
argument.

gen/map.go
```diff
@@ -281,3 +281,147 @@ func (m *mapGenerator) equalsUnhashable(g Generator, spec *compile.MapSpec) (str
...
+		return "", err
+	}
+	return fmt.Sprintf("(%v)(%v)", name, fieldValue), nil
+}
+
+func (m *mapGenerator) zapMapItemMarshaler(
+	g Generator,
+	keySpec compile.TypeSpec,
+	keyVar string,
+	valueSpec compile.TypeSpec,
```
abhinav: optional: instead of passing keySpec and valueSpec, can pass in just the MapSpec.

gen/map.go
```diff
@@ -281,3 +281,147 @@ func (m *mapGenerator) equalsUnhashable(g Generator, spec *compile.MapSpec) (str
...
+	valueVar string,
+) (string, error) {
+	name := fmt.Sprintf("_MapItem_%s_%s_Zapper", g.MangleType(keySpec), g.MangleType(valueSpec))
```
abhinav: optional: can mangle the MapSpec and append `_Item_Zapper` to that instead.

gen/zap.go
akshayjshah: :man_facepalming: My fault, noted this in https://github.com/uber-go/zap/issues/388.

gen/zap.go
```diff
@@ -0,0 +1,148 @@
...
+	case *compile.DoubleSpec:
+		return "Float64"
+	case *compile.StringSpec:
+		return "String"
+	case *compile.BinarySpec:
+		return "String" // encode binary as a string and log as string
+
+	// Containers
+	case *compile.MapSpec:
+		switch t.KeySpec.(type) {
```
abhinav: `compile.RootTypeSpec(t.KeySpec)` because the map key may be a typedef of a
string. If we don't have tests for this, we should add a test case to
typedefs.thrift.

gen/zap.go
```diff
@@ -0,0 +1,148 @@
...
+	case *compile.EnumSpec, *compile.StructSpec:
+		return "Object"
+	}
+	panic(root)
+}
+
+// zapMarshaler takes a TypeSpec, evaluates whether there are underlying elements
+// that require more Zap implementation to log everything, and returns a string
+// that properly casts the fieldValue, if needed, for logging.
+func (z *zapGenerator) zapMarshaler(g Generator, spec compile.TypeSpec, fieldValue string) (string, error) {
```
abhinav: Nice. Can you add,

    // This should be used in conjunction with zapEncoder,
    //
    //   v := ...
    //   enc.Add<zapEncoder .Type>("foo", <zapMarshaler .Type "v">)

gen/zap.go
abhinav: @akshayjshah FYI. All binary will be logged as base64. We will never rely on AddBinary because there's no AppendBinary.

akshayjshah:
> @akshayjshah FYI. All binary will be logged as base64. We will never rely on AddBinary because there's no AppendBinary.

Concurrent review :) Yeah, this is totally my fault - I added a note to zap's 2.0 wishlist about this. For now, it's fine as-is.

gen/list.go
abhinav: I think it'll be a bit complex because of the different corner cases with some
encoders returning errors, some not.

I'll create a follow-up task to explore that in the future but for v1 of this,
we can skip it IMO.

gen/map.go
abhinav: It may be a typedef of a string or a typedef of a typedef of a string.

gen/enum.go
abhinav: +1

abhinav: Quick update: Comments in the code aren't propagated by the templating logic
we have here. Godocs are propagated, though, so @mh-park added a comment about
how enums are serialized as a godoc.

PR #369, opened 2018-08-14
Add no-zap flag for opting out of Zap logging completely

mh-park:


gen/generator.go
```diff
@@ -139,19 +140,36 @@ type generator struct {
 	// TODO use something to group related decls together
 }
 
+type GeneratorOptions struct {
+	TImport     ThriftPackageImporter
```
abhinav: nit: "Importer"

gen/generate.go
```diff
@@ -230,7 +233,13 @@ func generateModule(m *compile.Module, i thriftPackageImporter, builder *generat
 	// will prepend $packageRelPath/ to all these paths.
 	files := make(map[string][]byte)
 
-	g := NewGenerator(i, importPath, packageName)
+	gopts := &GeneratorOptions{
+		TImporter:   i,
```
abhinav: nit: Call this Importer please

gen/generator.go
```diff
@@ -139,19 +140,36 @@ type generator struct {
...
-		thriftImporter: timport,
+		thriftImporter: o.TImporter,
 		fset:           token.NewFileSet(),
+		noZap:          o.NoZap,
+	}
+}
+
+// checkNoZap returns whether the NoZap flag is passed.
+func checkNoZap(g Generator) bool {
+	if castedGenerator, ok := g.(*generator); ok {
```
abhinav: nit: shorter variable names are okay when there are only 2-3 variables in scope. `gen` would be fine here.

gen/internal/tests/thrift/nozap.thrift
```diff
@@ -0,0 +1,17 @@
...
+    3: required i16 int16Field
+    4: required i32 int32Field
+    5: required i64 int64Field
+    6: required double doubleField
+    7: required string stringField
+    8: required binary binaryField
+}
+
+typedef i64 Timestamp  // alias of primitive
+typedef string State
```
abhinav: These typedef cases won't get MarshalLogObject/Array even with the flag.
 Can we replace them with,

    typedef map<string, string> StringMap
    typedef PrimitiveRequiredStruct Primitives
    typedef list<string> StringList

This will exercise the cases we care about, I think.

gen/internal/tests/thrift/nozap.thrift
```diff
@@ -0,0 +1,17 @@
+enum EnumDefault {
+    Foo, Bar, Baz
+}
+
+struct PrimitiveRequiredStruct {
+    1: required bool boolField
+    2: required byte byteField
+    3: required i16 int16Field
+    4: required i32 int32Field
+    5: required i64 int64Field
```
abhinav: I would make some of these fields containers so that it would *normally* generate zap wrappers for them.

gen/generator.go
```diff
@@ -139,19 +140,36 @@ type generator struct {
 	// TODO use something to group related decls together
 }
 
+type GeneratorOptions struct {
```
abhinav: This needs comments. If you rebase on master, the linter will fail on this now.

PR #373, opened 2018-08-15
Add zap opt out support via thrift annotations

### LINKED ISSUE #371: zap: Opt out sensitive struct fields
MarshalLogObject on struct fields should support opting specific fields of a
struct out of being logged in case they contain sensitive information.

    struct User {
        1: required string username
        2: required string emailAddress (go.log = "false")
    }
    
The annotation name is up for discussion. 

CC @prashantv @akshayjshah @kriskowal

mh-park:
Resolves #371

### DIFF
--- a/gen/zap.go
+++ b/gen/zap.go
@@ -27,12 +27,13 @@ import (
 )
 
 // NoZapLabel allows opt out of Zap logging for struct fields.
-// Thrift structs will use this annotation to omit it from logging, i.e.
+// Fields of Thrift structs will use this annotation to opt-out of being logged
+// when that struct is logged. i.e.
 //
-// struct ZapOptOutStruct {
-// 	1: required string name
-// 	2: required string optout (go.nolog)
-// }
+// 	struct ZapOptOutStruct {
+// 		1: required string name
+// 		2: required string optout (go.nolog)
+// 	}
 //
 // The above struct will be logged without the optout string.
 const NoZapLabel = "go.nolog"


gen/zap.go
```diff
@@ -26,6 +26,8 @@ import (
 	"go.uber.org/thriftrw/compile"
 )
 
+const NoZapLabel = "go.nolog"
```
abhinav: This should have docs on it. The linter would have complained about it but I
just realized that the linter hasn't been running in CI for a while. Fixing.

gen/zap.go
```diff
@@ -26,6 +26,17 @@ import (
 	"go.uber.org/thriftrw/compile"
 )
 
+// NoZapLabel allows opt out of Zap logging for struct fields.
+// Thrift structs will use this annotation to omit it from logging, i.e.
+//
+// struct ZapOptOutStruct {
+// 	1: required string name
+// 	2: required string optout (go.nolog)
```
abhinav: Indent code samples at least one more space to have them render as code blocks.

gen/zap.go
```diff
@@ -26,6 +26,17 @@ import (
 	"go.uber.org/thriftrw/compile"
 )
 
+// NoZapLabel allows opt out of Zap logging for struct fields.
+// Thrift structs will use this annotation to omit it from logging, i.e.
```
abhinav: Fields of Thrift structs will use this annotation to opt-out of being logged
when that struct is logged.

gen/field.go
```diff
@@ -91,8 +91,10 @@ func (f fieldGroupGenerator) Generate(g Generator) error {
 		return err
 	}
 
-	if err := f.Zap(g); err != nil {
-		return err
+	if !checkNoZap(g) {
```
amckinney: Reads a little funky with the double negative, but I see that the semantics are taken from other similar helper functions like `checkReservedIdentifier`.

gen/zap_test.go
```diff
@@ -476,6 +476,20 @@ func TestCrazyTownZapLogging(t *testing.T) {
 	}
 }
 
+func TestOptOutOfZap(t *testing.T) {
+	// These types are created to ease building map[string]interface{}
```
amckinney: Might not be necessary since it's only used in one place.

PR #375, opened 2018-08-17
gen/enum: TextUnmarshaler should try to parse the integer

### LINKED ISSUE #368: enum: TextUnmarshaler should try to parse the integer
Currently, the following won't work,

```
item := SomeEnum(249)  // where 249 is unrecognized
b, err := item.MarshalText()  // b == "249"
require.NoError(t, err)

var gotItem SomeEnum
require.NoError(t, gotItem.UnmarshalText(b))
assert.Equal(t, item, gotItem)
``` 

The UnmarshalText will fail because it won't be able to match "249".

Ideally, MarshalText and UnmarshalText can be round-tripped.

mh-park:
`MarshalText` and `UnmarshalText` now round-trips, even if the enum value is unrecognized. `UnmarshalText` will default to parsing the integer value, and will only return an error if the parsing fails.

Resolves #368


gen/enum.go
```diff
@@ -96,23 +96,32 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
 			<range .Spec.Items ->
 				case "<enumItemLabelName .>":
 					*<$v> = <enumItemName $enumName .>
 					return nil
 			<end ->
 				default:
-					return <$fmt>.Errorf("unknown enum value %q for %q", value, "<$enumName>")
+					<$val>, <$err> := <$strconv>.ParseInt(<$s>, 10, 64)
+					if <$err> != nil {
+						return <$fmt>.Errorf("unknown enum value %q for %q: %v", <$value>, "<$enumName>", <$err>)
```
abhinav: that `$value` should probably be `$s` since we've already made a `string` version of the `[]byte`.

gen/enum.go
```diff
@@ -96,23 +96,32 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
 		//
 		//   var <$v> <$enumName>
 		//   err := <$v>.UnmarshalText([]byte("<(index .Spec.Items 0).Name>"))
 		<- end>
-		func (<$v> *<$enumName>) UnmarshalText(value []byte) error {
-			switch string(value) {
+		func (<$v> *<$enumName>) UnmarshalText(<$value> []byte) error {
+			<- $s := newVar "s" ->
+			<- $val := newVar "val" ->
+			<- $err := newVar "err">
```
abhinav: nit: don't need `newVar` for err since the convention is to always shadow `err`. The newVar business tries to ensure that we don't have two templated variables named "val", "s", or something conflicting with an import in this scope.

gen/enum_test.go
```diff
@@ -508,19 +508,19 @@ func TestEnumLabelInvalidUnmarshal(t *testing.T) {
 	}{
 		{
 			`"some-random-str"`,
-			`unknown enum value "some-random-str" for "EnumWithLabel"`,
+			`unknown enum value "some-random-str" for "EnumWithLabel": strconv.ParseInt: parsing "some-random-str": invalid syntax`,
```
abhinav: optional: you can simplify this test by making it an `assert.Contains(t, err.Error(), msg)` check instead of `assert.Equal`. This will let you drop the part of the error message that's not under your control: the strconv.ParseInt failure.

abhinav:
Don't forget the changelog!

abhinav:
RE: Coverage: The coverage of the generated code went down because we have
more generated code now. It should be safe to ignore.

We can theoretically have the coverage report omit generated code if we care
about the actual coverage number, but I think it's valuable to keep it as-is
and use code coverage how it was intended: to find portions of the code that
need more testing. In this case, those code paths are tested elsewhere.

gen/enum.go
```diff
@@ -96,23 +96,31 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+			switch <$s> := string(<$value>); <$s> {
 			<- $enum := .Spec ->
 			<range .Spec.Items ->
 				case "<enumItemLabelName .>":
 					*<$v> = <enumItemName $enumName .>
 					return nil
 			<end ->
 				default:
-					return <$fmt>.Errorf("unknown enum value %q for %q", value, "<$enumName>")
+					<$val>, err := <$strconv>.ParseInt(<$s>, 10, 64)
```
prashantv: given that enums are `int32` (line 79), should this be `ParseInt(<$s>, 10, 32)` (32 bits instead of 64 bits)

Otherwise, we silently truncate 64-bit values to 32-bit

PR #377, opened 2018-08-27
gen: Support go.label on struct fields

### LINKED ISSUE #370: go.label support for struct fields
Similar to #362, we should support overriding the label for struct fields with
a `go.label` annotation.

```
struct Foo {
    1: optional string f (go.label = "function")
}
```

abhinav:
This adds support for the `go.label` annotation on struct fields.
Specifying this label on struct fields has two effects:

- The name used for the field during logging with Zap will use the
  value specified in `go.label`.
- The name of the field used while converting the struct to/from JSON
  will use the value specified in `go.label` **unless** a different
  value was specified with a `go.tag = 'json:".."'` annotation. I don't
  completely support making any assurances about the JSON
  representation of Thrift types but it seems to me that that ship
  sailed when we added the `json` tags.

Note that this intentionally leaves out the `String()` methods for
structs. Those methods currently output valid Go syntax with an
implication that the printed code can be used to rebuild the Go object.

Resolves #370

(This PR has another refactoring/cleanup commit that can be reviewed
separately.)


abhinav:
CC @cl1337

mh-park:
@abhinav Should we be using go.label for `String()`?

abhinav:
Good question. No, because the `String()` for structs currently returns Go
code that you would use to instantiate that struct. Changing it in `String()`
will break that code. This is basically a discrepancy between `String()` for
structs and `String()` for enums.

PR #378, opened 2018-08-27
gen/zap: Record all errors

### LINKED ISSUE #367: Logging: Record all errors
With #366, we fail marshaling for containers and structs when the first item
fails to encode. We should instead use multierr and combine all the errors.

abhinav:
When appending/adding an object or array to a Zap encoder, there's a
possibliity of failure. Previously, we were recording the first failure
and returning.

This changes the generated code to rely on `multierr` and collect all
such failures into a single error.

Resolves #367


abhinav:
Zap won't log the objects if they fail to marshal.

abhinav:
Going to merge and release this shortly if there are no objections.

akshayjshah:
> Zap won't log the objects if they fail to marshal.

This isn't *quite* true. It's entirely possible for implementations of `zapcore.ObjectMarshaler` and `zapcore.ArrayMarshaler` to corrupt the output. All implementations of these methods should be doing any required validation before they write to the encoder. 

Here, I think we're safe - everything is (eventually) primitive types, so none of these implementations can ever actually fail.

PR #379, opened 2018-08-28
gen/struct: Make Equals nil-safe

### LINKED ISSUE #345: Generated Equals methods should check for nil receiver
The Equals method is not nil-safe. We should return if this and that are both nil. Should return false otherwise if the receiver is nil.

abhinav:
The generated Equals methods on structs weren't nil safe and would panic
if either value was nil. This change adds a nil check on the generated
Equals methods. Two structs are considered equal if they're both nil.

Resolves #345

### DIFF
--- a/gen/field.go
+++ b/gen/field.go
@@ -454,6 +454,11 @@ func (f fieldGroupGenerator) Equals(g Generator) error {
 		//
 		// This function performs a deep comparison.
 		func (<$v> *<.Name>) Equals(<$rhs> *<.Name>) bool {
+			if <$v> == nil {
+				return <$rhs> == nil
+			} else if <$rhs> == nil {
+				return false
+			}
 			<range .Fields>
 				<- $fname := goName . ->
 				<- $lhsField := printf "%s.%s" $v $fname ->
--- a/gen/struct_test.go
+++ b/gen/struct_test.go
@@ -29,6 +29,7 @@ import (
 	tc "go.uber.org/thriftrw/gen/internal/tests/containers"
 	te "go.uber.org/thriftrw/gen/internal/tests/enums"
 	tx "go.uber.org/thriftrw/gen/internal/tests/exceptions"
+	tf "go.uber.org/thriftrw/gen/internal/tests/services"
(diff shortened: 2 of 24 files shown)


gen/field.go
```diff
@@ -454,6 +454,11 @@ func (f fieldGroupGenerator) Equals(g Generator) error {
 		//
 		// This function performs a deep comparison.
 		func (<$v> *<.Name>) Equals(<$rhs> *<.Name>) bool {
+			if <$v> == nil {
```
mh-park: optional:
```
if <$v> == nil || <$rhs> == nil {
  return <$v> == <$rhs>
}
```
would have the same logic in less lines, but maybe less readable.

abhinav: Hah, that was what I originally had. I decided more readable was preferable to
fewer lines.

abhinav:
I'm not sure what we can do about the coverage here outside of performing the
same check on every generated struct in internal/tests.

gen/struct_test.go
```diff
@@ -1856,3 +1857,111 @@ func TestStructLabel(t *testing.T) {
...
+		tf.KeyValue_DeleteValue_Args{},
+		tf.KeyValue_DeleteValue_Result{},
+		tf.KeyValue_GetValue_Args{},
+		tf.KeyValue_GetValue_Result{},
+		tf.NonStandardServiceName_NonStandardFunctionName_Args{},
+		tf.NonStandardServiceName_NonStandardFunctionName_Result{},
+		tf.Cache_ClearAfter_Args{},
+		tf.KeyValue_Size_Args{},
+		tf.KeyValue_Size_Result{},
+	}
```
akshayjshah: Small nit: would `structs` or `instances` be a better name for this slice? We're doing this below:

```
for _, st := range structTypes {
  structType := reflect.TypeOf(st)
```

so I'd expect `structTypes` to be a slice of `reflect.Type`s.

abhinav: In #383, I decided to refer to these as "samples" as-in "a sample of the kind of value I'm testing". Switching to that here.

PR #380, opened 2018-08-28
compile/struct: Check field ID bounds

### LINKED ISSUE #72: Bounds check on struct field IDs

abhinav:
Thrift field IDs are sent over the wire as 16-bit integers. Previously,
we were silently truncating the values specified in the IDL down to 16
bits.

This changes the compiler to validate field IDs during compilation. In
addition to field IDs that are too large, we reject the following as
well:

- Field ID 0: This is reserved for the return value of a non-void
  method.
- Negative field IDs: This is just wrong

Resolves #72


compile/field.go
```diff
@@ -102,6 +102,10 @@ type FieldSpec struct {
 
 // compileField compiles the given Field source into a FieldSpec.
 func compileField(src *ast.Field, options fieldOptions) (*FieldSpec, error) {
+	if src.ID < 1 || src.ID > 32767 {
```
prashantv: you can use `math.MaxInt16` here

PR #381, opened 2018-08-28
travis: Test against Go 1.11, drop 1.9

abhinav:
Now that Go 1.11 is out, we should start testing against it. Plus to
match the [Go Release Policy][1], we are now free to drop support for
Go 1.9, which lets us rely on the built-in coverage merging functionality
instead of our hacky cover.sh script.

  [1]: https://golang.org/doc/devel/release.html#policy


abhinav:
It looks like there's an issue in `go tool cover` in Go 1.11.

```
go tool cover -html=cover.out -o cover.html
cover: inconsistent NumStmt: changed from 2 to 3
```

This is reproducible locally. I'll try to figure out what's going on tomorrow.

akshayjshah:
Seems reasonable, and I'm really excited to drop all our janky coverage-merging scripts. I'll be bummed out if there's really a regression in `go tool cover`.

abhinav:
I've narrowed the issue down to the `//line` comments in the lex.go generated
by Ragel. Filed an issue (golang/go#27350) in case this is a regression.

Meanwhile, I'm going to filter out the `//line` clauses from the generated
files.

abhinav:
I think I've driven the coverage as much as I can in this change. We can make
more follow up changes to improve coverage.

This is good to go @akshayjshah @AlexAPB @amckinney. Commits are individually
reviewable or all-in-one go based on your preference.

abhinav:
To clarify: Previously, our coverage merging script was ignoring a number of
things when counting code coverage so we managed to get to ~85% with the
recent changes. With this change, `-coverpkg ./...` counts all Go code in
ThriftRW for coverage, including things that were previously ignored. We can
make follow-up changes to improve testing/coverage but I would rather not
block updating to the new compiler on that.

Makefile
```diff
@@ -104,9 +104,24 @@ verifyversion: build
...
+
+# literal space
+space :=
+space +=
+
 .PHONY: cover
 cover:
-	go test -v -covermode=atomic -coverprofile cover.out -coverpkg=./... ./...
+	go test -v -covermode=atomic -coverprofile cover.full.out -coverpkg=./... ./...
+	grep -v "$(subst $(space),\|,$(COVER_IGNORE_FILES))" cover.full.out > cover.out
```
AlexAPB: why do we need to replace spaces with `|`? nicer output?

abhinav: We're building a regular expression so we need to replace the spaces with `|` clauses.

AlexAPB: whoops, gotcha

PR #382, opened 2018-08-28
make/cover: Use built-in coverage merging

abhinav:
Since Go 1.9, we haven't had need for our hacky coverage script. This
changes our build setup to rely on the coverage merging built into the
tooling.


abhinav:
Derp, I guess I meant Go 1.10. Folding into #381.

PR #383, opened 2018-08-29
gen: testing/quick based tests

abhinav:
We had a stale `testing/quick` based test in the repository from way
back that did some basic round-tripping.

This PR adds a series of commits that expand that to test most of the
generated types in internal/tests.

The following functionalties are tested thus far:

- `wire.Value` round-tripping
- `String()`
- Zap logging
- For enums, JSON and encoding.TextMarhsaler round-tripping

I found a minor bug for empty enums where we weren't generating the
`*_Values` function. This fixes that bug too.

For reviewers: Every commit is individually reviewable. In fact, I would
recommend that route rather than the whole change.

### DIFF
--- a/gen/quick_test.go
+++ b/gen/quick_test.go
@@ -28,55 +28,151 @@ import (
 	"time"
 
 	tc "go.uber.org/thriftrw/gen/internal/tests/containers"
+	te "go.uber.org/thriftrw/gen/internal/tests/enums"
+	tx "go.uber.org/thriftrw/gen/internal/tests/exceptions"
+	tf "go.uber.org/thriftrw/gen/internal/tests/services"
 	ts "go.uber.org/thriftrw/gen/internal/tests/structs"
+	td "go.uber.org/thriftrw/gen/internal/tests/typedefs"
+	tu "go.uber.org/thriftrw/gen/internal/tests/unions"
+	"go.uber.org/thriftrw/wire"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
+func generateValue(t *testing.T, typ reflect.Type, rand *rand.Rand) thriftType {
+	for {
+		// We will keep trying to generate a value until a valid one
+		// is found.
+
+		v, ok := quick.Value(typ, rand)
+		require.True(t, ok, "failed to generate a value")
+
+		tval := v.Addr().Interface().(thriftType)
+
+		// TODO(abg): ToWire + EvaluateValue to validate here means we end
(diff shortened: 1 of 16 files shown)


gen/service_test.go
```diff
@@ -665,11 +665,9 @@ func TestArgsAndResultValidation(t *testing.T) {
 			t.Fatalf("invalid test %q: either typ or serialize must be set", tt.desc)
 		}
 
-		x := reflect.New(typ)
-		args := []reflect.Value{reflect.ValueOf(tt.deserialize)}
-		e := x.MethodByName("FromWire").Call(args)[0].Interface()
-		if assert.NotNil(t, e, "%v: expected failure but got %v", tt.desc, x) {
-			assert.Contains(t, e.(error).Error(), tt.wantError, tt.desc)
+		x := reflect.New(typ).Interface().(serviceType)
```
AlexAPB: optional nit: I don't have enough background on the existing tests so this may be unnecessary, but you may want to use an `ok` check with `require` so if this test ever fails we don't see a panic.

(I see this three times in the second commit)

gen/quick_test.go
```diff
@@ -260,6 +266,30 @@ func TestQuickRoundTrip(t *testing.T) {
 					}, "failed to String %#v", give)
 				}
 			})
+
+			if !tt.NoLog {
+				t.Run("Zap", func(t *testing.T) {
+					for _, give := range values {
+						assert.NotPanics(t, func() {
```
AlexAPB: Is this necessary? What can panic here?

mh-park: It panics if we see an unknown type instead of having some default behavior. It shouldn't actually ever panic because we've covered every type.

abhinav: Well, nothing should panic, but in that case I also wanted a function I could
`return` from for flow control.

gen/service_test.go
abhinav: Ibe more informative than something hand-written

abhinav: I considered that but the message from the runtime when that panics
will be more informative than the static string we can write there. (It
mentions which methods are missing, whether the method was on the
pointer rather than the value, etc.)

In this case, the panic has the same effect as failing the test for us:
catastrophic failure because we have a major bug in the codegen.

I would prefer to leave this as-is unless you feel strongly about this.

AlexAPB: sgtm 👍  just making sure it wasn't an oversight

gen/quick_test.go
AlexAPB: sgtm

PR #387, opened 2018-08-31
gen/test/quick: Refactor and add more tests

abhinav:
This refactors the randomized test in gen/ to be more readable; instead
of having all tests in one large block, I've pulled out the different
tests for different things into separate functions that are called from
the main test.

This made room for testing more functionality without having a
super-noisy test so I added a few more minor tests.

Commits are recommended to be reviewed individually.


gen/quick_test.go
```diff
@@ -144,6 +144,63 @@ func enumValueGenerator(valuesFunc interface{}) func(*testing.T, *rand.Rand) thr
...
+			result.DoesNotExist = doesNotExistGenerator(t, rand).(*tx.DoesNotExistException)
+		}
+		return &result
+	}
+}
+
+func keyValueGetManyValuesResultGenerator() func(*testing.T, *rand.Rand) thriftType {
+	arbitraryValueGenerator := unionValueGenerator(tu.ArbitraryValue{})
+	successGenerator := func(t *testing.T, rand *rand.Rand) []*tu.ArbitraryValue {
+		values := make([]*tu.ArbitraryValue, rand.Intn(10))
```
amckinney: Not at all necessary now, but if there's ever a need we could add `numVals` as a parameter and update to:
```
func keyValueGetManyValuesResultGenerator(numVals int) func(*testing.T, *rand.Rand) thriftType {
  ...
    values := make([]*tu.ArbitraryValue, rand.Intn(numVals))
    ...
```

gen/quick_test.go
```diff
@@ -625,3 +671,29 @@ func (q *quickSuite) testEnumPtr(t *testing.T, give thriftType) {
...
+		if field.Type.Kind() == reflect.Ptr && field.Type.Elem().Kind() != reflect.Struct {
+			valueType := field.Type.Elem()
+			if fieldValue.IsNil() {
+				fieldValue = reflect.Zero(valueType)
+			} else {
+				fieldValue = fieldValue.Elem()
+			}
+		}
+
+		assert.Equal(t, fieldValue.Interface(), accessorValue.Interface())
```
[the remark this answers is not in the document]
abhinav: Yeah, the assumption is that someone reading this is familiar with the
`reflect` package to some degree.

gen/quick_test.go
abhinav: We can add it in the future if we need it. There's no need for it yet.
Honestly, as long as we exercise, 0, 1, and >1, the number here doesn't
matter.

PR #389, opened 2018-10-02
gen/plugin: Add pointer to typedefs of structs

mh-park:
Typedefs of structs were mishandled with plugins; while they should have
been pointers, they were generated without `*`. This commit addresses
this issue.

### DIFF
--- a/gen/plugin.go
+++ b/gen/plugin.go
@@ -343,11 +343,8 @@ func (g *generateServiceBuilder) buildType(spec compile.TypeSpec, required bool)
 			},
 		}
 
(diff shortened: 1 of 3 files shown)


gen/plugin.go
```diff
@@ -343,6 +343,10 @@ func (g *generateServiceBuilder) buildType(spec compile.TypeSpec, required bool)
 			},
 		}
 
+		if isStructType(spec) {
+			return &api.Type{PointerType: t}, nil
+		}
+
```
abhinav: nit: Can we collapse this into the if condition below?

PR #394, opened 2018-10-18
gen/struct: IsSet methods for nillable fields

abhinav:
This changes struct generation to include `IsSet*` methods for nillable
fields of a struct. Note that this applies not only to fields marked
optional, but also to required fields that can be nil: maps, slices,
binary, structs, etc.

This change makes it easier to migrate code from Apache Thrift to
ThriftRW.

Ref: T2133401


andrewdeandrade:
LGTM, but you might want to get another review from someone more familiar with this repo.

gen/quick_test.go
```diff
@@ -793,3 +798,23 @@ func (q *quickSuite) testStructAccessors(t *testing.T, giveVal thriftType) {
 		assert.Equal(t, fieldValue.Interface(), accessorValue.Interface())
 	}
 }
+
+func (q *quickSuite) testIsSetAccessors(t *testing.T, giveVal thriftType) {
```
akshayjshah: I loooooove this test suite.

PR #396, opened 2018-12-19
gen/field: Make optional field accessors nil-safe

codygibb:
The generated IsSetX and GetX methods on optional fields weren't nil
safe and would panic if the receiver was nil. This change adds a nil
check on said methods. This allows chaining field accessors on nested
optional structs.


gen/internal/tests/thrift/structs.thrift
```diff
@@ -89,9 +89,14 @@ struct ContactInfo {
     1: required string emailAddress
 }
 
+struct PersonalInfo {
+    1: optional i32 age
```
zmt: I love your example here.

PR #398, opened 2019-01-18
plugin: Pass filenames, packagePrefix, and thriftRoot

### LINKED ISSUE #348: plugin.Api GenerateServiceRequest should contain the thrift file path
Hi,

It would be amazing if the `api.GenerateServiceRequest` had a field containing the file path of its thrift file or a pointer to its original ast (not the api.Ast).

The reason for this is that for every `api.GenerateServiceRequest` I have to generate another thrift definition containing a subset of the service functions (using annotations).

mattgode:
This PR makes plugin API changes to allow for more custom code
generation in plugins. Namely it does the following: 

1. Expose the thriftFileName on the `api.Module`
2. Expose the packagePrefix on the `api.GenerateServiceRequest`
3. Expose the thriftRoot on the `api.GenerateServiceRequest`

Resolves #348


plugin/api.thrift
```diff
@@ -242,6 +242,10 @@ struct Module {
      * absolute location of the directory.
      */
     2: required string directory
+    /**
+     * Import path for the thrift file from which the module was generated
```
```suggestion
     * Path to the Thrift file from which this module was generated.
```

plugin/api.thrift
```diff
@@ -341,6 +345,18 @@ struct GenerateServiceRequest {
      * definition in this map.
      */
     3: required map<ModuleID, Module> modules
+    /**
+     * Prefix for import paths of generated module In general, plugins should
```
```suggestion
     * Prefix for import paths of generated module. In general, plugins should
```

abhinav:
@mh-park @kriskowal Would one of you mind taking a look?

abhinav:
@kriskowal They have a few plugins that re-parse the Thrift files to
generate custom methods for structs and other types as we don't surface
this information in the plugin API.

They rely on some hacks to work back to the Thrift root and package
prefix so that they can use the [thriftrw/gen][1] APIs for code
generation.

The documentation in the Thrift file doesn't make it clear but the
purpose of the Thrift root is solely to inform the expected import path
of the generated code. The import path for a Thrift file is
`join($packagePrefix, relative($thriftRoot, $filePath))`. So if the
path to the Thrift file is idl/foo/bar.thrift, the Thrift root is idl/,
and the package prefix is example.com/gen, the import path for the
corresponding Thrift module is example.com/gen/foo/bar.

[1]: https://godoc.org/go.uber.org/thriftrw/gen

PR #401, opened 2019-01-23
plugin: Bump API Version

abhinav:
The previous release added new required fields to the ThriftRW Plugin
API. The fields aren't optional, so we should have bumped the API
Version to ensure that the ThriftRW binary and plugins agree about this.

Thanks to @twilly for pointing this out.


twilly:
LGTM

PR #403, opened 2019-01-30
[plugin] Add annotation support for function arguments

mh-park:
Previously, plugins could not reference annotations for arguments and
exceptions in functions, because they were omitted in the plugin API.
This commit adds these annotations to the plugin API.


plugin/api.thrift
```diff
@@ -130,6 +130,27 @@ struct Argument {
...
+     *   } (key = "id", validate)
+     *
+     * The annotations will be,
+     *
+     *   {
+     *     "key": "id",
+     *     "validate": "",
+     *   }
+     */
+    3: optional map<string, string> annotations
```
abhinav: Do you mean the annotation on the argument or the type? I would expect the argument annotations here.

```
void foo(
  1: string bar (x = "y")
)
```

abhinav: annotations for custom types are already exposed here https://github.com/thriftrw/thriftrw-go/pull/403/files/672994bbbd6cb0b9302bb91760659f38d4fe3d24#diff-6d5ca9a3f7768e98dd3dbcccfef08c37R50

mh-park: I mean to add annotations for this:

https://github.com/thriftrw/thriftrw-go/pull/403/files/672994bbbd6cb0b9302bb91760659f38d4fe3d24#diff-6d5ca9a3f7768e98dd3dbcccfef08c37R184

but it looks like function arguments and exceptions both use `list<Argument>` as its type in the API.

abhinav: exceptions and functions parameters are both 'arguments' at the AST level so annotations work on both.

```
void foo(
  1: string x (a = "b")
) throws (
  1: GreatSadnessError fail (c = "d")
)
```

Either way, the type you're adding is correct but the docs need to be updated.

mh-park: It looks like the comments in `api.thrift` might be inaccurate and unclear; I'll update them.

plugin/api.thrift
```diff
@@ -130,6 +130,28 @@ struct Argument {
      * Argument type.
      */
     2: required Type type
+    /**
+     * Annotations defined on this type.
+     *
+     * Note that these are the Thrift annotations listed after the type
+     * declaration in the Thrift file.
```
abhinav: Annotations defined on this argument.

    Note that these are listed after the parameter or exception name.

plugin/api.thrift
```diff
@@ -130,6 +130,28 @@ struct Argument {
...
+     *     1: BazError fail (bar = "foo")
+     *   )
+     *
+     * The annotations for the first argument will be,
+     *
+     *   {
+     *     "foo": "bar",
+     *     "validate": "",
+     *   }
+     */
```
abhinav: Also worth listing annotations on BazError.

plugin/api.thrift
```diff
@@ -130,6 +130,35 @@ struct Argument {
...
+     *   )
+     *
+     * The annotations for the first argument will be,
+     *
+     *   {
+     *     "foo": "bar",
+     *     "validate": "",
+     *   }
+     *
+     * and the annotations for BazError will be,
```
```suggestion
     * Similarly, the annotations for BazError will be,
```

(It's preferable to start a new sentence here rather than try to continue the previous after the code block.)

PR #404, opened 2019-02-11
Add support for exceptions not implementing error

mh-park:
This commit adds a `--no-error` flag, which compiles Thrift exceptions
without the error method. This effectively removes the distinction
between a struct and exception in compiled Go code.

Consequently, this changes how services are compiled as well, because
the helper functions relied on exceptions implementing error.

Specifically:
  - `IsException(error) bool` --> `IsException(interface{}) bool`
  - `WrapResponse(successType, error) (Result, error)`
     --> `WrapResponse(interface{}) (Result, error)`
    - This collapses the success and exceptions case into one argument,
      and the function handles type casting. This loses type safety for
      the success case. It also only returns an error if WrapResponse
      sees an unrecognized type.
  - `UnwrapResponse(Result) {error,(successType, error)}` 
     --> `UnwrapResponse(Result) ({success,},{interface{},},error)`
    - This adds an additional `exception` return value. Collapsing the
      exceptions case into one of the other return values was
      considered but ruled out, because it would be too cryptic and
      reliant on the user to type cast.


gen/service.go
```diff
@@ -156,14 +156,39 @@ func functionHelper(g Generator, s *compile.ServiceSpec, f *compile.FunctionSpec
...
 			Args func(<params $f>) *<$prefix>Args
 			<if not $f.OneWay>
+				<if (checkNoError) ->
+				// IsException returns true if the given value can be thrown
+				// by <$f.Name>.
+				//
+				// An exception can be thrown by <$f.Name> only if the
+				// corresponding exception type was mentioned in the 'throws'
+				// section for it in the Thrift file.
+				<- else ->
```
AlexAPB: optional: we can be more succinct
```
// An exception can be thrown by <$f.Name> if the type was mentioned in the 'throws' section of the IDL.
```

gen/service.go
```diff
@@ -156,14 +156,39 @@ func functionHelper(g Generator, s *compile.ServiceSpec, f *compile.FunctionSpec
...
+				<- end>
+				IsException func(<if (checkNoError)>interface{}<else>error<end>) bool
 				<if $f.ResultSpec.ReturnType>
+					<if (checkNoError) ->
+					// WrapResponse returns the result struct for <$f.Name>
+					// given its return value.
+					//
+					// This allows mapping values and exceptions returned by
+					// <$f.Name> into a serializable result struct.
+					// WrapResponse returns a non-nil error if the provided
```
AlexAPB: `returns an error`

gen/service.go
```diff
@@ -156,14 +156,39 @@ func functionHelper(g Generator, s *compile.ServiceSpec, f *compile.FunctionSpec
...
+					// value cannot be returned by <$f.Name>
```
AlexAPB: missing period.

gen/service.go
```diff
@@ -188,8 +223,34 @@ func functionHelper(g Generator, s *compile.ServiceSpec, f *compile.FunctionSpec
...
 				<else>
+					<if (checkNoError) ->
+					// WrapResponse returns the result struct for <$f.Name>
+					// given the value returned by it. The value should hold
+					// an exception that can be thrown by <$f.Name>.
+					//
+					// This allows mapping exceptions returned by <$f.Name> into a
+					// serializable result struct. WrapResponse returns a
+					// non-nil error if the provided error cannot be thrown by
+					// <$f.Name>
```
AlexAPB: nit period

gen/service.go
```diff
@@ -205,8 +266,19 @@ func functionHelper(g Generator, s *compile.ServiceSpec, f *compile.FunctionSpec
...
 
+					<if (checkNoError) ->
+					// UnwrapResponse takes the result struct for <$f.Name>
+					// and returns the exception returned by it (if any).
+					//
+					// The error is non-nil only if <$f.Name> returned an
+					// unrecognized value.
+					//
+					//   result := deserialize(bytes)
+					//   <if $f.ResultSpec.Exceptions>exception,<end> err := <$prefix>Helper.UnwrapResponse(result)
```
AlexAPB: should we have an error check here too?

gen/service.go
```diff
@@ -243,14 +318,15 @@ func functionHelper(g Generator, s *compile.ServiceSpec, f *compile.FunctionSpec
...
+		TemplateFunc("checkNoError", checkNoError),
 	)
 }
 
 // functionIsException generates an expression that provides the IsException
 // function for the given Thrift function.
 func functionIsException(g Generator, f *compile.FunctionSpec) (string, error) {
 	return g.TextTemplate(
-		`func(err error) bool {
+		`func(err <if (checkNoError)>interface{}<else>error<end>) bool {
```
AlexAPB: A thrift exception can be a struct or a pointer, so we should check that the exception is both a pointer _or_ non-pointer in the switch statement below.

In the generated code for `NoErrorService_GetValue_Helper` I'd expect something similar to:
```go
NoErrorService_GetValue_Helper.IsException = func(err interface{}) bool {
	switch err.(type) {
		case *NoErrorException, NoErrorException:
			return true
		default:
			return false
		}
	}
```

instead of just checking the pointer value.

twilly:
This looks fine to me as well. The option does break some plugins, so we should also think about extending the API to communicate options in GenerateServiceRequest so plugins can bomb out if needed.

abhinav:
You'll want to update the gen/quick_test to include the new types to bring the coverage back up.

abhinav:
```
=== RUN   TestCodeIsUpToDate
--- FAIL: TestCodeIsUpToDate (0.41s)
    golden_test.go:123: Generated code for "/home/travis/gopath/src/go.uber.org/thriftrw/gen/internal/tests/thrift/noerror.thrift" is out of date. Please run 'make' in gen/internal/tests.
```

Did you run `make generate`?

PR #405, opened 2019-02-15
Simplify import alias resolution

amckinney:
With `build.Import`, we depend on a variety of Go environment variables, such as `GOPATH` to determine the appropriate package alias to use in the generated output. Although possible to set the appropriate variables in the executing system, this change opts to simplify our import path resolution so that we can remove our dependency on these environment variables.

Rather than using a clever analysis of the package source, we simply opt to _always_ create an alias for all of our import path dependencies. This will prevent issues for packages with inconsistencies in their import path structure and their actual package declaration.

For example, consider an import path `foo/bar/baz` that actually has package `qux`. If we fail to resolve the package declaration with `build.Import`, we would previously generate `import "foo/bar/baz"` and reference it as `baz.<$type>`, which is invalid. This updates the behavior so that we always generate `import baz "foo/bar/baz"` and guarantee a safe import alias, regardless of the actual package declaration.


CHANGELOG.md
```diff
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- No changes yet.
+### Changed
+- Changed generated code to always include import aliases.
```
abhinav: nit:

    - Imports in generated code are now always named imports.

amckinney:
> plugin/ has its own templating and import system. That may be affected unless it's reusing this code. (I forget.)

Yup, that package imports from `goast`.

PR #406, opened 2019-02-21
Sanitize import aliases

amckinney:
Re: https://github.com/thriftrw/thriftrw-go/pull/405

The recent fix to always create an import alias did not use the same sanitization logic found in the `gen.importer.Import` method. This consolidates the `sanitizeImportName` into `goast.DeterminePackageName` so that both codepaths are correctly sanitized. This also update our test to include a `*.git` import path and notice that the `.` is appropriately transformed into a `_`.

### DIFF
--- a/gen/imports.go
+++ b/gen/imports.go
@@ -24,8 +24,8 @@ import (
 	"go/ast"
 	"go/token"
 	"path/filepath"
-	"strings"
-	"unicode"
+
+	"go.uber.org/thriftrw/internal/goast"
 )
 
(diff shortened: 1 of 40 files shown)


amckinney:
> Ah, number of changes to generated code makes more sense now.
> Please update the changelog.

I figured that I would leave the entry as is from the last commit since it wasn't released yet:
```
- Changed generated code to always include import aliases.
```

Thoughts?

abhinav:
I didn't realize the previous change hadn't been released yet. This is good as-is.

PR #407, opened 2019-02-22
gen/enum: Support periods in item names

ramprasadg:
This adds support for enum items with periods in their names.

Previously, the following enum would parse successfully but generate
invalid Go code.

    enum X { foo.bar, baz }

With this change, `foo.bar` is treated similarly to `foo_bar`.


abhinav:
Hello! Thanks for your contribution! Before we can review this, please,

- Sign the CLA
- Provide a descriptive PR title
- Provide a detailed PR description that explains *what* this change is doing and *why*; the *how* should be self-evident from the code

ramprasadgopal:
@abhinav can you please have a look now?
I have added the unit tests.

ramprasadg:
@abhinav its because RTAPI thrift files use them.
looks is  Android, iOS code gens unfortunately support this non standard version.

kriskowal:
@ramprasadg We’ve had many conversations over the years about RTAPI’s enums. The road for RTAPI IDL to be welcomed into the fold of IDL that can cross-compile is to move the dot delimited name into an annotation and change the name to an identifier that will compile in other languages. This is important especially since Eats services in particular use RTAPI to translate JSON to TBinary for Go services, so the IDL needs to compile for both.

I was under the impression that there was an effort in progress to convert enum names to annotations.

ramprasadg:
@kriskowal do you know who would be the right people to contact?
do you see a problem with this commit as an interim solution until the RTAPI effort is completed?

abhinav:
We decided against this after all.

PR #408, opened 2019-03-17
Fix function comments based on best practices from Effective Go

CodeLingoTeam:
Hi, we updated some exported function comments based on best practices from [Effective Go](https://golang.org/doc/effective_go.html). It’s admittedly a relatively minor fix up. Does this help you?


mh-park:
Typo in template code: https://github.com/thriftrw/thriftrw-go/blob/1c52f516bdc5ca90dc090ba2a8ee0bd11bf04f96/gen/map.go#L409

abhinav:
This change is pretty stale, and the fix would be to update the template. Closing.

PR #409, opened 2019-03-25
`func <T>Ptr(v <T>) *<T>` for primitive typedefs

twilly:
Using typedefs on a base type is a common thrift pattern used to
describe the kind of value that is within. While structs are trivial to
form a pointer to (`&Foo{"foo"}`), this is not the case with primitive
types (`type Jack string`, `&Jack("RJ45")`). This adds `<T>Ptr`
functions for primitive type aliases (`JackPtr("RJ45")`).

### DIFF
--- a/ast/basetypeid_string.go
+++ b/ast/basetypeid_string.go
@@ -24,6 +24,20 @@ package ast
 
 import "strconv"
 
+func _() {
+	// An "invalid array index" compiler error signifies that the constant values have changed.
+	// Re-run the stringer command to generate them again.
+	var x [1]struct{}
+	_ = x[BoolTypeID-1]
+	_ = x[I8TypeID-2]
+	_ = x[I16TypeID-3]
+	_ = x[I32TypeID-4]
+	_ = x[I64TypeID-5]
+	_ = x[DoubleTypeID-6]
+	_ = x[StringTypeID-7]
+	_ = x[BinaryTypeID-8]
+}
+
 const _BaseTypeID_name = "BoolTypeIDI8TypeIDI16TypeIDI32TypeIDI64TypeIDDoubleTypeIDStringTypeIDBinaryTypeID"
 
 var _BaseTypeID_index = [...]uint8{0, 10, 18, 27, 36, 45, 57, 69, 81}
--- a/gen/internal/tests/collision/types.go
+++ b/gen/internal/tests/collision/types.go
@@ -446,6 +446,11 @@ func (v *AccessorNoConflict) IsSetGetName() bool {
 
 type LittlePotatoe int64
 
+// LittlePotatoePtr returns a pointer to a LittlePotatoe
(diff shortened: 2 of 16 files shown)


abhinav:
Instead of a top-level function, it would be more consistent with the enum Ptr
methods for these to be methods too,

```
type UUID string

func (UUID) Ptr() *UUID
```

Usage would be similar: `UUID("...").Ptr()`.

Caveat: An advantage of the top-level function is that constant literals being
untyped, they are automatically converted to the argument type of the
function. That is, if `"foo"` is passed to `func UUIDPtr(UUID) *UUID`, it
automatically gets treated as `UUID("foo")`. With the method, that has to be
done as part of the Ptr call: `UUID("foo").Ptr()`, although it's not
significantly more characters.

I still think that `UUID("foo").Ptr()` is more consistent with what ThriftRW
already does, but I'm okay with `UUIDPtr` as a function if you feel strongly
about it.

@mattgode, what do you think?

mattgode:
> Instead of a top-level function, it would be more consistent with the enum Ptr
> methods for these to be methods too,
> 
> ```
> type UUID string
> 
> func (UUID) Ptr() *UUID
> ```
> Usage would be similar: `UUID("...").Ptr()`.
> 
> Caveat: An advantage of the top-level function is that constant literals being
> untyped, they are automatically converted to the argument type of the
> function. That is, if `"foo"` is passed to `func UUIDPtr(UUID) *UUID`, it
> automatically gets treated as `UUID("foo")`. With the method, that has to be
> done as part of the Ptr call: `UUID("foo").Ptr()`, although it's not
> significantly more characters.
> 
> I still think that `UUID("foo").Ptr()` is more consistent with what ThriftRW
> already does, but I'm okay with `UUIDPtr` as a function if you feel strongly
> about it.
> 
> @mattgode, what do you think?

The reason I prefer the `<Type>Ptr` format is for ease of migration from code generated with Apache Thrift generator to code generated with the ThriftRW generator. With the originally proposed addition, there is no change necessary other than an update to the import path. With this change, you'd need to manually update code or write a script to migrate. It's possible to write a script for this transformation, but it is _slightly_ tricky. As you noted, the go compiler infers the types of constants and so we have to handle two different transformations to support explicit and implicit types.

1. `uuid.UUIDPtr(uuid.UUID("some-uuid"))` -> `uuid.UUID("some-uuid").Ptr()`
2. `uuid.UUIDPtr("some-uuid")` -> `uuid.UUID("some-uuid").Ptr()`

With that being said, we have already done the enum pointer utilities as methods, so we need to handle the first transformation via a script. Given that handling the second transformation in a script is a one-off cost, I think @abhinav's suggestion makes sense.

twilly:
Agreed on all counts. I'll make this change more ThriftRW and less Apache Thrift.

twilly:
Derp, I accidentally deleted the branch. Reopening as a new PR. :(

PR #410, opened 2019-03-26
`func (v <T>) Ptr() *<T>` for primitive typedefs

twilly:
Using typedefs on a base type is a common thrift pattern used to
describe the kind of value that is within. While structs are trivial to
form a pointer to (`&Foo{"foo"}`), this is not the case with primitive
types (`type Jack string`, `&Jack("RJ45")`). This adds `<T>.Ptr()`
methods for primitive type aliases (`Jack("RJ45").Ptr()`).

### DIFF
--- a/gen/typedef.go
+++ b/gen/typedef.go
@@ -64,6 +64,14 @@ func typedef(g Generator, spec *compile.TypedefSpec) error {
 
 		<$v := newVar "v">
 		<$x := newVar "x">
+
+		<- if isPrimitiveType .>
+		// <typeName .>Ptr returns a pointer to a <$typedefType>
+		func (<$v> <typeName .>) Ptr() *<$typedefType>{
+			return &<$v>
+		}
+		<- end>
+
 		// ToWire translates <typeName .> into a Thrift-level intermediate
 		// representation. This intermediate representation may be serialized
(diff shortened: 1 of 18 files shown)


twilly:
@mattgode

gen/typedef.go
```diff
@@ -64,6 +64,14 @@ func typedef(g Generator, spec *compile.TypedefSpec) error {
 
 		<$v := newVar "v">
 		<$x := newVar "x">
+
+		<- if isPrimitiveType .>
+		// <typeName .>Ptr returns a pointer to a <$typedefType>
+		func (<$v> <typeName .>) Ptr() *<$typedefType>{
+			return &<$v>
```
abhinav: A test for this in quick_test would be nice.

You want to add another case [here](https://github.com/thriftrw/thriftrw-go/blob/ff6c015d46769061e36ceaa685adfe68b2c1afde/gen/quick_test.go#L514) for when the kind is `thriftTypedef`, and the object has a `Ptr` method, to do something similar to [testEnumPtr](https://github.com/thriftrw/thriftrw-go/blob/ff6c015d46769061e36ceaa685adfe68b2c1afde/gen/quick_test.go#L777).

This will make sure that the Ptr methods on all generated types get exercised.

PR #411, opened 2019-04-19
 Support thrift set to go slice generation with go.type annotation

ChuntaoLu:
This PR adds support to allow thriftrw to generate Go `slice` type from thrift `set` type when a `(go.type = "slice")` annotation is present. For example, with thrift definition 
```thrift
typedef set<string> (go.type = "slice") Foo

struct Bar {
    1: required set<string> (go.type = "slice") Quxx
}
```
thriftrw will now generate `Foo` and `Bar.Quxx` with type `[]string`, instead of `map[string]struct{}`.

The reason for this feature is that for languages that do not have native `set` type, different languages may have different types generated for the same thrift `set` type, e.g., the `Foo` field in the above example could be generated as `array` in JavaScript while it could be `map` in Go by default.


ChuntaoLu:
@abhinav Thanks for taking a look, that makes total sense. I have updated the PR with the approach you suggested, please take another look.

gen/container_test.go
```diff
@@ -686,7 +686,7 @@ func TestCrazyTown(t *testing.T) {
 				},
 			},
 			wire.NewValueStruct(wire.Struct{Fields: []wire.Field{
-				{ID: 4, Value: wire.NewValueSet(
+				{ID: 4, Value: wire.NewValueList(
```
ChuntaoLu: @abhinav This change is somewhat less related to my change, it should have been `NewValueList` instead of `NewValueSet` since type [here](https://github.com/thriftrw/thriftrw-go/blob/dev/gen/internal/tests/containers/types.go#L25) is slice, I think it is something overlooked.

abhinav: `wire.Value` is the encoded representation of objects in ThriftRW.
Regardless of the language-level representation used, over-the-wire we
should say that this is a set as defined [here][1].

(It's a slice at the language-level because `set<string>` isn't
hashable: `map[map[string]struct{}]struct{}` isn't a valid map in Go.)

So this should be NewValueSet.

[1]: https://github.com/thriftrw/thriftrw-go/blob/dev/gen/internal/tests/thrift/containers.thrift#L32

ChuntaoLu: I see, that makes sense. I have now updated diff to undo the change.

gen/gotype.go
```diff
@@ -0,0 +1,35 @@
...
+	// a typedef target type or a struct field type. By default, thrift set type
+	// corresponds to go map type. Following annotation on a typedef target type
+	// or struct field type, where the type is direct set type, causes thriftrw
+	// to generate go code with slice type instead of the default map type.
+	//
+	//     (go.type = "slice")
+	//
+	// Currently, only thrift set to go slice type overriding is supported.
+	goTypeKey = "go.type"
+	slice     = "slice"
```
abhinav: nit: change to "sliceType" please. The name "slice" is too generic for this shared namespace.

```suggestion
	sliceType = "slice"
```

gen/internal/tests/thrift/set_to_slice.thrift
```diff
@@ -0,0 +1,26 @@
...
+typedef StringList MyStringList
+typedef MyStringList AnotherStringList
+
+typedef set<set<string> (go.type = "slice")> (go.type = "slice") StringListLsit
+
+struct Foo {
+    1: required string stringField
+}
+
+struct Bar {
```
abhinav: This is a fairly thorough example. Thanks!

Would you mind adding these types to the `gen/quick_test.go` test suite? It'll exercise a lot of the standard code paths in these types. That'll help that little coverage drop as well.

gen/internal/tests/thrift/set_to_slice.thrift
```diff
@@ -0,0 +1,26 @@
+typedef set<string> (go.type = "slice") StringList
+typedef set<Foo> (go.type = "slice") FooList
+typedef StringList MyStringList
+typedef MyStringList AnotherStringList
+
+typedef set<set<string> (go.type = "slice")> (go.type = "slice") StringListLsit
```
abhinav: typo: StringListList?

```suggestion
typedef set<set<string> (go.type = "slice")> (go.type = "slice") StringListList
```

gen/type.go
```diff
@@ -49,6 +49,13 @@ func isHashable(t compile.TypeSpec) bool {
 	return isPrimitiveType(t)
 }
 
+// setUsesMap returns true if the given set type is not annotated with
+// (go.type = "slice") and the value of the set is considered hashable
+// by thriftrw.
+func setUsesMap(spec *compile.SetSpec) bool {
+	return (spec.Annotations[goTypeKey] != slice) && isHashable(spec.ValueSpec)
```
abhinav:
```suggestion
	return (spec.Annotations[goTypeKey] != sliceType) && isHashable(spec.ValueSpec)
```

gen/internal/tests/thrift/set_to_slice.thrift
ChuntaoLu: Done

gen/quick_test.go
```diff
@@ -392,6 +398,11 @@ func TestQuickSuite(t *testing.T) {
 		{Sample: tz.StringMap{}, NoLog: true, Kind: thriftTypedef},
 		{Sample: tz.Primitives{}, NoLog: true, Kind: thriftTypedef},
 		{Sample: tz.StringList{}, NoLog: true, Kind: thriftTypedef},
+		{Sample: tss.StringList{}, NoLog: true, Kind: thriftTypedef},
+		{Sample: tss.FooList{}, NoLog: true, Kind: thriftTypedef},
+		{Sample: tss.MyStringList{}, NoLog: true, Kind: thriftTypedef},
+		{Sample: tss.AnotherStringList{}, NoLog: true, Kind: thriftTypedef},
+		{Sample: tss.StringListList{}, NoLog: true, Kind: thriftTypedef},
```
abhinav: I suspect you don't need the NoLog entries. Those are needed for typedefs of primitives only.

(The `tz.*` types have NoLog because they are testing the zap opt-out flag.)

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- No changes yet.
+- Add `(go.type = "slice")` annotation to allow generate thrift `set` type as go `slice` type.
```
abhinav:
```suggestion
- Sets now support a `(go.type = "slice")` annotation to be generated as
  slices rather than maps.
```

abhinav:
Thanks @ChuntaoLu!

PR #414, opened 2019-05-16
gen/set: Fix variance in go.type

abhinav:
With support for `go.type = "slice"` for sets, we introduced a bug where
if `set<T>` occurs twice in the same Thrift file, once with `go.type`
set to `slice`, and the other a `map`, we would generate code that does
not compile.

The reason for this is that the mangled name we use for `set<T>` is
based just on the type name. The mangled name is used to generate
helpers for this type, so only one set of helpers gets generated despite
two internal representations of `set<T>`.

For example, for `set<string>`, despite using both `[]string` and
`map[string]struct{}`, we will only generate one of the following,

    type _Set_String_ValueList []string
    type _Set_String_ValueList map[string]struct{}

This changes type mangling to include the `go.type` of sets in the
mangled name so that instead we will now generate,

    type _Set_String_sliceType_ValueList []string
    type _Set_String_mapType_ValueList map[string]struct{}


abhinav:
@ChuntaoLu would you mind taking a look as well?

PR #416, opened 2019-05-20
Generate single file for ThriftRW output

rhang-uber:
ThriftRW (excluding plugin outputs) now only generates a single file
that is lower-cased named after the value specified
by the added single-file flag.

If the single-file flag is not specified, the name of the generated
file will be named after the thrift file that ThriftRW is invoked upon.

This file is generated in the directory created by ThriftRW, which
is named after the thrift file that thriftRW is invoked on.


gen/service.go
```diff
@@ -23,10 +23,9 @@ package gen
 import (
 	"bytes"
 	"fmt"
+	"go.uber.org/thriftrw/compile"
```
mh-park: nit: imports formatting here should look like previous

gen/generate.go
```diff
@@ -284,45 +292,69 @@ func generateModule(m *compile.Module, i thriftPackageImporter, builder *generat
...
+						files[name] = buff.Bytes()
+					}
+				}
+			}
 		}
+	} else if len(o.SingleFile) > 0 {
+		buff := new(bytes.Buffer)
+		if err := g.Write(buff, nil); err != nil {
+			return nil, fmt.Errorf(
+				"could not write non-service code when no services were found, but the single-service flag was specified: %v", err)
```
mh-park: I don't fully understand this error output. The "single-service flag" part doesn't make sense; from what I understand, the flag only specifies that everything gets written into a single file, but I see some error output about services, so it's confusing.

gen/generate.go
```diff
@@ -284,45 +292,69 @@ func generateModule(m *compile.Module, i thriftPackageImporter, builder *generat
...
 			if err != nil {
 				return nil, fmt.Errorf(
-					"could not generate code for service %q: %v",
-					serviceName, err)
+					"could not generate code for all services in the thrift file: %v", err)
 			}
 
 			if !o.NoServiceHelpers {
 				for name, buff := range serviceFiles {
 					files[name] = buff.Bytes()
```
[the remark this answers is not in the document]
mh-park: Just kidding, it looks like they're in different logic groups, so you won't hit this and line 350-357. Either way, it's a little confusing and duplicative, and we should be only calling `files[o.SingleFile] = buff.Bytes()` once. (This one is calling `files[o.SingleFile] = buff.Bytes()` indirectly through the `Services(...)` call above.)

gen/generate.go
```diff
@@ -284,45 +292,69 @@ func generateModule(m *compile.Module, i thriftPackageImporter, builder *generat
...
-			// considered root services; plugins will generate code only for
-			// root services, even though they have information about the
-			// whole service tree.
-			if _, err := builder.AddRootService(service); err != nil {
-				return nil, err
-			}
-
-			serviceFiles, err := Service(g, service)
+		if len(o.SingleFile) > 0 {
+			serviceFiles, err := Services(g, m.Services, o.SingleFile)
```
mh-park: From digging a little bit, it looks like we're actually smuggling a bunch of file output into the generator ( when we call `Constant(..)`/`TypeDefinition(..)`/`embedIDL(..)`/ etc. ), and then in this call to `Services(..)`, we suddenly write to a buffer, which writes everything held in the generator into this single file. 

Rather than doing this, `Services(..)` should just make the generator hold a bunch of output, and then write it into a single buffer at then end. Let me know if you need some clarification.

mh-park:
Please add a more detailed title!

gen/service.go
rhang-uber: > Please add a more detailed title!

Apologies! I mixed up the location box of the branch name and the title name! The commit message title has been moved to the correct place in the github UI!

gen/generate.go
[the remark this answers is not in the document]
rhang-uber: I've changed to code to follow similar patterns in gen/generate, where the code is generated and stored in the generator and written out to the file buffer map as a separate step!

gen/generate.go
rhang-uber: Agreed, I've changed the error message. The original motivation behind the error message was to give context to my worry about the case in which the single-file flag was specified, but because there were no services defined in the thrift file no file would be written. In such as case, a file should always be written. I've now separated the logic to more naturally express that a file should be written if the single-file flag is specified, and therefore the error message can be changed to something more understandable.

gen/generate.go
rhang-uber: Agreed that this is more consistent with the rest of the code in gen/generator.go. I originally implemented my Services logic like this since generating and then writing the code is a pattern of the Service function, which was functionality that I originally tried to emulate.

Breaking out the code templating/generation step and the file writing step does allow me to
share more code in gen/generate.go so I've separated the logic out.

gen/generate.go
```diff
@@ -76,6 +76,9 @@ type Options struct {
 
 	// Do not generate Zap logging code
 	NoZap bool
+
+	// Name of the file to be generated by thriftRW.
+	SingleFile string
```
abhinav: At this point, seems like this should be "OutputFile"?

main.go
```diff
@@ -52,13 +52,14 @@ type genOptions struct {
...
-	NoEmbedIDL        bool `long:"no-embed-idl" description:"Do not embed IDLs into the generated code."`
-	NoZap             bool `long:"no-zap" description:"Do not generate code for Zap logging."`
+	GeneratePluginAPI bool   `long:"generate-plugin-api" hidden:"true" description:"Generates code for the plugin API"`
+	NoVersionCheck    bool   `long:"no-version-check" hidden:"true" description:"Does not add library version checks to generated code."`
+	NoTypes           bool   `long:"no-types" description:"Do not generate code for types, implies --no-service-helpers."`
+	NoConstants       bool   `long:"no-constants" description:"Do not generate code for const declarations."`
+	NoServiceHelpers  bool   `long:"no-service-helpers" description:"Do not generate service helpers."`
+	NoEmbedIDL        bool   `long:"no-embed-idl" description:"Do not embed IDLs into the generated code."`
+	NoZap             bool   `long:"no-zap" description:"Do not generate code for Zap logging."`
+	SingleFile        string `long:"single-file" value-name:"FILENAME" description:"Generates a single .go file as an output."`
```
abhinav: OutputFile?

Also, we'll need some sort of check somewhere to ensure that you haven't provided a filename if multiple Thrift files are involved because the name won't apply to included files.

main.go
```diff
@@ -155,6 +156,10 @@ func do() (err error) {
 		}
 	}
 
+	if len(gopts.SingleFile) > 0 && filepath.Ext(gopts.SingleFile) != ".go" {
+		return fmt.Errorf("Name: %v invalid. A {FILENAME}.go name must be provided", gopts.SingleFile)
```
abhinav: name %q is invalid

Error messages must begin with lower case (https://github.com/golang/go/wiki/CodeReviewComments#error-strings) and you want to use `%q` to have it quote things like empty strings.

gen/generate.go
```diff
@@ -309,22 +282,29 @@ func generateModule(m *compile.Module, i thriftPackageImporter, builder *generat
...
-					serviceName, err)
-			}
-
-			if !o.NoServiceHelpers {
-				for name, buff := range serviceFiles {
-					files[name] = buff.Bytes()
-				}
-			}
+		err = Services(g, m.Services)
+		if err != nil {
```
abhinav: can inline error into if statement

```
if err := Services(..); err != nil {
  ..
}
```

Reduces the scope of the variable

gen/generate.go
```diff
@@ -309,22 +282,29 @@ func generateModule(m *compile.Module, i thriftPackageImporter, builder *generat
...
+			return nil, fmt.Errorf(
+				"could not generate code for services %v", err)
 		}
 	}
 
+	buff := new(bytes.Buffer)
+	if err := g.Write(buff, nil); err != nil {
+		return nil, fmt.Errorf(
+			"could not write code for services into %q: %v", o.SingleFile, err)
```
abhinav: code is not just for services in this case. this should just be `"could not write file %q: %v"`

gen/generate.go
```diff
@@ -309,22 +282,29 @@ func generateModule(m *compile.Module, i thriftPackageImporter, builder *generat
...
+	}
+
+	// The default is the package name
+	name := packageName + ".go"
+	if len(o.SingleFile) > 0 {
+		name = o.SingleFile
```
abhinav: Should decide on the output file name before including it in the error message above.

gen/generate_test.go
```diff
@@ -112,27 +120,29 @@ func TestGenerate(t *testing.T) {
...
 	}
 
 	tests := []struct {
-		desc      string
-		noRecurse bool
-		getPlugin func(*gomock.Controller) plugin.Handle
+		desc       string
+		noRecurse  bool
+		getPlugin  func(*gomock.Controller) plugin.Handle
+		singleFile string
```
abhinav: outputFile

gen/generate_test.go
```diff
@@ -153,8 +162,22 @@ func TestGenerate(t *testing.T) {
...
+		{
+			desc: "single file output",
+			getPlugin: func(mockCtrl *gomock.Controller) plugin.Handle {
+				return plugin.EmptyHandle
+			},
+			singleFile: "services.go",
+			notWantFiles: []string{
 				"foo/types.go",
 				"common/bar/types.go",
+				"common/bar/serivces.go",
```
abhinav: typo: services

gen/generate_test.go
```diff
@@ -112,27 +120,29 @@ func TestGenerate(t *testing.T) {
...
-		noRecurse bool
-		getPlugin func(*gomock.Controller) plugin.Handle
+		desc       string
+		noRecurse  bool
+		getPlugin  func(*gomock.Controller) plugin.Handle
+		singleFile string
 
-		wantFiles []string
-		wantError string
+		notWantFiles []string
```
abhinav: given that this is the only behavior now, notWantFiles seems unnecessary

gen/namespace.go
```diff
@@ -112,7 +112,7 @@ func (n *namespace) Rotate(base string) (string, error) {
 }
 
 func (n *namespace) Reserve(name string) error {
-	if n.isTaken(name) {
+	if n.isTaken(name) && (name != "init") {
```
abhinav: I think the init special case should stand out and be explained.

```suggestion
	// A single Go file can have multiple init() functions so we don't need to
	// reserve it.
	if name == "init" {
		return nil
	}

	if n.isTaken(name) {
```

gen/service.go
```diff
@@ -58,6 +58,27 @@ func Service(g Generator, s *compile.ServiceSpec) (map[string]*bytes.Buffer, err
 	return files, nil
 }
 
+// Services generates code for all services into a single file and stores the code
+// in the generator to be written.
+//
+// Returns a map from file name to contents for that file. The file names are
+// relative to the package directory for the service.
+func Services(g Generator, services map[string]*compile.ServiceSpec) error {
```
abhinav: Docs are outdated. "single file" is always true and a map isn't returned.

gen/generate.go
rhang-uber: Changed the name.

main.go
rhang-uber: Changed the name. I don't believe this check is needed. From my reading of the code I don't think multiple thrift files can be specified in an input.

abhinav: Multiple Thrift files can't be specified as input but without the `--no-recurse` flag, if a file includes another Thrift file, it will be generated too.

rhang-uber: If this behaviour is not allowed I think it would be safe to mandate --no-recurse if the flag is specified. It seems like restricting and documenting the behaviour is safer, what are your thoughts?

abhinav: Yeah, I'm okay with saying "--output-file" implies "--no-recurse"

rhang-uber: Okay, i've made the change.

gen/generate.go
```diff
@@ -76,6 +76,9 @@ type Options struct {
 
 	// Do not generate Zap logging code
 	NoZap bool
+
+	// Name of the file to be generated by thriftRW.
```
abhinav:
```suggestion
	// Name of the file to be generated by ThriftRW.
```

gen/generate.go
```diff
@@ -309,22 +285,28 @@ func generateModule(m *compile.Module, i thriftPackageImporter, builder *generat
...
+	buff := new(bytes.Buffer)
+	if err := g.Write(buff, nil); err != nil {
+		return nil, fmt.Errorf(
+			"could not write file %q: %v", o.OutputFile, err)
+	}
+
+	// The default is the package name
+	name := packageName + ".go"
+	if len(o.OutputFile) > 0 {
+		name = o.OutputFile
```
abhinav: Name must be calculated before the error message thrown on line 299.

rhang-uber: Adjusted

gen/namespace.go
rhang-uber: Adjusted.

gen/generate.go
```diff
@@ -200,72 +202,83 @@ func (i thriftPackageImporter) Package(file string) (string, error) {
...
+		return "", nil, err
 	}
 
-	// Mapping of file names relative to packageRelPath to their contents.
-	// Note that we need to return a mapping relative to o.OutputDir so we
-	// will prepend $packageRelPath/ to all these paths.
-	files := make(map[string][]byte)
-
-	gopts := &GeneratorOptions{
+	g := NewGenerator(&GeneratorOptions{
```
rhang-uber: Thank you for cleaning up the code! It looks much cleaner with the Options object declared in the NewGenerator called!

PR #417, opened 2019-06-03
Always add module to GenerateServiceResponse

rhang-uber:
Adds Thrift module data to the GenerateServiceResponse even if
the module provided to the GenerateServiceRequest contains no Thrift
Service information. This ensures that ThrifRW plugins always receive
Thirft module data even if a Thrift file has no service declarations.

### DIFF
--- a/gen/generate.go
+++ b/gen/generate.go
@@ -282,6 +282,15 @@ func generateModule(
 		}
 	}
 
+	addModules := func(m *compile.Module) error {
+		_, err := builder.AddModule(m.ThriftPath)
+		return err
+	}
+
+	if err := m.Walk(addModules); err != nil {
+		return "", nil, err
+	}
+
 	// Services must be generated last because names of user-defined types take
 	// precedence over the names we pick for the service types.
 	if len(m.Services) > 0 {
--- a/gen/plugin.go
+++ b/gen/plugin.go
@@ -85,8 +85,8 @@ func (g *generateServiceBuilder) AddRootService(spec *compile.ServiceSpec) (api.
 	return id, err
 }
 
-// addModule adds the module for the given Thrift file to the request.
-func (g *generateServiceBuilder) addModule(thriftPath string) (api.ModuleID, error) {
+// AddModule adds the module for the given Thrift file to the request.
+func (g *generateServiceBuilder) AddModule(thriftPath string) (api.ModuleID, error) {
 	if id, ok := g.moduleIDs[thriftPath]; ok {
 		return id, nil
 	}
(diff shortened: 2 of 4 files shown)


gen/generate.go
```diff
@@ -303,6 +303,11 @@ func generateModule(
 		if err = Services(g, m.Services); err != nil {
 			return "", nil, fmt.Errorf("could not generate code for services %v", err)
 		}
+	} else {
+		// Add module even if the Thrift file does not contain a service.
+		if _, err := builder.AddModule(m.ThriftPath); err != nil {
```
abhinav: Two completely divergent code paths here doesn't seem right. There's very little in common between the if-else branches.

We currently have.

```
if services { addServices() }
else { addModules() }
```

Where `addServices()` at some point does all or part of the work done by `addModules()`.

Perhaps we should do,

```
addModules()
if services { addServices() }
```

Less divergence is cleaner, easier to reason about, and easier to test.

And if you make `addModules` add *all* modules: the current module and all its included modules, then AddService just needs to look up module IDs and never actually add them.

([compile/Module.Walk][1] will help in adding a module, its included modules, and their included modules.)

[1]: https://godoc.org/go.uber.org/thriftrw/compile#Module.Walk

rhang-uber: I've updated the logic to reflect these suggestions. I also want to flag a change I made in the test cases, which i'll do inline there.

gen/plugin_test.go
```diff
@@ -169,6 +169,12 @@ func TestAddRootService(t *testing.T) {
 		}
 
 		g := newGenerateServiceBuilder(importer)
+
+		if spec.Parent != nil {
```
rhang-uber: @abhinav, I wonder if you feel this testing logic is sufficient. Here i'm manually replicating the logic by which addService adds the modules to the GenerateServiceBuilder. I'm particular, I'm ordering the addition of the module of parent first as the addService builder followed this logic before I moved the the AddModule logic out of AddService.

gen/generate.go
```diff
@@ -282,6 +282,17 @@ func generateModule(
 		}
 	}
 
+	addModules := func(m *compile.Module) error {
+		if _, err := builder.AddModule(m.ThriftPath); err != nil {
+			return err
+		}
+		return nil
```
abhinav: nit: Can shorten to,

```
_, err := builder.AddModule(m.ThriftPath)
return err
```

gen/plugin.go
```diff
@@ -135,9 +135,10 @@ func (g *generateServiceBuilder) addService(spec *compile.ServiceSpec) (api.Serv
 	g.nextServiceID++
 	g.serviceIDs[spec.ThriftFile()][serviceName(spec.Name)] = serviceID
 
-	moduleID, err := g.addModule(spec.ThriftFile())
-	if err != nil {
-		return 0, err
+	// Modules must already be populated.
+	moduleID, ok := g.moduleIDs[spec.ThriftFile()]
+	if !ok {
+		return 0, fmt.Errorf("unable to lookup moduleID for thrift file: %s", spec.ThriftFile())
```
abhinav: nit:

```suggestion
		return 0, fmt.Errorf("unable to lookup module ID for Thrift file %q", spec.ThriftFile())
```

gen/plugin_test.go
abhinav: This is fine.  Optionally, you can just mutate the Modules map directly.  The
contract from AddRootService's point of view is that every Thrift file has a
ModuleID entry.

At some point it would make sense to break apart the logic to fill
GenerateServiceRequests into more separate units rather than methods on the
same object.

rhang-uber: Right, I didn't mutate the modules map directly in order to exercise the functionality of the AddModules method. 👍 on the break apart logic step.

PR #420, opened 2019-07-18
compile/service: fix nil deref for exception

argouber:
If you attempt to use a type that isn't user defined as an exception,
the system panics instead of returning an error.

This change fixes the issue and adds a failing test case.

### DIFF
--- a/compile/service.go
+++ b/compile/service.go
@@ -313,7 +313,12 @@ func (rs *ResultSpec) Link(scope Scope) (err error) {
 	// verify that everything listed under throws is an exception.
 	for _, exception := range rs.Exceptions {
 		spec, ok := exception.Type.(*StructSpec)
-		if !ok || spec.Type != ast.ExceptionType {
+		if !ok {
+			return notAnExceptionError{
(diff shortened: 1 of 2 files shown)


argouber:
@abhinav mind taking a look

compile/service.go
```diff
@@ -313,7 +313,12 @@ func (rs *ResultSpec) Link(scope Scope) (err error) {
 	// verify that everything listed under throws is an exception.
 	for _, exception := range rs.Exceptions {
 		spec, ok := exception.Type.(*StructSpec)
-		if !ok || spec.Type != ast.ExceptionType {
+		if !ok {
+			return notAnExceptionError{
+				FieldName: exception.ThriftName(),
+			}
+		}
+		if spec.Type != ast.ExceptionType {
```
abhinav: I think you can collapse the original if statement back but use
`exception.Type.ThriftName()` instead of `spec.ThriftName()`.

argouber:
Updated. @abhinav

abhinav:
Thanks!

PR #421, opened 2019-07-19
Changed field compilation to be case-sensitive.

alancastrouber:
This behavior is consistent with thriftrw-node as well as Apache thrift.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- No changes yet.
+### Fixed
+- Fixed field compilation to allow fields with similar looking names and
+  different casing.
 
 ## [1.20.0] - 2019-06-12
 ### Changed
--- a/compile/field.go
+++ b/compile/field.go
@@ -175,7 +175,7 @@ type FieldGroup []*FieldSpec
 
 // compileFields compiles a collection of AST fields into a FieldGroup.
 func compileFields(src []*ast.Field, options fieldOptions) (FieldGroup, error) {
-	fieldsNS := newNamespace(caseInsensitive)
+	fieldsNS := newNamespace(caseSensitive)
 	usedIDs := make(map[int16]string)
 
 	fields := make([]*FieldSpec, 0, len(src))
--- a/compile/struct_test.go
+++ b/compile/struct_test.go
@@ -90,6 +90,35 @@ func TestCompileStructSuccess(t *testing.T) {
 			},
 		},
 		{
+			`struct Foo {
+				1: string island
+				2: string isLand
+			}`,
+			nil,
+			defaultToOptional,
+			&StructSpec{
+				Name: "Foo",
+				File: "test.thrift",
+				Type: ast.StructType,
+				Fields: FieldGroup{
+					{
+						ID:       1,
+						Name:     "island",
+						Type:     &StringSpec{},
+						Required: false,
+						Default:  nil,
+					},
+					{
+						ID:       2,
+						Name:     "isLand",
+						Type:     &StringSpec{},
+						Required: false,
+						Default:  nil,
+					},
+				},
+			},
+		},
+		{
 			`exception KeyNotFoundError {
 				1: required string message
 				2: optional Key key


rliu054:
@abhinav Would you mind taking a look? Thanks!

rhang-uber:
Hey @alancastrouber, as an update to this issue: cases were intentionally left fields case insensitive in order to discourage similar names. ThriftRW intentionally breaks from Apache Thrift conventions in a couple different ways and this is one of those differences.

Things that we are worried about in this change are:
(1) Thrift fields firstName and FirstName will be the same in the generated Go code, which won't compile since all fields must be exported.
(2) There is logic in ThriftRW where First_Name also becomes FirstName, so if you have first_name, firstName or First_name, they'll all conflict in the generated code.

Case insensitivity was made to guard against these issues. The theory behind this change is good, but more testing around the cases above is needed to merge the code in.

Though there could be a workaround for your issue: (Excuse us for the poor documentation), but in the Changelog (1.13.0 - 2018-09-10), go. annotations are exposed which allows the user to override the fields of the generated code. Take a look at go.name and go.label annotations, these will allow you to change the generated field names.

alancastrouber:
@rhang-uber, @mh-park,

Thanks for taking a look at this!
Please see [here](https://code.uberinternal.com/T3475825) if you would like a bit more context on this issue.

---

> Take a look at go.name and go.label annotations, these will allow you to change the generated field names.

I don't believe this solution works. Consider the following example:

```
struct FooStruct {
    1: optional string island
    2: optional string isLand (go.label="isGround", go.tag='json:"isGround"', go.name="isGround")
}

```

Attempting to pass this through thriftrw's latest dev branch still results in `cannot compile "isLand" on line 3: the name "isLand" has already been used on line 2`.

I believe that the case-insensitive restriction occurs within compilation before any annotations are even processed and before code-generation is even attempted.

@rhang-uber, Can you confirm this? 

---

>(1) Thrift fields firstName and FirstName will be the same in the generated Go code, which won't compile since all fields must be exported.

I don't think this is a problem at all. Consider the following example. 

```
struct FooStruct {
    1: optional string firstName
    2: optional string FirstName
}
```

Currently on the dev branch, this fails with `Failed to compile... cannot compile "FirstName" on line 3: the name "FirstName" has already been used on line 2`. This is failing due to the **artificial restriction** during compilation that field names should be (case-insensitive) unique.

After removing that restriction (with this pull request), the example above fails with, `Failed to generate code... could not declare field "FirstName": name "FirstName" is already taken`

Either way, the example fails to generate valid go-code. The only thing that changes is whether it fails during compilation, or during code-generation. I would argue that failing during code-generation is more accurate than failing during compilation. Enforcing case-insensitive uniqueness is trying to anticipate the code-generation collision prematurely. 

yes, `firstName`  and `FirstName` will both try to be exported as `FirstName` and result in a collision. But I don't think that trying to anticipate that collision by enforcing case-insensitive uniqueness is the right approach, since it is more limiting than necessary. We should let it arrive at the collision, and fail there.

In the `island`, and `isLand` example, they would be exported as `Island` and `IsLand`, which would result in no collision whatsoever, and code-generation would work just fine. Enforcing case-insensitive uniqueness is overly restrictive here for no reason.

> (2) There is logic in ThriftRW where First_Name also becomes FirstName, so if you have first_name, firstName or First_name, they'll all conflict in the generated code.

Same thing as above. In the case of true collisions, then code-generation will fail either way. We should let it conflict in the generated code and fail there. We should not try to anticipate the collision by enforcing (case-insensitive) uniqueness. 

---

> fields were left case insensitive in order to discourage similar names.

I think this means well, but there are many examples where snake-casing carries alot of information to allow for different meanings. `together` means joined, where-as `ToGetHer` means something else entirely.

---

> Name:     "isLand", // previously "island"

@mh-park, Could you double-check this? I don't think your example is correct.

On the dev branch,

```
struct Foo {
	1: optional string isLand
}
```

generates

```
&StructSpec{
	Name: "Foo",
	File: "test.thrift",
	Type: ast.StructType,
	Fields: FieldGroup{
		{
			ID:       1,
			Name:     "isLand",
			Type:     &StringSpec{},
			Required: false,
			Default:  nil,
		},
	},
},
```

This remains the same even after the change in my pull-request.

---

> This would be a breaking change:

Kind of.

**Any thrift file that generated go-code before should still generate the same go-code.**

But things that would have failed earlier (isLand vs island) will now **also** be able to generate go-code.

---

Please let me know your thoughts.

compile/field.go
```diff
@@ -175,7 +175,7 @@ type FieldGroup []*FieldSpec
 
 // compileFields compiles a collection of AST fields into a FieldGroup.
 func compileFields(src []*ast.Field, options fieldOptions) (FieldGroup, error) {
-	fieldsNS := newNamespace(caseInsensitive)
+	fieldsNS := newNamespace(caseSensitive)
```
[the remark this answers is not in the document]
alancastrouber: > Name:     "isLand", // previously "island"

@mh-park, Could you double-check this? I don't think your example is correct.

On the dev branch,

```
struct Foo {
	1: optional string isLand
}
```

generates

```
&StructSpec{
	Name: "Foo",
	File: "test.thrift",
	Type: ast.StructType,
	Fields: FieldGroup{
		{
			ID:       1,
			Name:     "isLand",
			Type:     &StringSpec{},
			Required: false,
			Default:  nil,
		},
	},
},
```

This remains the same even after the change in my pull-request.

abhinav:
Thanks for the PR @alancastrouber! I can buy the argument that collision
failures should be deferred to code generation time. ThriftRW works with the
mental model that compilation interprets the Thrift file agnostic of the
target language, and that code generation deals with language-specific issues.
This is one of the reasons why we've avoided leaking Go-level details into the
compilation step.

Seeing as the case-insensitive check is to guard against a Go issue, not a
Thrift issue, I think this change makes sense.

To add some history to this: the check was added before the `go.name` or
`go.label` annotations existed. There was no way to guard against issues like
this except to ensure that users didn't craft invalid Thrift files (from Go's
point of view).

With these annotations, it's possible for the Thrift and Go-level field names
to diverge, so this check can be removed. If a Go-level collision occurs,
users can use `go.name` or `go.label` to fix the issue in the generated Go
code.

---

RE: Breaking change: This is not a breaking change because it's going from
restrictive to less restrictive. Existing customers won't be broken, and code
that previously failed to work will start working.

rhang-uber:
Works for me, would it make sense to add a test case for "breaking-ness" discussion even though it works out in theory?

alancastrouber:
Thanks again for taking a look, everyone! I appreciate learning about some of the history - it totally make sense :)

> would you mind adding an entry to the changelog

Sure thing, I updated the pull request.

> would it make sense to add a test case for "breaking-ness" discussion 

I'm not sure how I would add a unit test to irrefutably demonstrate that this is not a breaking change. However, the fact that all of the existing tests continue to work without modification gives me some confidence.

Is there something you had in mind?

rhang-uber:
@alancastrouber, Minho raised a possible breaking case above. However, seeing as this case is already covered as you pointed out:

> EDIT: See Here for an already existing test that demonstrates this. Notice that lines 133 and 139 are plainText and richText (rather than plaintext and richtext

Nothing else needs to be added.

alancastrouber:
I apologize if this has been answered before, but what does the release process look like? I'm hoping to update a depending service to use the changes from this pull-request. Please let me know what I can expect in terms of timeline

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 - No changes yet.
 
+## [1.21.0] - 2019-07-23
+### Fixed
+- Field compilation is now case-sensitive.
```
abhinav: This should reflect how it affects the user.

```suggestion
- Fixed field compilation to allow fields with similar looking names and
  different casing.
```

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 - No changes yet.
 
+## [1.21.0] - 2019-07-23
```
abhinav: The new entries will go under the Unreleased section above.

(Also, seeing as it's a bugfix, when it gets cut, it'll be a 1.20.1
unless there's also a feature in here somewhere.)

alancastrouber: > The new entries will go under the Unreleased section above

Thanks! I updated it. 👍

> when it gets cut, it'll be a 1.20.1

What can I expect in terms of timeline?

CHANGELOG.md
alancastrouber: 👍

alancastrouber:
Hello everyone, what are the next steps towards release?

abhinav:
Oops, sorry I realize I never posted here about the next steps. From
the conversation with Alan:

Someone will try to release this on Monday (we don't do releases on
Fridays) if they have time. Otherwise, I'll be able to do it by
Wednesday.

abhinav:
Thanks for the PR!

rhang-uber:
@alancastrouber, I just released the latest version of Thriftrw-go, it should be v1.20.1!

PR #423, opened 2019-10-09
Hyphenated file + include-as support

abhishekparwal:
added support for the Hyphenated files along with include-as directive support for including these files. Normal including is against thrift grammar for identifier with a hyphen, thus leading thrift file useable as top-level thrift file as well as included thrift file.

Details:
Go Code generated would be having a package name as abc_def, i.e hyphen converted to underscore for working around golang casing constraints.

include-as is similar to golang import alias, i.e.:
```
include t "abc-def.thrift"
t.XYZ
```

Hyphenated files normally can not be included as identifier of that form are not allowed in thrift grammar. This is because of hyphens like `abc-def.XYZ`. However, if the include-as directive is used, even with hyphen names we can use like `include t "abc-def.thrift" t.XYZ`. Since include-as was not supported previously added support for it as now even these thrift files are usable other than top level files.


abhinav:
Closing as discussed

abhishekparwal:
> We cannot introduce custom syntax.

however, as @abhinav said, Hyphenated files can not be included as an identifier of that form are not allowed in thrift grammar.
Not adding support for include-as directive as this is a custom thrift syntax which is not officially supported and leaked into thriftrw grammar earlier. though, it tried adding support for it, which we are not landing now and closing it.

Issue #143, opened 2016-05-27
Dashes in Thrift file names


abhinav:
We need to replace dashes in file names with underscores.

```
$ cat > hello-world.thrift
struct Hello { 1: required string world }
$ thriftrw-go hello-world.thrift
$ head hello-world/types.go
// Code generated by thriftrw

package hello-world
[..]
```

Also, if we add a service, `service Foo { Hello bar() }`,

```
$ thriftrw-go hello-world.thrift
2016/05/27 15:57:28 failed to generate code for "[..]/hello-world.thrift": could not generate code for service "Foo": could not generate types for Foo.bar: failed to generate code for "Foo.bar": could not parse generated code: thriftrw.go:6:20: expected ';', found '-' (and 1 more errors):
package thriftrw

type BarResult struct {


					Success *hello-world.Hello `json:"success,omitempty"`


		}
```

abhinav:
A similar issue applies to Thrift files named after Go keywords: for, return, switch, type, etc.

abhinav:
Follow up:
Apache Thrift does not support this either. In fact, files with dashes in
their names cannot be included by other Thrift files.

    include "foo-bar.thrift"

The Thrift IDL format does not provide a means of referencing identifiers
defined in foo-bar.thrift. It does not get renamed to `foo_bar`.

It still makes sense for us to have a graceful degradation when generating
code for top-level Thrift files that aren't included by other files, so it
should be okay to make the code-generator more accepting without changing the
parsing behavior.

abhishekparwal:
direct include is a limitation of thrift grammar which doesn't allow hyphenated names to be used as an identifier, like `abc-def.XYZ xyz`is disallowed. however we can do include-as to even include these files, which renders the use of these files beyond top level.

#423 adds support for the Hyphenated file as both top level thrift files as well as included thrift file.

include-as is similar to golang import alias, i.e.:
```
include t "abc-def.thrift"
t.XYZ
```
Hyphenated files normally can not be included as identifier of that form are not allowed in thrift grammar. This is because of hyphens like abc-def.XYZ. However, if the include-as directive is used, even with hyphen names we can use like include t "abc-def.thrift" t.XYZ. Since include-as was not supported previously added support for it as now even these thrift files are usable other than top level files.

> Follow up:
> Apache Thrift does not support this either. In fact, files with dashes in
> their names cannot be included by other Thrift files.
> 
> ```
> include "foo-bar.thrift"
> ```
> 
[...]

PR #424, opened 2019-10-10
Hyphenated file support

abhishekparwal:
added support for the Hyphenated files as discussed in #143 

Details:
Go Code generated would be having a package name as abc_def, i.e hyphen converted to underscore for working around golang casing constraints.

Hyphenated files can not be included as an identifier of that form are not allowed in thrift grammar. This is because of usages with hyphens like abc-def.XYZ.

Not adding support for include-as directive as this is a custom thrift syntax which is not officially supported and leaked into thriftrw grammar earlier. however, #423 adds support for it, which we are not landing now.


compile/error.go
```diff
@@ -64,6 +64,14 @@ func (e includeAsDisabledError) Error() string {
 	return "include-as syntax is currently disabled"
 }
 
+// includingHyphenatedFileNotAllowed is raised when the user attempts to include hyphenated thrift file which is not
```
AlexAPB: break up the comment line to 80 chars width

AlexAPB: And re-word please. Adding that it is not allowed is redundant since this is an error type:
```
// includingHyphenatedFileNotAllowed is raised when the user attempts to include hyphenated file names.
```

gen/generate.go
```diff
@@ -165,6 +169,43 @@ func Generate(m *compile.Module, o *Options) error {
...
+func isDuplicateFileAfterNormalization(f string) error {
+	absFileName, err := filepath.Abs(f)
+	if err != nil {
+		return generateError{Name: f, Reason: fmt.Errorf("File %q does not exist: %v", f,
+			err)}
+	}
+	normalizedFileName := filePathWithUnderscore(absFileName)
+	if fileNormalized(absFileName, normalizedFileName) && fileExists(normalizedFileName) {
+		return generateError{Name: f,
+			Reason: fmt.Errorf("File after normalization %q is colliding with existing file %q in the path, with error: %v",
```
AlexAPB: - go errors always start lower-case (I think `go vet` should catch this 🤔 ?)
- we can simplify the error message, and should re-order the arguments
- the `err` value used here is not relevant and will always be `nil`
```suggestion
			Reason: fmt.Errorf("normalizing %q collides with existing file %q", absFileName, normalizedFileName)
```

gen/generate.go
```diff
@@ -165,6 +169,43 @@ func Generate(m *compile.Module, o *Options) error {
 	return nil
 }
 
+func isDuplicateFileAfterNormalization(f string) error {
+	absFileName, err := filepath.Abs(f)
+	if err != nil {
+		return generateError{Name: f, Reason: fmt.Errorf("File %q does not exist: %v", f,
```
AlexAPB: nit: awkward formatting
```golang
return generateError{
  Name: f, 
  Reason: fmt.Errorf("File %q does not exist: %v", f, err),
}
```

gen/generate.go
```diff
@@ -226,12 +267,14 @@ func generateModule(
 	builder *generateServiceBuilder,
 	o *Options,
 ) (outputFilepath string, contents []byte, err error) {
+	// converts file from /home/abc/ab-def.thrift to /home/abc/ab_def.thrift for golang code generation
+	normalizedThriftPath := filePathWithUnderscore(m.ThriftPath)
```
AlexAPB: should we just call this function `normalizeFilePath`?

gen/generate_test.go
```diff
@@ -79,6 +79,62 @@ func TestGenerateWithRelativePaths(t *testing.T) {
...
+func TestGenerateWithHyphenPaths(t *testing.T) {
+	outputDir, err := ioutil.TempDir("", "thriftrw-generate-test")
+	require.NoError(t, err)
+	defer os.RemoveAll(outputDir)
+
+	thriftRoot, err := os.Getwd()
+	require.NoError(t, err)
+
+	tests := []struct {
+		filepath      string
```
AlexAPB: Lets add a `name string` property to the tests and use [subtests](https://blog.golang.org/subtests).
https://github.com/uber-go/guide/blob/master/style.md#test-tables

gen/generate_test.go
```diff
@@ -79,6 +79,62 @@ func TestGenerateWithRelativePaths(t *testing.T) {
...
+		{
+			filepath:      "internal/tests/thrift/nestedfiles_error/abc-defs-nested.thrift",
+			generateError: "is colliding with existing file",
+		},
+		{
+			filepath: "internal/tests/thrift/nestedfiles_error/abc_defs_nested.thrift",
+		},
+	}
+
+	for _, test := range tests {
```
AlexAPB: style nit: generally seen the following at Uber: 
```suggestion
	for _, tt := range tests {
```
https://github.com/uber-go/guide/blob/master/style.md#test-tables

gen/generate_test.go
```diff
@@ -79,6 +79,62 @@ func TestGenerateWithRelativePaths(t *testing.T) {
...
+		module, err := compile.Compile(test.filepath)
+		if test.compileError != "" {
+			assert.Error(t, err)
```
AlexAPB: `require.Error`

if there's no error here, the next line will panic

gen/generate_test.go
```diff
@@ -79,6 +79,62 @@ func TestGenerateWithRelativePaths(t *testing.T) {
...
+			assert.Contains(t, err.Error(), test.compileError)
+		} else {
+			assert.NoError(t, err)
+		}
+		if module == nil {
+			continue
+		}
```
AlexAPB: simplify:
```
			assert.Contains(t, err.Error(), test.compileError)
			continue
		} 

		require.NoError(t, err)
		require.NotNil(t, module)
```

gen/generate_test.go
```diff
@@ -79,6 +79,62 @@ func TestGenerateWithRelativePaths(t *testing.T) {
...
+
+		opt := &Options{
+			OutputDir:     outputDir,
+			PackagePrefix: "go.uber.org/thriftrw/gen",
+			ThriftRoot:    thriftRoot,
+		}
+		err = Generate(module, opt)
+		if test.generateError != "" {
+			assert.Error(t, err, "expected code generation with filepath %v to fail", test.filepath)
```
AlexAPB: `require.Error`

gen/internal/tests/thrift/abc-defs.thrift
```diff
@@ -0,0 +1,5 @@
+include "./non_hyphenated.thrift"
```
AlexAPB: Can we rename this file (and the others)? `abc-defs.thrift` gives no indication of what this file is used for.

gen/generate.go
```diff
@@ -165,6 +169,43 @@ func Generate(m *compile.Module, o *Options) error {
...
+	absFileName, err := filepath.Abs(f)
+	if err != nil {
+		return generateError{Name: f, Reason: fmt.Errorf("File %q does not exist: %v", f,
+			err)}
+	}
+	normalizedFileName := filePathWithUnderscore(absFileName)
+	if fileNormalized(absFileName, normalizedFileName) && fileExists(normalizedFileName) {
+		return generateError{Name: f,
+			Reason: fmt.Errorf("File after normalization %q is colliding with existing file %q in the path, with error: %v",
+				normalizedFileName, absFileName, err)}
```
AlexAPB:
```suggestion
```

gen/generate.go
```diff
@@ -105,6 +105,10 @@ func Generate(m *compile.Module, o *Options) error {
 	genBuilder := newGenerateServiceBuilder(importer)
 
 	generate := func(m *compile.Module) error {
+		if err := isDuplicateFileAfterNormalization(m.ThriftPath); err != nil {
```
AlexAPB: This name is a bit of a mouthful, consider renaming to something more succinct.

gen/generate_test.go
```diff
@@ -79,6 +79,62 @@ func TestGenerateWithRelativePaths(t *testing.T) {
 	}
 }
 
+func TestGenerateWithHyphenPaths(t *testing.T) {
+	outputDir, err := ioutil.TempDir("", "thriftrw-generate-test")
```
AlexAPB: let's use `tmp/thriftrw-generate-test` so we don't clutter the existing directory

compile/error.go
```diff
@@ -64,6 +64,14 @@ func (e includeAsDisabledError) Error() string {
...
+// allowed
+type includingHyphenatedFileNotAllowed struct{}
+
+func (e includingHyphenatedFileNotAllowed) Error() string {
+	return "including hyphenated thrift file is not allowed"
```
AlexAPB: Let's be more consistent with the existing errors
```suggestion
	return "cannot include hyphenated Thrift files"
```

gen/generate_test.go
abhishekparwal: usages need to renamed too. hence would do in a commit

gen/generate_test.go
abhishekparwal: thanks!

gen/generate_test.go
abhishekparwal: 👍

gen/internal/tests/thrift/abc-defs.thrift
abhishekparwal: sure

gen/generate.go
abhishekparwal: okay

gen/generate.go
abhishekparwal: done

compile/compiler.go
```diff
@@ -242,6 +243,8 @@ func (c compiler) gather(m *Module, prog *ast.Program) error {
 // include loads the file specified by the given include in the given Module.
 //
 // The path to the file is relative to the ThriftPath of the given module.
+// Among other errors, one place it will error out if you attempt to include
+// a file whose name starts with hyphen.
```
AlexAPB: It's not "starts" right, it's "contains"? We can make this short and sweet:
```
Including hyphenated file names will error.
```

compile/compiler.go
```diff
@@ -251,6 +252,14 @@ func (c compiler) include(m *Module, include *ast.Include) (*IncludedModule, err
 		}
 	}
 
+	includeName := fileBaseName(include.Path)
+	if strings.Contains(includeName, "-") {
```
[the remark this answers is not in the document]
AlexAPB: Thoughts on collapsing this still? The intermediate variable doesn't save us much here.

compile/error.go
```diff
@@ -64,6 +64,14 @@ func (e includeAsDisabledError) Error() string {
 	return "include-as syntax is currently disabled"
 }
 
+// includeHyphenatedFileNameError is raised when the user attempts to
+// include hyphenated file names
```
AlexAPB: please add a period to the end of comments

gen/generate.go
```diff
@@ -165,6 +169,38 @@ func Generate(m *compile.Module, o *Options) error {
 	return nil
 }
 
+func duplicateAfterNormalization(fileName string) error {
+	absFileName, err := filepath.Abs(fileName)
+	if err != nil {
+		return generateError{Name: fileName, Reason: fmt.Errorf("File %q does not exist: %v", fileName,
```
AlexAPB: let's break up the formatting here too for easier readability
```go
return  generateError{
  Name: fileName,
  Reason: fmt.Errorf("File %q does not exist: %v", fileName, err),
}
```

gen/generate.go
```diff
@@ -165,6 +169,38 @@ func Generate(m *compile.Module, o *Options) error {
...
+			err)}
+	}
+	normalizedFileName := normalizeFilePath(absFileName)
+	if fileNormalized(absFileName, normalizedFileName) && fileExists(normalizedFileName) {
```
AlexAPB: what if we replaced the `fileNormalized` function with an inline variable?
```go

fileIsNormalized := (absFileName != normalizedFileName)
if fileIsNormalized && fileExists(normalizedFileName) {
 ...
}

File: gen/generate.go [hunk 20]
```diff
@@ -165,6 +169,38 @@ func Generate(m *compile.Module, o *Options) error {
...
+}
+
+func fileExists(filename string) bool {
+	info, err := os.Stat(filename)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return !info.IsDir()
+}
+
```
AlexAPB: let's add a small comment here since the functionality isn't immediately obvious.

```suggestion
// normalizeFilePath replaces hyphens in the file name with underscores.
```

gen/generate_test.go
AlexAPB: whoops, I misread the API!

gen/generate_test.go
```diff
@@ -79,6 +79,67 @@ func TestGenerateWithRelativePaths(t *testing.T) {
...
+	require.NoError(t, err)
+
+	tests := []struct {
+		name          string
+		filepath      string
+		generateError string
+		compileError  string
+	}{
+		{
+			name:         "hyphen files inclusion gives include error",
```
AlexAPB:
```suggestion
			name:         "include with hyphens error",
```

gen/generate_test.go
```diff
@@ -79,6 +79,67 @@ func TestGenerateWithRelativePaths(t *testing.T) {
...
+			filepath:     "internal/tests/thrift/nestedfiles_error/include_hyphen_files.thrift",
+			compileError: "cannot include hyphenated Thrift files",
+		},
+		{
+			name:     "hyphen files as top level file allowed",
+			filepath: "internal/tests/thrift/hyphenated-file.thrift",
+		},
+		{
+			name:          "hyphen file post normalization code gen fails with non hyphen same name file",
```
AlexAPB:
```suggestion
			name:          "normalization collision err",
```

gen/generate_test.go
```diff
@@ -79,6 +79,67 @@ func TestGenerateWithRelativePaths(t *testing.T) {
...
+			filepath:      "internal/tests/thrift/nestedfiles_error/hyphenated-file-nested.thrift",
+			generateError: "collides with existing file",
+		},
+		{
+			name:     "non hyphen file code gen doesn't collide with same name file post normalization of hyphen file",
```
AlexAPB:
```suggestion
			name:     "normalization success",
```

gen/generate_test.go
```diff
@@ -79,6 +79,67 @@ func TestGenerateWithRelativePaths(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			module, err := compile.Compile(tt.filepath)
+			if tt.compileError != "" {
+				require.Error(t, err)
+				assert.Contains(t, err.Error(), tt.compileError)
+				return
+			}
+			assert.NoError(t, err)
```
AlexAPB:
```suggestion
			require.NoError(t, err)
```

gen/generate_test.go
AlexAPB: It's expected to error, yes, but if we ended up changing some functionality in thriftrw that causes this test to fail, it's better to have it fail early instead of panic

gen/generate.go
```diff
@@ -105,6 +105,10 @@ func Generate(m *compile.Module, o *Options) error {
 	genBuilder := newGenerateServiceBuilder(importer)
 
 	generate := func(m *compile.Module) error {
+		if err := duplicateAfterNormalization(m.ThriftPath); err != nil {
```
AlexAPB: I think that this error check may make more sense in the `generateModule` function, since that is where we attempt to normalize the file name.

gen/generate.go
```diff
@@ -186,7 +222,7 @@ type thriftPackageImporter struct {
 }
 
 func (i thriftPackageImporter) RelativePackage(file string) (string, error) {
-	return filepath.Rel(i.ThriftRoot, strings.TrimSuffix(file, ".thrift"))
+	return filepath.Rel(i.ThriftRoot, strings.TrimSuffix(normalizeFilePath(file), ".thrift"))
```
AlexAPB: Since we pass in the normalized path on line 272, can we remove this change?

RohanRathi:
Hi, when can this PR be merged? My work is pending on its approval

abhinav:
Apologies for the delay @abhishekparwal, @RohanRathi. We've been pretty under-
water with other higher priority work, so there hasn't been ample opportunity
to thoroughly review this change. We'll try to give it more review passes as
soon as we're able.

This is a nice graceful degradation for certain failure scenarios, but it
shouldn't be a blocker for any work. It's a matter of renaming your file. It
should be perfectly safe to rename those files because no other Thrift file
can legally include and use them.

gen/generate.go
```diff
@@ -226,12 +256,17 @@ func generateModule(
 	builder *generateServiceBuilder,
 	o *Options,
 ) (outputFilepath string, contents []byte, err error) {
+	if err := duplicateAfterNormalization(m.ThriftPath); err != nil {
```
abhinav: This is normalizing twice. The logic is effectively,

    newPath = normalize(path)
    if checkStuff(newPath) {
    }
    
    newPath = normalize(path)

I would drop `duplicateAfterNormalization` and move the duplication check here.

gen/generate.go
```diff
@@ -165,6 +165,36 @@ func Generate(m *compile.Module, o *Options) error {
 	return nil
 }
 
+func duplicateAfterNormalization(fileName string) error {
+	absFileName, err := filepath.Abs(fileName)
+	if err != nil {
+		return generateError{
+			Name:   fileName,
+			Reason: fmt.Errorf("File %q does not exist: %v", fileName, err),
```
abhinav: 1. [error strings should start with lower case](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)
2. This error message isn't accurate. `filepath.Abs` could fail for a different reason, not just because the file doesn't exist. I would use `Reason: err` here directly.
3. According to the [compile.Module docs](https://godoc.org/go.uber.org/thriftrw/compile#Module), ThriftPath is known to be the absolute path to the Thrift file so `filepath.Abs` on it is unnecessary.

gen/generate.go
```diff
@@ -165,6 +165,36 @@ func Generate(m *compile.Module, o *Options) error {
...
+	}
+	return nil
+}
+
+func fileExists(filename string) bool {
+	info, err := os.Stat(filename)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return !info.IsDir()
```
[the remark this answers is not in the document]
abhishekparwal: done

abhinav:
Hey,

1. So you have. I must've misread. My apologies.
2. I'll need to take a second pass to verify but I thought I read something that would break this.
3. As discussed offline, the expectation is that `path/to/thrift/file.thrift` becomes `path/to/thrift/file/*.go`, whatever `file` is. We just need to change the generated package name, `package $(basename $file)` to `package $(normalize $(basename $file))`.

gen/generate.go
abhishekparwal: done

gen/generate.go
abhishekparwal: done

abhishekparwal:
> Hey,
> 
> 1. So you have. I must've misread. My apologies.
> 2. I'll need to take a second pass to verify but I thought I read something that would break this.
> 3. As discussed offline, the expectation is that `path/to/thrift/file.thrift` becomes `path/to/thrift/file/*.go`, whatever `file` is. We just need to change the generated package name, `package $(basename $file)` to `package $(normalize $(basename $file))`.

2. TestGenerateWithHyphenPaths/normalization are the name of test/subtests

3. have modified to generate the package and not change the name of the generated go file.

gen/generate.go
```diff
@@ -253,10 +259,16 @@ func generateModule(
 		return "", nil, err
 	}
 
+	// converts package name from ab-def to ab_def for golang code generation
+	packageRelPath, err := i.RelativePackage(m.ThriftPath)
+	if err != nil {
+		return "", nil, err
+	}
+	normalizedPackageName := filepath.Base(packageRelPath)
```
abhinav: This seems suspect. I think we want the original package path to remain the same, and to normalize here,

    packageName := normalizeRelativePath(filepath.Base(packageRelPath))

`RelativePackage` is supposed to be the import path to the generated code relative to the output directory. You can see that in the `Package` method implementation. It's basically a `join(ImportPrefix, RelativePackage)`.

So if you had ThriftRoot `foo/`, import prefix `example.com/gen/`, and a Thrift file at `foo/bar/baz-qux.thrift`, we want, `RelativePackage = "bar/baz-qux"`, which gives us import path `example.com/gen/bar/baz-qux`, and we want package name `baz_qux`.

abhishekparwal: done.

gen/generate_test.go
```diff
@@ -79,6 +79,62 @@ func TestGenerateWithRelativePaths(t *testing.T) {
...
+		generateError string
+		compileError  string
+	}{
+		{
+			name:         "include with hyphens error",
+			filepath:     "internal/tests/thrift/nestedfiles_error/include_hyphen_files.thrift",
+			compileError: "cannot include hyphenated Thrift files",
+		},
+		{
+			name:     "normalization - hyphen files as top level file allowed",
```
abhinav: can use `/` to separate the test name to match subtests format.

```suggestion
			name:     "normalization/hyphen files as top level file allowed",
```

gen/generate_test.go
```diff
@@ -79,6 +79,62 @@ func TestGenerateWithRelativePaths(t *testing.T) {
...
+			filepath: "internal/tests/thrift/hyphenated-file.thrift",
+		},
+		{
+			name:     "normalization - non hyphen file with hyphen in the directory",
```
abhinav:
```suggestion
			name:     "normalization/non hyphen file with hyphen in the directory",
```

gen/internal/tests/thrift/hyphenated_file.thrift
```diff
@@ -0,0 +1,5 @@
+include "./non_hyphenated.thrift"
```
abhinav: It took me a second to understand why this was called `hyphenated_file` without a hyphen.

We should add a comment.

```suggestion
// This file is named hyphenated_file to possibly conflict with the code
// generated from hyphenated-file.

include "./non_hyphenated.thrift"
```

PR #427, opened 2019-11-20
Add enumItemName as a template function for plugins to use

orangerd:
Add support for `enumItemName` as a template function so that plugins reference the generated item names.


gen/generator.go
```diff
@@ -318,6 +319,54 @@ func (g *generator) recordGenDeclNames(d *ast.GenDecl) (conflict bool, err error
 	return false, nil
 }
 
+func (g *generator) declare(ignoreConflicts bool, s string, data interface{}, opts ...TemplateOption) error {
```
abhinav: Did this need to be moved?

orangerd: The comment was supposed to be for `DeclareFromTemplate`, but it was on `declare` - I could move it back if needed.

orangerd:
> Please add more context on what you're trying to do.
> 
> This change won't make the function available to plugins. It will only make it available to the "native" templates that don't use plugins.
> 
> Plugins are meant to use the `go.uber.org/thriftrw/plugin` APIs, which do not use this code-generation setup.

We're using `thriftrw/plugin/api` for setting up our plugin, but when we generate code from our plugin, we inevitably have to use `gen/generator.go` (otherwise we really can't generate any meaningful code, since we need to know the structure of what thriftrw generates).

gen/generator.go
abhinav: Ah, I wonder how that happened. No, this is fine.

abhinav:
> We're using `thriftrw/plugin/api` for setting up our plugin, but when we generate code from our plugin, we inevitably have to use `gen/generator.go` (otherwise we really can't generate any meaningful code, since we need to know the structure of what thriftrw generates).

That makes sense. Note that until now, the enumItemName function was opt-in in
existing templates with a `TemplateFunc("enumItemName", enumItemName)` call.
So, we can either update the existing templates which opt-in like that to drop
the option, or we can export the function as `EnumItemName` so that you can
opt into it yourself.

orangerd:
> That makes sense. Note that until now, the enumItemName function was opt-in in
> existing templates with a `TemplateFunc("enumItemName", enumItemName)` call.
> So, we can either update the existing templates which opt-in like that to drop
> the option, or we can export the function as `EnumItemName` so that you can
> opt into it yourself.

Thanks! Removed all the refs to `TemplateFunc("enumItemName", enumItemName)` opt-in - please let me know if looks fine or if you'd rather export the function instead.

orangerd:
@abhinav would you be able to merge? I'm not a collaborator so apparently can't merge (I don't have write access to this repo) :-(

PR #430, opened 2019-12-02
Add the ErrorName function to structs that are exceptions

felipefiali:
More context on T4112649.

In order to allow YARPC to tag its metrics with the application error name and also to propagate that to any upstream consumer, we have to expose the exception name in a function to avoid having to reflect on it. 

Adding the ErrorName function allows us to do that.


gen/field.go
```diff
@@ -444,6 +450,17 @@ func (f fieldGroupGenerator) String(g Generator) error {
 		`, f)
 }
 
+func (f fieldGroupGenerator) ErrorName(g Generator) error {
+	return g.DeclareFromTemplate(
+		`
+		<$v := newVar "v">
+		// ErrorName returns a string representation of the type of <.Name>.
+		func (<$v> *<.Name>) ErrorName() string {
+			return "<.Name>"
```
abhinav: This will return the name of the Go struct. I think it'll be preferable here
to return the Thrift-level name in case it was renamed at the Go level,
especially if similar functionality gets added to other languages.

You'll need to plumb the Thrift-level name of the struct to the
fieldGroupGenerator.

felipefiali: Added the tests and change the implementation to return the Thrift name instead.

felipefiali:
@abhinav Added the tests and change the implementation as you suggested.

kriskowal:
This is a great change. We should add a tag in YARPC internal observability middleware for ErrorNamers.

felipefiali:
@abhinav Thank you!

@kriskowal That is precisely the goal. I'll modify YARPC next and then Muttley as T4112649 describes.

PR #431, opened 2019-12-05
Include annotations in Arguments as defined in the Thrift file

dmtsui:
Currently annotations defined in the Thrift file on arguments and exceptions for a function are not available in the plugin. This request adds the annotation field to the Arguments struct to expose the annotations for plugins to use.


idl/internal/lex.rl
```diff
@@ -4,9 +4,7 @@ package internal
 
 import (
     "fmt"
-    "io"
     "strconv"
-    "strings"
```
abhinav: 🤦‍♂ 
thanks for fixing

idl/internal/y.go
```diff
@@ -1,5 +1,37 @@
 // @generated Code generated by yacc
 
+
+// @generated Code generated by yacc
+
+// Copyright (c) 2019 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
```
abhinav: there's something wrong here. would you mind reverting the changes to this file?

plugin/api.thrift
```diff
@@ -130,6 +130,24 @@ struct Argument {
...
+    /**
+     * Annotations defined on this argument.
+     *
+     * Given,
+     *
+     *   void setValue(
+     *   	1: SetValueRequest req
+     *   ) throws (
+     *		1: BadRequestError badRequestError (cache = "false")
+     *   )
```
abhinav: indentation is a bit off here with tabs/spaces. for the thrift file, we should use spaces only.

dmtsui: using spaces only now

idl/internal/y.go
dmtsui: I've reset this as well as the lex.go file to the original state

plugin/api.thrift
```diff
@@ -130,6 +130,24 @@ struct Argument {
...
+     *
+     * Given,
+     *
+     *   void setValue(
+     *     1: SetValueRequest req
+     *   ) throws (
+     *     1: BadRequestError badRequestError (cache = "false")
+     *   )
+     *
+     * The annotations will be,
```
```suggestion
     * The annotations for the Argument representing badRequestError will be,
```

abhinav:
Oops!

PR #432, opened 2019-12-28
Migrate to Go modules

abhinav:
This migrates ThriftRW Go to Go modules. To perform this migration, a
fair amount of bit rot had to be cleaned up so all changes are in
individually reviewable commits.

Things to note:

- TestDeterminePackagePrefix assumed GOPATH
- `go:generate` clauses invoking ThriftRW assumed GOPATH
- `make generate` was generally broken
- `staticcheck` wasn't passing
- We were still on coveralls instead of codecov

---

Commits are individually reviewable.


plugin/gen.go
```diff
@@ -20,6 +20,5 @@
 
 package plugin
 
-//go:generate thriftrw --generate-plugin-api api.thrift
+//go:generate thriftrw --pkg-prefix=go.uber.org/thriftrw/plugin --generate-plugin-api api.thrift
```
kriskowal: this looks like it will ripple. will all projects need to add this flag to their thrift gen lines?

plugin/gen.go
```diff
@@ -20,6 +20,5 @@
...
 //go:generate mockgen -destination plugintest/api.go -package plugintest go.uber.org/thriftrw/plugin/api Plugin,ServiceGenerator
-//go:generate ../scripts/updateLicenses.sh
```
kriskowal: will generated code still have a license block?

Makefile
```diff
@@ -1,92 +1,91 @@
-WANT_VERSION = $(shell grep '^v[0-9]' CHANGELOG.md | head -n1 | cut -d' ' -f1)
+BUILD_DIR = $(shell pwd)/build
```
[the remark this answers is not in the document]
abhinav: Yeah, the gen/internal/tests Makefile calls into it with `make -C` and it works as expected.

wire/value_equals.go
```diff
@@ -28,7 +28,7 @@ import (
 
 // errNotEquals is a sentinel error used while iterating through ValueLists to
 // indicate that two values did not match.
-var errNotEquals = errors.New("Values are not equal")
+var errNotEquals = errors.New("values are not equal")
```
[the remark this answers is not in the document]
abhinav: part of the golint fix

plugin/gen.go
abhinav: No, only projects using Go modules need to specify pkg-prefix. All
internal projects already provide it. This change affects ThriftRW's
invocations of itself because we can no longer assume it's on GOPATH.
In the future, when we use go/packages in ThriftRW, this will be
unnecessary again.

plugin/gen.go
abhinav: Yeah, `make generate` calls it at the end.

PR #440, opened 2020-03-24
Enable users to pass in !omitempty to avoid the default addition of omitempty

styeung:
Hi team,

Sai To Yeung from Uber Eats Restaurant Platform Team here.   Looking to enable a feature to enable some flexibility with how ThriftRW adds in `omitempty` to all optional fields.

With this change, if a user marks a field in the thrift definition with `!omitempty`, the `omitempty` tag will not be added in the final struct.

This is to support use cases in which some clients expect the marshalled response of an empty list or map field to be `[]` or `{}` instead of the entire field being removed altogether.

Happy to provide more context via Slack.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,3 +28,41 @@ release.
 [cov-img]: https://codecov.io/gh/thriftrw/thriftrw-go/branch/dev/graph/badge.svg
 [ci]: https://travis-ci.com/thriftrw/thriftrw-go
 [cov]: https://codecov.io/gh/thriftrw/thriftrw-go
+
+## Development
+
+To install dependencies and build the project run:
+
+```
+make build
+```
+
+### Testing
+
+To run the entire test suite, run:
+
+```
+make test
+```
+
+#### Integration tests
+
+A lot of the codebase has its logic exercised via integration tests residing in `gen/internal/tests`.
+In other words, since ThriftRW is a code generation library, the way to test that code generation behavior is
+implemented correctly is to create a real Thrift struct definition, run code generation, and assert that the output is correct.
+
+Step by step, this process is:
+
+1. Create or use a Thrift struct in `gen/internal/tests/thrift`.  For example, in `gen/internal/tests/thrift/structs.thrift`, you can find
+`struct GoTags` that is used to exercise go.tag generation behavior, or `struct NotOmitEmpty` that is used to exercise behavior 
+when a field is tagged with `!omitempty`
+1. Run `make generate`.  This will generate the go struct from your definition, and place it in `gen/internal/tests/structs/structs.go`
+1. Write your test in one of the `*_test.go` files that is pertinent to the feature you are adding.  Oftentimes, these tests 
+take in the generated go structs as inputs, and assert on various aspects of code generation like struct tags or json marshaling/unmarshaling behavior.
+1. Also remember to add your struct to `gen/quick_test.go` so that your new struct and all of its generic methods (e.g., ToWire, FromWire, String, Equals, etc.) 
+can be exercised for code coverage
+
+*Note*: Code coverage is measured across packages, rather than per package.  This is because `go test` is run with the `-coverpkg=./...` flag,
+which tells the code coverage tool to measure coverage for this package and all packages in the subdirectories.
+
+
--- a/gen/field.go
+++ b/gen/field.go
@@ -183,15 +183,23 @@ func compileJSONTag(f *compile.FieldSpec, name string, opts ...string) *structta
 	// defined as false, 0, a nil pointer, a nil interface value, and any empty
 	// array, slice, map, or string.
 	//
-	// If the field is marked with "!omitempty", then "omitempty" will not be added
-	// For example:
-	//
-	// struct MyStruct {
-	//   1: optional list<Option> options  (go.tag='json:"options,!omitempty"')
-	// }
+	// If the field is marked with "!omitempty", then "omitempty" will not be added.
+	// If both "!omitempty" and "omitempty" are present, then "omitempty" is removed.
+	hasNotOmitempty := t.HasOption(notOmitempty)
+	hasOmitempty := t.HasOption(omitempty)
 	if (isReferenceType(f.Type) || isStructType(f.Type) || isPrimitiveType(f.Type)) &&
-		!f.Required && !t.HasOption(notOmitempty) && !t.HasOption(omitempty) {
+		!f.Required && !hasNotOmitempty && !hasOmitempty {
 		t.Options = append(t.Options, omitempty)
+	} else if hasNotOmitempty && hasOmitempty {
+		newOptions := make([]string, 0)
+		for _, option := range t.Options {
+			if option == omitempty {
+				continue
+			}
+			newOptions = append(newOptions, option)
+		}
+
+		t.Options = newOptions
 	}
 
 	if f.Required && !t.HasOption("required") {
--- a/gen/field.go
+++ b/gen/field.go
@@ -187,19 +187,20 @@ func compileJSONTag(f *compile.FieldSpec, name string, opts ...string) *structta
 	// If both "!omitempty" and "omitempty" are present, then "omitempty" is removed.
 	hasNotOmitempty := t.HasOption(notOmitempty)
 	hasOmitempty := t.HasOption(omitempty)
-	if (isReferenceType(f.Type) || isStructType(f.Type) || isPrimitiveType(f.Type)) &&
-		!f.Required && !hasNotOmitempty && !hasOmitempty {
-		t.Options = append(t.Options, omitempty)
-	} else if hasNotOmitempty && hasOmitempty {
-		newOptions := make([]string, 0)
-		for _, option := range t.Options {
-			if option == omitempty {
-				continue
+	if (isReferenceType(f.Type) || isStructType(f.Type) || isPrimitiveType(f.Type)) && !f.Required {
+		if !hasNotOmitempty && !hasOmitempty {
+			t.Options = append(t.Options, omitempty)
+		} else if hasNotOmitempty && hasOmitempty {
+			newOptions := make([]string, 0)
+			for _, option := range t.Options {
+				if option == omitempty {
+					continue
+				}
+				newOptions = append(newOptions, option)
 			}
-			newOptions = append(newOptions, option)
-		}
 
-		t.Options = newOptions
+			t.Options = newOptions
+		}
 	}
 
 	if f.Required && !t.HasOption("required") {
--- a/gen/field.go
+++ b/gen/field.go
@@ -189,9 +189,13 @@ func compileJSONTag(f *compile.FieldSpec, name string, opts ...string) *structta
 	hasOmitempty := t.HasOption(omitempty)
 	if (isReferenceType(f.Type) || isStructType(f.Type) || isPrimitiveType(f.Type)) && !f.Required {
 		if !hasNotOmitempty && !hasOmitempty {
+			// Add omitempty if it's not there and the user
+			// has not opted out with !omitempty.
 			t.Options = append(t.Options, omitempty)
 		} else if hasNotOmitempty && hasOmitempty {
-			newOptions := make([]string, 0)
+			// !omitempty takes precedence so remove omitempty
+			// if present.
+			newOptions := make([]string, 0, len(t.Options))
 			for _, option := range t.Options {
 				if option == omitempty {
 					continue
--- a/gen/field.go
+++ b/gen/field.go
@@ -185,9 +185,9 @@ func compileJSONTag(f *compile.FieldSpec, name string, opts ...string) *structta
 	//
 	// If the field is marked with "!omitempty", then "omitempty" will not be added.
 	// If both "!omitempty" and "omitempty" are present, then "omitempty" is removed.
-	hasNotOmitempty := t.HasOption(notOmitempty)
-	hasOmitempty := t.HasOption(omitempty)
 	if (isReferenceType(f.Type) || isStructType(f.Type) || isPrimitiveType(f.Type)) && !f.Required {
+		hasNotOmitempty := t.HasOption(notOmitempty)
+		hasOmitempty := t.HasOption(omitempty)
 		if !hasNotOmitempty && !hasOmitempty {
 			// Add omitempty if it's not there and the user
 			// has not opted out with !omitempty.
--- a/gen/field.go
+++ b/gen/field.go
@@ -34,6 +34,9 @@ const (
 
 	// key for tag set on all generated go structs used by encoding/json
 	jsonTagKey = "json"
+
+	omitempty    = "omitempty"
+	notOmitempty = "!omitempty"
 )
 
 var reservedIdentifiers = map[string]struct{}{
@@ -179,9 +182,16 @@ func compileJSONTag(f *compile.FieldSpec, name string, opts ...string) *structta
 	// should be omitted from the encoding if the field has an empty value,
 	// defined as false, 0, a nil pointer, a nil interface value, and any empty
 	// array, slice, map, or string.
+	//
+	// If the field is marked with "!omitempty", then "omitempty" will not be added
+	// For example:
+	//
+	// struct MyStruct {
+	//   1: optional list<Option> options  (go.tag='json:"options,!omitempty"')
+	// }
 	if (isReferenceType(f.Type) || isStructType(f.Type) || isPrimitiveType(f.Type)) &&
-		!f.Required && !t.HasOption("omitempty") {
-		t.Options = append(t.Options, "omitempty")
+		!f.Required && !t.HasOption(notOmitempty) && !t.HasOption(omitempty) {
+		t.Options = append(t.Options, omitempty)
 	}
 
 	if f.Required && !t.HasOption("required") {
--- a/gen/field_test.go
+++ b/gen/field_test.go
@@ -105,7 +105,7 @@ func TestCompileJSONTag(t *testing.T) {
 		{
 			desc: "required with !omitempty",
 			required: true,
-			options: []string{notOmitEmpty},
+			options: []string{notOmitempty},
 			wantOmitEmpty: false,
 		},
 	}
--- a/gen/field_test.go
+++ b/gen/field_test.go
@@ -90,11 +90,13 @@ func TestCompileJSONTag(t *testing.T) {
 			desc:          "optional fields",
 			required:      false,
 			wantOmitempty: true,
-		}, {
+		},
+		{
 			desc:          "required fields",
 			required:      true,
 			wantOmitempty: false,
-		}, {
+		},
+		{
 			desc:          "optional with !omitempty",
 			required:      false,
 			options:       []string{notOmitempty},
--- a/gen/field_test.go
+++ b/gen/field_test.go
@@ -81,26 +81,24 @@ func TestFieldLabelConflict(t *testing.T) {
 
 func TestCompileJSONTag(t *testing.T) {
 	tests := []struct {
-		desc             string
-		required         bool
-		options          []string
-		isOmitemptyAdded bool
+		desc          string
+		required      bool
+		options       []string
+		wantOmitempty bool
 	}{
 		{
-			desc:             "for optional fields",
-			required:         false,
-			options:          make([]string, 0),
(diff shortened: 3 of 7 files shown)


gen/field_test.go
```diff
@@ -76,3 +76,59 @@ func TestFieldLabelConflict(t *testing.T) {
 		})
 	}
 }
+
+func TestCompileJSONTag_ForOptionalFields_OmitemptyIsAdded(t *testing.T) {
+	required := false
+	fieldSpec := &compile.FieldSpec{
+		ID:   0,
+		Name: "numbers",
+		Type: &compile.ListSpec{
```
abhinav: I think these three tests should be collapsed into a table test.

gen/field_test.go
```diff
@@ -76,3 +78,54 @@ func TestFieldLabelConflict(t *testing.T) {
...
+	tests := []struct {
+		desc             string
+		required         bool
+		options          []string
+		isOmitemptyAdded bool
+	}{
+		{
+			desc:             "for optional fields",
+			required:         false,
+			options:          make([]string, 0),
```
abhinav: `nil` is a valid empty slice. This line can be deleted.

https://github.com/uber-go/guide/blob/master/style.md#nil-is-a-valid-slice

```suggestion
```

gen/field_test.go
```diff
@@ -76,3 +78,54 @@ func TestFieldLabelConflict(t *testing.T) {
...
+
+func TestCompileJSONTag(t *testing.T) {
+	tests := []struct {
+		desc             string
+		required         bool
+		options          []string
+		isOmitemptyAdded bool
+	}{
+		{
+			desc:             "for optional fields",
```
abhinav: nit: "for" prefix unnecessary for test case descriptions.

```suggestion
			desc:             "optional fields",
```

gen/field_test.go
```diff
@@ -76,3 +78,54 @@ func TestFieldLabelConflict(t *testing.T) {
...
+			required:         false,
+			options:          make([]string, 0),
+			isOmitemptyAdded: true,
+		}, {
+			desc:             "for required fields",
+			required:         true,
+			options:          make([]string, 0),
```
abhinav:
```suggestion
```

gen/field_test.go
```diff
@@ -76,3 +78,54 @@ func TestFieldLabelConflict(t *testing.T) {
...
+		options          []string
+		isOmitemptyAdded bool
+	}{
+		{
+			desc:             "for optional fields",
+			required:         false,
+			options:          make([]string, 0),
+			isOmitemptyAdded: true,
+		}, {
+			desc:             "for required fields",
```
abhinav:
```suggestion
			desc:             "required fields",
```

gen/field_test.go
```diff
@@ -76,3 +78,54 @@ func TestFieldLabelConflict(t *testing.T) {
...
+			required:         true,
+			options:          make([]string, 0),
+			isOmitemptyAdded: false,
+		}, {
+			desc:             "when !omitempty tag is present",
```
abhinav:
```suggestion
			desc:             "optional with !omitempty",
```

gen/field_test.go
```diff
@@ -76,3 +78,54 @@ func TestFieldLabelConflict(t *testing.T) {
...
+				Name: fieldName,
+				Type: &compile.ListSpec{
+					ValueSpec: &compile.I64Spec{},
+				},
+				Required: tt.required,
+				Annotations: compile.Annotations{
+					goTagKey: fmt.Sprintf(
+						"json:%s,%s",
+						fieldName,
+						strings.Join(tt.options, ","),
```
abhinav: The part after the `json:` should be quoted so the result is `json:"foo,bar"`.
Also, can we build this value in a local variable above? The fmt.Sprintf and strings.Join is getting pretty loud here.

gen/field_test.go
```diff
@@ -76,3 +78,54 @@ func TestFieldLabelConflict(t *testing.T) {
...
+			required:         true,
+			options:          make([]string, 0),
+			isOmitemptyAdded: false,
+		}, {
+			desc:             "when !omitempty tag is present",
+			required:         false,
+			options:          []string{notOmitempty},
+			isOmitemptyAdded: false,
+		},
+	}
```
abhinav: Can we get a case for required with `!omitempty` present?

gen/field_test.go
```diff
@@ -76,3 +78,54 @@ func TestFieldLabelConflict(t *testing.T) {
 		})
 	}
 }
+
+func TestCompileJSONTag(t *testing.T) {
+	tests := []struct {
+		desc             string
+		required         bool
+		options          []string
+		isOmitemptyAdded bool
```
abhinav: How about naming this `wantOmitempty`?

README.md
```diff
@@ -28,3 +28,36 @@ release.
...
+```
+
+### Testing
+
+To run the entire test suite, run:
+```
+make test
+```
+
+#### Integration tests
```
abhinav: Thanks for adding this!

gen/field_test.go
```diff
@@ -76,3 +78,54 @@ func TestFieldLabelConflict(t *testing.T) {
...
+		desc          string
+		required      bool
+		options       []string
+		wantOmitempty bool
+	}{
+		{
+			desc:          "optional fields",
+			required:      false,
+			wantOmitempty: true,
+		}, {
```
abhinav: style nit:
```suggestion
		},
		{
```

gen/field_test.go
```diff
@@ -76,3 +78,54 @@ func TestFieldLabelConflict(t *testing.T) {
...
+			desc:          "required fields",
+			required:      true,
+			wantOmitempty: false,
+		}, {
```
abhinav:
```suggestion
		},
		{
```

gen/field_test.go
abhinav:
```suggestion
		{
			desc: "required with !omitempty",
			required: true,
			options: []string{notOmitEmpty},
			wantOmitEmpty: false,
		},
	}
```

gen/struct_test.go
```diff
@@ -1193,6 +1193,56 @@ func TestStructGoTags(t *testing.T) {
...
+	notOmitemptyStruct := &ts.NotOmitEmpty{
+		NotOmitEmptyString: nil,
+		NotOmitEmptyInt:    nil,
+		NotOmitEmptyBool:   nil,
+		NotOmitEmptyList:   nil,
+		NotOmitEmptyMap:    nil,
+	}
+
+	// assert that when the !omitempty tag is present in thrift, omitempty is not added to the go tags
+	notOmitEmptyString, _ := reflect.TypeOf(notOmitemptyStruct).Elem().FieldByName("NotOmitEmptyString")
```
abhinav: Can we pull the repeated part here into a local variable.

```suggestion
	structT := reflect.TypeOf(ts.NotOmitEmpty{})
	notOmitEmptyString, _ := structT.FieldByName("NotOmitEmptyString")
```

gen/struct_test.go
```diff
@@ -1193,6 +1193,56 @@ func TestStructGoTags(t *testing.T) {
...
 }
 
+func TestNotOmitEmpty_GoTagBehavior(t *testing.T) {
+	notOmitemptyStruct := &ts.NotOmitEmpty{
+		NotOmitEmptyString: nil,
+		NotOmitEmptyInt:    nil,
+		NotOmitEmptyBool:   nil,
+		NotOmitEmptyList:   nil,
+		NotOmitEmptyMap:    nil,
+	}
```
abhinav: nit: can delete this struct

gen/struct_test.go
```diff
@@ -1193,6 +1193,56 @@ func TestStructGoTags(t *testing.T) {
 	assert.Equal(t, `json:"foobarWithRequired,required"`, string(foobarWithRequired.Tag))
 }
 
+func TestNotOmitEmpty_GoTagBehavior(t *testing.T) {
```
abhinav: How about

```suggestion
func TestNotOmitEmptyStructTags(t *testing.T) {
```

gen/struct_test.go
```diff
@@ -1193,6 +1193,56 @@ func TestStructGoTags(t *testing.T) {
...
+	assert.Equal(t, `json:"notOmitEmptyInt,!omitempty"`, string(notOmitEmptyInt.Tag))
+	notOmitEmptyBool, _ := reflect.TypeOf(notOmitemptyStruct).Elem().FieldByName("NotOmitEmptyBool")
+	assert.Equal(t, `json:"notOmitEmptyBool,!omitempty"`, string(notOmitEmptyBool.Tag))
+	notOmitEmptyList, _ := reflect.TypeOf(notOmitemptyStruct).Elem().FieldByName("NotOmitEmptyList")
+	assert.Equal(t, `json:"notOmitEmptyList,!omitempty"`, string(notOmitEmptyList.Tag))
+	notOmitEmptyMap, _ := reflect.TypeOf(notOmitemptyStruct).Elem().FieldByName("NotOmitEmptyMap")
+	assert.Equal(t, `json:"notOmitEmptyMap,!omitempty"`, string(notOmitEmptyMap.Tag))
+}
+
+func TestNotOmitEmpty_MarshallingBehavior(t *testing.T) {
```
abhinav: Since this is testing both,

```suggestion
func TestOmitEmptyMarshal(t *testing.T) {
```

gen/struct_test.go
```diff
@@ -1193,6 +1193,56 @@ func TestStructGoTags(t *testing.T) {
...
+		ByteField:   nil,
+		Int16Field:  nil,
+		Int32Field:  nil,
+		Int64Field:  nil,
+		DoubleField: nil,
+		StringField: nil,
+		BinaryField: nil,
+	}
+	defaultStructResult, _ := json.Marshal(defaultStruct)
+	assert.Equal(t, "{}", string(defaultStructResult))
```
abhinav: nits:

- can simplify.
- we need to check errors
- prefer short variable names https://github.com/golang/go/wiki/CodeReviewComments#variable-names
- let's use subtests here

```suggestion
	t.Run("omitempty", func(t *testing.T) {
		out, err := json.Marshal(ts.PrimitiveOptionalStruct{})
		require.NoError(t, err)
		assert.Equal(t, "{}", string(out))
	})
```

gen/struct_test.go
```diff
@@ -1193,6 +1193,56 @@ func TestStructGoTags(t *testing.T) {
...
+		NotOmitEmptyString: nil,
+		NotOmitEmptyInt:    nil,
+		NotOmitEmptyBool:   nil,
+		NotOmitEmptyList:   nil,
+		NotOmitEmptyMap:    nil,
+	}
+	expected := `{"notOmitEmptyString":null,"notOmitEmptyInt":null,"notOmitEmptyBool":null,"notOmitEmptyList":null,"notOmitEmptyMap":null}`
+	result, _ := json.Marshal(notOmitemptyStruct)
+	assert.Equal(t, expected, string(result))
+}
```
abhinav: same as above plus can use [assert.JSONEq][1] because ordering of items in a
map isn't guaranteed.

  [1]: https://godoc.org/github.com/stretchr/testify/assert#JSONEq
```suggestion
	t.Run("!omitempty", func(t *testing.T) {
		expected := `{"notOmitEmptyString":null,"notOmitEmptyInt":null,"notOmitEmptyBool":null,"notOmitEmptyList":null,"notOmitEmptyMap":null}`
		result, err := json.Marshal(ts.NotOmitEmpty{})
		require.NoError(t, err)
		assert.JSONEq(t, expected, string(result))
	})
}
```

gen/field_test.go
```diff
@@ -76,3 +78,62 @@ func TestFieldLabelConflict(t *testing.T) {
...
+		{
+			desc:          "optional with !omitempty",
+			required:      false,
+			options:       []string{notOmitempty},
+			wantOmitempty: false,
+		},
+		{
+			desc: "required with !omitempty",
+			required: true,
+			options: []string{notOmitEmpty},
```
abhinav:
```suggestion
			options: []string{notOmitempty},
```

gen/field_test.go
```diff
@@ -76,3 +78,62 @@ func TestFieldLabelConflict(t *testing.T) {
...
+			desc:          "optional with !omitempty",
+			required:      false,
+			options:       []string{notOmitempty},
+			wantOmitempty: false,
+		},
+		{
+			desc: "required with !omitempty",
+			required: true,
+			options: []string{notOmitempty},
+			wantOmitEmpty: false,
```
abhinav:
```suggestion
			wantOmitempty: false,
```

abhinav:
The remaining coverage failure is fine. It's inflated because of generated code volume.

gen/internal/tests/thrift/structs.thrift
```diff
@@ -137,6 +137,14 @@ struct GoTags {
...
 }
 
+struct NotOmitEmpty {
+    1: optional string NotOmitEmptyString (go.tag = 'json:"notOmitEmptyString,!omitempty"')
+    2: optional string NotOmitEmptyInt (go.tag = 'json:"notOmitEmptyInt,!omitempty"')
+    3: optional string NotOmitEmptyBool (go.tag = 'json:"notOmitEmptyBool,!omitempty"')
+    4: optional list<string> NotOmitEmptyList (go.tag = 'json:"notOmitEmptyList,!omitempty"')
+    5: optional map<string, string> NotOmitEmptyMap (go.tag = 'json:"notOmitEmptyMap,!omitempty"')
+}
+
```
mh-park: Can I ask for a case where there is a mix of `!omitempty` and `omitempty`?

styeung: Sure, will add that in.  The expected behavior will be that !omitempty will take precedence.  !omitempty is an opt-in behavior, whereas because omitempty has become the default in our codegen, it is rarely added, if ever.

styeung: Added the test cases.  I'm glad you pointed it out because we have to actively strip out `omitempty` when both are present.  Let me know if you have any further comments or questions.

gen/internal/tests/thrift/structs.thrift
```diff
@@ -143,6 +143,8 @@ struct NotOmitEmpty {
     3: optional string NotOmitEmptyBool (go.tag = 'json:"notOmitEmptyBool,!omitempty"')
     4: optional list<string> NotOmitEmptyList (go.tag = 'json:"notOmitEmptyList,!omitempty"')
     5: optional map<string, string> NotOmitEmptyMap (go.tag = 'json:"notOmitEmptyMap,!omitempty"')
+    6: optional list<string> NotOmitEmptyListMixedWithOmitEmpty (go.tag = 'json:"notOmitEmptyListMixedWithOmitEmpty,!omitempty,omitempty"')
+    7: optional list<string> NotOmitEmptyListMixedWithOmitEmptyV2 (go.tag = 'json:"notOmitEmptyListMixedWithOmitEmptyV2,omitempty,!omitempty"')
```
mh-park: Ah, I meant when different fields have different options, but this is a good case as well. Sorry about the constant back and forth, but can I still ask for that as well? Something like:

```diff
struct NotOmitEmpty {
    1: optional string NotOmitEmptyString (go.tag = 'json:"notOmitEmptyString,!omitempty"')
    2: optional string NotOmitEmptyInt (go.tag = 'json:"notOmitEmptyInt,!omitempty"')
    3: optional string NotOmitEmptyBool (go.tag = 'json:"notOmitEmptyBool,!omitempty"')
    4: optional list<string> NotOmitEmptyList (go.tag = 'json:"notOmitEmptyList,!omitempty"')
    5: optional map<string, string> NotOmitEmptyMap (go.tag = 'json:"notOmitEmptyMap,!omitempty"')
    6: optional list<string> NotOmitEmptyListMixedWithOmitEmpty (go.tag = 'json:"notOmitEmptyListMixedWithOmitEmpty,!omitempty,omitempty"')
    7: optional list<string> NotOmitEmptyListMixedWithOmitEmptyV2 (go.tag = 'json:"notOmitEmptyListMixedWithOmitEmptyV2,omitempty,!omitempty"')
+   8: optional string OmitEmptyString (go.tag = 'json:"omitEmptyString,omitempty")
}
```

I don't see any reason why that wouldn't work, but just to make sure :)

gen/field.go
```diff
@@ -183,15 +183,23 @@ func compileJSONTag(f *compile.FieldSpec, name string, opts ...string) *structta
...
-	// }
+	// If the field is marked with "!omitempty", then "omitempty" will not be added.
+	// If both "!omitempty" and "omitempty" are present, then "omitempty" is removed.
+	hasNotOmitempty := t.HasOption(notOmitempty)
+	hasOmitempty := t.HasOption(omitempty)
 	if (isReferenceType(f.Type) || isStructType(f.Type) || isPrimitiveType(f.Type)) &&
-		!f.Required && !t.HasOption(notOmitempty) && !t.HasOption(omitempty) {
+		!f.Required && !hasNotOmitempty && !hasOmitempty {
 		t.Options = append(t.Options, omitempty)
+	} else if hasNotOmitempty && hasOmitempty {
```
abhinav: I believe these omitempty conditionals should go inside a nested if clause after it has determined that omitempty would be needed on a field. That is, up to:

```
if (isReferenceType(f.Type) || isStructType(f.Type) || isPrimitiveType(f.Type)) && !f.isRequired {
  // We would normally add omitempty here.
}
```
Inside that block, we can add the logic of adding it, removing it, or leaving it.

gen/field.go
```diff
@@ -183,15 +183,23 @@ func compileJSONTag(f *compile.FieldSpec, name string, opts ...string) *structta
...
+		newOptions := make([]string, 0)
+		for _, option := range t.Options {
+			if option == omitempty {
+				continue
+			}
+			newOptions = append(newOptions, option)
```
abhinav: Can we not use Tags.DeleteOptions here?

https://godoc.org/github.com/fatih/structtag#Tags.DeleteOptions

styeung: unfortunately, that method is on `Tags` while the `compileJSONTag`  is operating on a single tag level.

abhinav: Makes sense.

gen/field.go
```diff
@@ -179,9 +182,25 @@ func compileJSONTag(f *compile.FieldSpec, name string, opts ...string) *structta
...
 	// array, slice, map, or string.
-	if (isReferenceType(f.Type) || isStructType(f.Type) || isPrimitiveType(f.Type)) &&
-		!f.Required && !t.HasOption("omitempty") {
-		t.Options = append(t.Options, "omitempty")
+	//
+	// If the field is marked with "!omitempty", then "omitempty" will not be added.
+	// If both "!omitempty" and "omitempty" are present, then "omitempty" is removed.
+	hasNotOmitempty := t.HasOption(notOmitempty)
+	hasOmitempty := t.HasOption(omitempty)
+	if (isReferenceType(f.Type) || isStructType(f.Type) || isPrimitiveType(f.Type)) && !f.Required {
```
abhinav: nit: can move these into the if block
```suggestion
	if (isReferenceType(f.Type) || isStructType(f.Type) || isPrimitiveType(f.Type)) && !f.Required {
		hasNotOmitempty := t.HasOption(notOmitempty)
		hasOmitempty := t.HasOption(omitempty)
```

gen/field.go
```diff
@@ -179,9 +182,25 @@ func compileJSONTag(f *compile.FieldSpec, name string, opts ...string) *structta
...
+		if !hasNotOmitempty && !hasOmitempty {
+			t.Options = append(t.Options, omitempty)
+		} else if hasNotOmitempty && hasOmitempty {
+			newOptions := make([]string, 0)
```
abhinav: Let's add an explanation.
Also, we can pre-allocate the slice to the desired capacity.

```suggestion
		if !hasNotOmitempty && !hasOmitempty {
			// Add omitempty if it's not there and the user
			// has not opted out with !omitempty.
			t.Options = append(t.Options, omitempty)
		} else if hasNotOmitempty && hasOmitempty {
			// !omitempty takes precedence so remove omitempty
			// if present.
			newOptions := make([]string, 0, len(t.Options))
```

gen/internal/tests/thrift/structs.thrift
styeung: Added, thanks for the comments.

PR #443, opened 2020-03-31
Prepare release v1.23.0

abhinav:
This releases v1.23.0 of ThriftRW Go.


abhinav:
Coverage failures can be ignored because they're due to code-gen paths that literally cannot be reached. i.e.:

```go
foo, err := something, error(nil)
if err != nil {
  // ...
}
```

PR #444, opened 2020-04-21
gen: retype Options.Plugin for external implements

twilly:
Options.Plugin was a internal/plugin.Handle. That meant packages outside
of go.uber.org/thriftrw could not implement a plugin that gen.Generate
could use.

A new, user implementable type, struct is defined: CodeGenerator. It
currently has a single field, the service generator, but it could be
extended by adding new generator fields if we want to maintain API
compatibility (this package currently does not have an API guarantee).
Options.Plugin is now specified as a CodeGenerator.

The previous type included a io.Closer and factory for generators. This
makes more sense from a higher level as plugins are typically forked
programs talking over stdin/out. However, it's a detail the generator
does not care for. It's happy so long as it can call the Generate method
of a ServiceGenerator. Including the I/O detail breaks the abstraction
of what the plugin truly are: a callable that generates code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- No changes yet.
+### Changed
+- gen: Redefine Options.Plugin as a struct usable outside go.uber.org/thriftrw.
 
 ## [1.23.0] - 2020-03-31
 ### Added
--- a/gen/generate.go
+++ b/gen/generate.go
@@ -30,9 +30,15 @@ import (
 
 	"go.uber.org/thriftrw/compile"
 	"go.uber.org/thriftrw/internal/plugin"
+	"go.uber.org/thriftrw/plugin/api"
 
 	"go.uber.org/multierr"
 )
+
+// CodeGenerator lists possible code generators for a plugin.
+type CodeGenerator struct {
+	ServiceGenerator api.ServiceGenerator
+}
 
 // Options controls how code gets generated.
 type Options struct {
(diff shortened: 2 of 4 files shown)


gen/generate.go
```diff
@@ -30,10 +30,16 @@ import (
 
 	"go.uber.org/thriftrw/compile"
 	"go.uber.org/thriftrw/internal/plugin"
+	"go.uber.org/thriftrw/plugin/api"
 
 	"go.uber.org/multierr"
 )
 
+// CodeGenerator lists possible code generators for a plugin.
```
r-hang, line 38: is 'lists possible code generators' a correct description? api.SerrviceGenerator is anything that takes Service data and returns generated code files?

gen/generate.go
```diff
@@ -30,10 +30,16 @@ import (
...
+type CodeGenerator struct {
```
r-hang, line 39: I think Name() would be a carryover from internal/plugin.Handle

CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- No changes yet.
+### Changed
+- gen.Options.Plugin was redefined to be a gen.CodeGenerator, allowing for
```
abhinav, line 9:
```suggestion
- gen: Redefine Options.Plugin as a struct usable outside go.uber.org/thriftrw.
```

gen/generate.go
abhinav: I think the idea is that if we ever add another hook to plugin/api, that'll go here as a separate code generator. (ServiceGenerator is the pluggable generator for "I want to generate code for `service`s declared in the Thrift file.")

gen/generate.go
abhinav: Looks like name is unused at this level.

internal/plugin.Handle is a reference to a specific external ThriftRW plugin. The name is probably meant for debugging.

At this layer, we really just care about the different code generators provided by the plugins, if at all. Their names are less important. (Note that the plugin.Handle is still in scope in main.go where the  name may be relevant.)

gen/generate.go
twilly: Yep, that's the idea. We'll see if it ever happens. :)

PR #445, opened 2020-06-16
Fix mutation dependent tests

SyrieBianco:
Currently, the `ToWire` functionality mutates the generated thrift type to overwrite nil values with default values. We intend to change this. However, the `thriftRoundTrip` and `GetAccessor` tests depend on this mutation to pass. 

Split into two commits, this PR does the following:
1. add a `Default_*` constructor method to the generated code to provide easy access to the generated types.
2. Decouple the tests from their dependence on mutation by copying values before calling `ToWire` and using the default values exposed in the previous commit to make assertions.

The `ToWire` changes themselves can be expected in an immediate follow-up PR, based on the branch [syrie/fix_towire_mutation](https://github.com/SyrieBianco/thriftrw-go/tree/syrie/fix_towire_mutation)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Generate `Default_*` methods that construct Thrift structs with defined 
+  default values pre-populated.  
+
 ### Changed
 - gen: Redefine Options.Plugin as a struct usable outside go.uber.org/thriftrw.
 
--- a/gen/field.go
+++ b/gen/field.go
@@ -83,6 +83,10 @@ func (f fieldGroupGenerator) Generate(g Generator) error {
 		return err
 	}
 
+	if err := f.DefineDefaultConstructor(g); err != nil {
+		return err
+	}
+
 	if err := f.ToWire(g); err != nil {
 		return err
 	}
@@ -242,6 +246,35 @@ func (f *fieldGroupGenerator) declFieldName(fs *compile.FieldSpec) (string, erro
 		return "", fmt.Errorf("could not declare field %q%s: %v", name, note, err)
 	}
 	return name, nil
+}
+
+func (f fieldGroupGenerator) DefineDefaultConstructor(g Generator) error {
+	var hasDefaults bool
+	for _, f := range f.Fields {
+		if f.Default != nil {
+			hasDefaults = true
+			break
+		}
+	}
+	if !hasDefaults {
+		return nil
+	}
+	return g.DeclareFromTemplate(
+		`
+		// Default_<.Name> constructs a new <.Name> struct,
+		// pre-populating any fields with defined default values. 
+		func Default_<.Name>() *<.Name> {
+			<- $v := newVar "v" ->
+			var v <.Name>
+			<- range .Fields ->
+				<- $fname := goName . ->
+				<- if .Default>
+					<$v>.<$fname> = <constantValuePtr .Default .Type>
+				<- end ->
+			<end>
+			return &v
+		}
+		`, f, TemplateFunc("constantValuePtr", ConstantValuePtr))
 }
 
 func (f fieldGroupGenerator) ToWire(g Generator) error {
--- a/gen/quick_test.go
+++ b/gen/quick_test.go
@@ -58,7 +58,7 @@ func thriftTypeIsValid(v thriftType) bool {
 	// Validate method on generated types.
 
 	// We validate while serializing.
-	w, err := v.ToWire()
+	w, err := deepCopy(v).ToWire() // todo(syrie) remove deepCopy when mutation is fixed
 	if err != nil {
 		return false
 	}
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Generate `Default_*` methods that construct thrift structs with defined
```
abhinav, line 9: nit:

```suggestion
- Generate `Default_*` methods that construct Thrift structs with defined 
```

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Generate `Default_*` methods that construct thrift structs with defined 
+  default values already pre-populated.
```
abhinav, line 10: nit: redundant

```suggestion
  default values pre-populated.  
```

or

```suggestion
  default values already populated.  
```

gen/field.go
```diff
@@ -244,6 +248,37 @@ func (f *fieldGroupGenerator) declFieldName(fs *compile.FieldSpec) (string, erro
...
+	if !hasDefaults {
+		return nil
+	}
+	return g.DeclareFromTemplate(
+		`
+		// Default_<.Name> constructs a new <.Name> struct,
+		// pre-populating any fields with defined default values. 
+		func Default_<.Name>() *<.Name> {
+			<- $v := newVar "v" ->
+			v := &<.Name>{}
```
abhinav, line 268: nit: prefer `var` form for zero value

```suggestion
			var v <.Name>
```

gen/field.go
```diff
@@ -244,6 +248,37 @@ func (f *fieldGroupGenerator) declFieldName(fs *compile.FieldSpec) (string, erro
...
+
+			<range .Fields>
+				<- $fname := goName . ->
+				<- if .Default ->
+					<$v>.<$fname> = <constantValuePtr .Default .Type>
+				<end>
+			<end>
+
+			return v
```
abhinav, line 277: with var change above, this becomes a pointer:

```suggestion
			return &v
```

gen/field.go
```diff
@@ -244,6 +248,37 @@ func (f *fieldGroupGenerator) declFieldName(fs *compile.FieldSpec) (string, erro
...
+			<- $v := newVar "v" ->
+			v := &<.Name>{}
+
+			<range .Fields>
+				<- $fname := goName . ->
+				<- if .Default ->
+					<$v>.<$fname> = <constantValuePtr .Default .Type>
+				<end>
+			<end>
+
```
abhinav, line 276: I know it's just generated code, but I'd like to reduce the amount of unnecessary empty lines. 
We can do the following to get rid of the extraneous empty lines in the generated code.

```suggestion
			<- range .Fields ->
				<- $fname := goName . ->
				<- if .Default>
					<$v>.<$fname> = <constantValuePtr .Default .Type>
				<- end ->
			<end>
```

gen/quick_test.go
```diff
@@ -148,6 +148,34 @@ func enumValueGenerator(valuesFunc interface{}) func(*testing.T, *rand.Rand) thr
...
+	}
+	d := reflect.ValueOf(defaults).Elem()
+	g := reflect.ValueOf(give).Elem()
+
+	// copy values into a new struct, selecting default values from
+	// d when g values are nil
+	w := reflect.New(reflect.TypeOf(give).Elem()).Elem()
+	for i := 0; i < g.NumField(); i++ {
+		wantField := w.Field(i)
+		wantField.Set(g.Field(i))
```
abhinav, line 166: nit: it's less "want" and more "out". We should rename this to "outField" or similar.

gen/quick_test.go
```diff
@@ -148,6 +148,34 @@ func enumValueGenerator(valuesFunc interface{}) func(*testing.T, *rand.Rand) thr
...
+func copyWithDefaults(give, defaults thriftType) thriftType {
+	if defaults == nil {
+		return deepCopy(give)
+	}
+	d := reflect.ValueOf(defaults).Elem()
+	g := reflect.ValueOf(give).Elem()
+
+	// copy values into a new struct, selecting default values from
+	// d when g values are nil
+	w := reflect.New(reflect.TypeOf(give).Elem()).Elem()
```
abhinav, line 163: Can probably simplify to
```suggestion
	w := reflect.New(g.Type()).Elem()
```

gen/quick_test.go
```diff
@@ -856,13 +901,15 @@ func (q *quickSuite) testTypedefPrimitivePtr(t *testing.T, give thriftType) {
 }
 
 // Tests that each field of a struct has an accessor that returns the same
-// value as the field.
-func (q *quickSuite) testGetAccessors(t *testing.T, giveVal thriftType) {
+// value as the field. The wantVal accounts for any defaults we expect to
+// be set by the accessors.
```
abhinav, line 905: Worth expanding on this comment with what "accounts for" means.

gen/quick_test.go
```diff
@@ -944,3 +991,22 @@ func (q *quickSuite) testErrorName(t *testing.T) {
 
 	assert.NotPanics(t, func() { _ = v.ErrorName() })
 }
+
+// deepCopy copies the thriftType interface as a work-around until
+// mutation in the ToWire methods are removed.
+func deepCopy(v thriftType) thriftType {
+	c := reflect.New(reflect.TypeOf(v).Elem())
+
+	val := reflect.ValueOf(v).Elem()
```
abhinav, line 1000: minor simplification: can re-use `val` to get type instead of reflecting it again.

```suggestion
	val := reflect.ValueOf(v).Elem()
	c := reflect.New(val.Type())
```

gen/quick_test.go
SyrieBianco: +1

PR #446, opened 2020-06-16
Do not mutate receiver in ToWire call

SyrieBianco:
Avoid mutating the receiver with default values in `ToWire` calls
by changing the template to use a local variable to store defaults
rather than writing to the object. Also, fix tests accoringly.

*Note*: this change should be rolled out with caution, because,
as many of our own tests depended on this behavior (see #445),
it is possible that users might be broken by this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Generate `Default_*` methods that construct Thrift structs with defined 
   default values pre-populated.  
 
-### Changed
+### Changed  
 - gen: Redefine Options.Plugin as a struct usable outside go.uber.org/thriftrw.
+
+### Fixed
+- Serializing generated Thrift objects no longer mutates them with default
+  values for unspecified fields.
 
 ## [1.23.0] - 2020-03-31
 ### Added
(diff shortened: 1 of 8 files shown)


gen/quick_test.go
```diff
@@ -58,7 +58,7 @@ func thriftTypeIsValid(v thriftType) bool {
 	// Validate method on generated types.
 
 	// We validate while serializing.
-	w, err := deepCopy(v).ToWire() // todo(syrie) remove deepCopy when mutation is fixed
```
abhinav, line 61: should this PR delete the deepCopy function too?

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   default values pre-populated.  
 
 ### Changed
+- Generated `ToWire` methods no longer mutate the receiver with default values, 
+  leaving nil fields as nil on the object though serializing the defaults.
```
abhinav, line 14: I consider this a bugfix, honestly. How about,

> Serializing generated Thrift objects no longer mutates them with default values for unspecified fields.

SyrieBianco: 👌 👍

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   default values pre-populated.  
 
 ### Changed
+- Serializing generated Thrift objects no longer mutates them with default
```
abhinav, line 13: Consider moving this entry to `### Fixed` section

PR #449, opened 2020-09-01
Add RootModules to api.GenerateServiceResponse

r-hang:
Even if a Thrift file has no service declaration and therefore no
req.RootServices entries, it's stil useful to have Module information
for the Thrift Files that ThriftRW is invoked with.

To provide an example of where this Module information can be
useful, the Directory field of a Module can help place ThriftRW
plugin generated code for Thrift files that don't contain service
declarations next to where ThriftRW has already generated its code
for that module.

### DIFF
--- a/gen/generate_test.go
+++ b/gen/generate_test.go
@@ -21,10 +21,13 @@
 package gen
 
 import (
+	"bytes"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sort"
 	"testing"
 
 	"go.uber.org/thriftrw/compile"
@@ -251,6 +254,28 @@ func TestGenerate(t *testing.T) {
 			},
 		},
 		{
+			desc: "ServiceGenerator request contains RootModules info",
+			getPlugin: func(mockCtrl *gomock.Controller) plugin.Handle {
+				sgen := handletest.NewMockServiceGenerator(mockCtrl)
+				sgen.EXPECT().Generate(
+					newRootModulesMatcher(
+						[]api.ModuleID{1},
+					)).Return(&api.GenerateServiceResponse{
+					Files: map[string][]byte{
+						"foo.txt":    []byte("hello world\n"),
+						"bar/baz.go": []byte("package bar\n"),
+					},
+				}, nil)
+
+				handle := handletest.NewMockHandle(mockCtrl)
+				handle.EXPECT().ServiceGenerator().Return(sgen)
+				return handle
+			},
+			wantFiles: []string{
+				"foo/foo.go",
+			},
+		},
+		{
 			desc: "ServiceGenerator plugin conflict",
 			getPlugin: func(mockCtrl *gomock.Controller) plugin.Handle {
 				sgen := handletest.NewMockServiceGenerator(mockCtrl)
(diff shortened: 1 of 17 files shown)


gen/plugin_test.go
```diff
@@ -9,13 +9,33 @@ import (
...
+
+	g := newGenerateServiceBuilder(importer)
+
+	_, err := g.AddRootModule(spec.ThriftFile())
+	require.NoError(t, err)
+
+	assert.Len(t, g.rootModules, 1)
+}
+
 func TestAddRootService(t *testing.T) {
```
abhinav, line 39: should we rename this test?

gen/plugin_test.go
```diff
@@ -171,9 +195,14 @@ func TestAddRootService(t *testing.T) {
 		g := newGenerateServiceBuilder(importer)
```
abhinav, line 195: this is some old code; while we're here, would you mind making use of subtests here with t.Run?

```
for _, tt := range tests {
  t.Run(tt.desc, func(t *testing.T) {
    // ...
  })
}
```

abhinav (commented):
Should we do a more end-to-end test as well? We can call [`gen.Generate`](https://pkg.go.dev/go.uber.org/thriftrw/gen?tab=doc#Generate) on a testdata Thrift file that imports other files, with a [fake/mock plugin](https://pkg.go.dev/go.uber.org/thriftrw/gen?tab=doc#Options.Plugin), and test the GenerateServiceRequest received by it.

r-hang:
Closing PR to turn around comments and restructure code.

gen/plugin_test.go
r-hang: Done

gen/plugin_test.go
r-hang: Renamed to include RootModulesandServices to reflect additional functionality.

r-hang:
@abhinav, yes! I just added a more 'end2end-ish test within the structure gen/generate_test.go! The existing setup hand creates a compile.Module instead of parsing one directly from a test Thrift, but where a compile.Module comes from is not central to this added functionality so I don't feel parsing a test Thrift is a hard constraint.

I mock the plugin Generate interface and I've added a custom matcher to verify that the GenerateServiceResponse does contain the new RootModules field that i've added.

gen/generate_test.go
```diff
@@ -395,3 +422,58 @@ func TestThriftPackageImporter(t *testing.T) {
...
+	if !ok {
+		return false
+	}
+
+	var intsA []int
+	for _, mID := range s.req.RootModules {
+		intsA = append(intsA, int(mID))
+	}
+
+	var intsB []int
```
abhinav, line 449: if we're not matching other fields of GenerateServivceRequest, perhaps this should be `RootModulesMatcher`?

In fact, it can probably be `type rootModulesMatcher []api.ModuleID`.

gen/plugin.go
```diff
@@ -44,13 +44,17 @@ type generateServiceBuilder struct {
...
+	rootModules map[api.ModuleID]struct{}
+
 	// To ensure there are no duplicates
 	rootServices map[api.ServiceID]struct{}
 }
 
 func newGenerateServiceBuilder(i thriftPackageImporter) *generateServiceBuilder {
 	return &generateServiceBuilder{
 		GenerateServiceRequest: api.GenerateServiceRequest{
+			RootModules:   make([]api.ModuleID, 0),
```
abhinav, line 57: we should give it a default capacity of 1 since that's what'll usually go in there.

```suggestion
			RootModules:   make([]api.ModuleID, 0, 1),
```

abhinav (approved):
don't forget the changelog

PR #454, opened 2020-09-09
travis: Test against supported versions of Go

r-hang:
Update Travis CI jobs to use Go to stable (currently 1.15.x) and oldstable
(currently 1.14.x) to CI on supported versions of Go.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,8 @@ env:
 
 matrix:
   include:
-  - go: 1.12.x
-  - go: 1.13.x
+  - go: oldstable
+  - go: stable
     env: LINT=1
 
 # Download modules and install tools.


.travis.yml
```diff
@@ -8,8 +8,8 @@ env:
 
 matrix:
   include:
-  - go: 1.12.x
-  - go: 1.13.x
+  - go: 1.14.x
+  - go: 1.15.x
```
abhinav, line 12: we can use `stable` and `oldstable` here, which will always point to the currently supported versions.

https://github.com/travis-ci/gimme/releases/tag/v1.5.4

```suggestion
  - go: oldstable
  - go: stable
```

abhinav (approved):

PR #459, opened 2020-12-17
Bugfix: handle defaults for boolean fields

jronak:
Template evaluation of default field didn't work as expected for `nil` boolean value due to ```falsy if condition```. Fixed it by explicitly asserting default field is not not nil


AllenLuUber (approved):

abhinav (approved):
Good catch, thanks!
Don't worry about the missing coverage. It's for generated lines that literally cannot happen:

```
x, err := ..., error(nil)
if err != nil { .. }
```

PR #464, opened 2021-02-10
Remove inline typeconvert when logging some thrift typedefs

robbertvanginkel:
https://github.com/thriftrw/thriftrw-go/pull/366 added fast zap logging
by generating extra `MarshalLog...` methods on generated types. For
typedefs in thrift, other types that log them will cast to the
underlying type in their `MarshalLog...` method, causing the generated
code to need to import multiple go packages generated from both direct
and indirect thrift dependencies.

This should not be necessary, as typedefs of complex types (anything
that cannot be mapped directly to a go stdlb type) get their own
`MarshalLog...` methods, which cast to underlying type when necessary.

This PR changes the logging for typedef'ed fields to:
- cast to the underlying type inline when the underlying thrift type
directly maps to a go type.
- defer to the typedef's generated `MarshalLog...` method otherwise.

This results in having more predictable importpaths (eg knowing that one
thrift import will lead to one go import in generated code) makes
implementing good bazel rules for thriftrw easier.

See internal issue GO-362 for a more elaborate example of how the
current behaviour complicates our Bazel setup.

### DIFF
--- a/gen/typedef.go
+++ b/gen/typedef.go
@@ -114,20 +114,18 @@ func typedef(g Generator, spec *compile.TypedefSpec) error {
 		</* We want the behavior of the underlying type for typedefs: in the case that
 				they are objects or arrays, we need to cast to the underlying object or array;
 				else, zapMarshaler in zap.go will take care of it. */>
-		<if (eq (zapEncoder .Target) "Object") ->
+		<if (zapTypedefHasGeneratedMarshaler .Target) ->
 			<$zapcore := import "go.uber.org/zap/zapcore">
 			<$enc := newVar "enc">
+			<if (eq (zapEncoder .Target) "Object") ->
 			func (<$v> <$typedefType>) MarshalLogObject(<$enc> <$zapcore>.ObjectEncoder) error {
-				return (<zapMarshaler . $v>).MarshalLogObject(<$enc>)
+				return (<zapTypedefGenerateMarshaler . $v>).MarshalLogObject(<$enc>)
 			}
-		<- end>
-
-		<if (eq (zapEncoder .Target) "Array") ->
-			<$zapcore := import "go.uber.org/zap/zapcore">
-			<$enc := newVar "enc">
+			<- else if (eq (zapEncoder .Target) "Array") ->
 			func (<$v> <$typedefType>) MarshalLogArray(<$enc> <$zapcore>.ArrayEncoder) error {
-				return (<zapMarshaler . $v>).MarshalLogArray(<$enc>)
+				return (<zapTypedefGenerateMarshaler . $v>).MarshalLogArray(<$enc>)
 			}
+			<- end>
 		<- end>
 		<- end>
 		`,
--- a/gen/zap.go
+++ b/gen/zap.go
@@ -101,11 +101,11 @@ func (z *zapGenerator) zapTypedefHasGeneratedMarshaler(g Generator, spec compile
 // to the next value. This should be used when generating the code for the Zap
 // marshal implementation of the typedef.
 func (z *zapGenerator) zapTypedefGenerateMarshaler(g Generator, spec *compile.TypedefSpec, fieldValue string) (string, error) {
-	rootName, err := typeReference(g, spec.Target)
+	targetName, err := typeReference(g, spec.Target)
 	if err != nil {
 		return "", err
 	}
-	return z.zapMarshalerGenerator(g, spec, fmt.Sprintf("(%v)(%v)", rootName, fieldValue))
+	return z.zapMarshalerGenerator(g, spec, fmt.Sprintf("(%v)(%v)", targetName, fieldValue))
 }
 
 // zapMarshaler takes a TypeSpec, evaluates whether there are underlying elements
--- a/gen/zap.go
+++ b/gen/zap.go
@@ -86,6 +86,28 @@ func (z *zapGenerator) zapEncoder(g Generator, spec compile.TypeSpec) string {
 	panic(root)
 }
 
+// zapTypedefHasGeneratedMarshaler defines if a typedef will have generated code
+// zapcore.ObjectMarshaler or zapcore.ArrayMarshaler.
+func (z *zapGenerator) zapTypedefHasGeneratedMarshaler(g Generator, spec compile.TypeSpec) bool {
+	switch z.zapEncoder(g, spec) {
+	case "Array", "Object":
+		return true
+	default:
+		return false
+	}
+}
+
+// zapTypedefGenerateMarshaler is similar to zapMarshaler but always converts a typedef
+// to the next value. This should be used when generating the code for the Zap
+// marshal implementation of the typedef.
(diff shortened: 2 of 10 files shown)


gen/zap.go
```diff
@@ -86,6 +86,27 @@ func (z *zapGenerator) zapEncoder(g Generator, spec compile.TypeSpec) string {
 	panic(root)
 }
 
+// zapTypedefHasGeneratedMarshaler defines if a typedef will have generated code
+// zapcore.ObjectMarshaler or zapcore.ArrayMarshaler.
+func (z *zapGenerator) zapTypedefHasGeneratedMarshaler(g Generator, spec compile.TypeSpec) bool {
+	return z.zapEncoder(g, spec) == "Array" || z.zapEncoder(g, spec) == "Object"
```
abhinav, line 92: nit: can use switch here. Will also reduce number of zapEncoder calls.

```suggestion
	switch z.zapEncoder(g, spec) {
	case "Array", "Object":
		return true
	default:
		return false
	}
```

gen/zap.go
```diff
@@ -86,6 +86,27 @@ func (z *zapGenerator) zapEncoder(g Generator, spec compile.TypeSpec) string {
...
+}
+
+// zapTypedefGenerateMarshaler is similar to zapMarshaler but always converts a typedef
+// to its root value. This should be used when generating the code for the Zap
+// marshal implementation of the typedef.
+func (z *zapGenerator) zapTypedefGenerateMarshaler(g Generator, spec compile.TypeSpec, fieldValue string) (string, error) {
+	if _, ok := spec.(*compile.TypedefSpec); !ok {
+		panic(spec)
```
abhinav, line 100: Seeing as this panics by default anyway without the `!ok` form and we're not
providing any additional context, we can just let it panic on the cast and get
the coverage back.

```suggestion
	_ = spec.(*compile.TypedefSpec)
```

gen/zap.go
```diff
@@ -86,6 +86,27 @@ func (z *zapGenerator) zapEncoder(g Generator, spec compile.TypeSpec) string {
...
+// zapTypedefHasGeneratedMarshaler defines if a typedef will have generated code
+// zapcore.ObjectMarshaler or zapcore.ArrayMarshaler.
+func (z *zapGenerator) zapTypedefHasGeneratedMarshaler(g Generator, spec compile.TypeSpec) bool {
+	return z.zapEncoder(g, spec) == "Array" || z.zapEncoder(g, spec) == "Object"
+}
+
+// zapTypedefGenerateMarshaler is similar to zapMarshaler but always converts a typedef
+// to its root value. This should be used when generating the code for the Zap
+// marshal implementation of the typedef.
+func (z *zapGenerator) zapTypedefGenerateMarshaler(g Generator, spec compile.TypeSpec, fieldValue string) (string, error) {
```
abhinav, line 98: I'm not sure why we need this separate function exposed in the template? Looks
like zapMarshaler can (and does) already do this if spec is a
compile.TypedefSpec.

robbertvanginkel: This `zapMarshaler` is used in the template that generates `MarshalLogObject`/`MarshalLogArray` for typedef, as well as in the template for other object's `MarshalLog...` function if they have typedef for a field. In one case you need the underlying one, in the other one you need to typeconvert to the underlying to avoid the typedef's generated `MarshalLog...` to call itself.

To avoid adding adding an extra parameter to `zapMarshaler` (something like `forceTypedefToRoot bool`), I made/exported a new function.

gen/zap.go
```diff
@@ -86,6 +86,28 @@ func (z *zapGenerator) zapEncoder(g Generator, spec compile.TypeSpec) string {
...
+	switch z.zapEncoder(g, spec) {
+	case "Array", "Object":
+		return true
+	default:
+		return false
+	}
+}
+
+// zapTypedefGenerateMarshaler is similar to zapMarshaler but always converts a typedef
+// to the next value. This should be used when generating the code for the Zap
```
AllenLuUber, line 101: you mean ~~next~~ `root` value?

gen/typedef.go
```diff
@@ -114,20 +114,18 @@ func typedef(g Generator, spec *compile.TypedefSpec) error {
 		</* We want the behavior of the underlying type for typedefs: in the case that
 				they are objects or arrays, we need to cast to the underlying object or array;
 				else, zapMarshaler in zap.go will take care of it. */>
-		<if (eq (zapEncoder .Target) "Object") ->
+		<if (zapTypedefHasGeneratedMarshaler .Target) ->
```
AllenLuUber, line 117: optional: this change seems not necessary tho. We can keep as `<if (eq (zapEncoder .Target) "Object") ->` and same for the `else if`. But logically they are the same.

AllenLuUber (approved):

gen/zap.go
robbertvanginkel: In this case it's actually next (`spec.Target`) instead of root. Considering the types in `https://github.com/thriftrw/thriftrw-go/pull/464/files/bbdb77f0f1437cac94ff1619d508168e847cc015#diff-9634c2ea01f5b97b2d07789614e0893c96f71e0b34db720c2c884d2b14fe8ad2`, using root on the example `MyUUID` would cast it to go value for `i128`, which could still cause an unexpected import. Using next makes sure we use the go type for `UUID`.

gen/typedef.go
robbertvanginkel: You're right that this one isn't strictly necessary, but I wanted to avoid having loosly coupled if checks for `==Object || == "Array"` in `zapMarshaler` and this template over here, so it seemed like a good idea to introduce `zapTypedefHasGeneratedMarshaler`

gen/zap.go
AllenLuUber: ohh gotcha. next as next in the alias chain.

gen/typedef.go
AllenLuUber: cool. I'm fine either way. 🚢

gen/zap.go
```diff
@@ -86,6 +86,28 @@ func (z *zapGenerator) zapEncoder(g Generator, spec compile.TypeSpec) string {
...
+// marshal implementation of the typedef.
+func (z *zapGenerator) zapTypedefGenerateMarshaler(g Generator, spec *compile.TypedefSpec, fieldValue string) (string, error) {
+	rootName, err := typeReference(g, spec.Target)
+	if err != nil {
+		return "", err
+	}
+	return z.zapMarshalerGenerator(g, spec, fmt.Sprintf("(%v)(%v)", rootName, fieldValue))
```
abhinav, line 108: rootName is no longer referencing the root name.
```suggestion
	targetName, err := typeReference(g, spec.Target)
	if err != nil {
		return "", err
	}
	return z.zapMarshalerGenerator(g, spec, fmt.Sprintf("(%v)(%v)", targetName, fieldValue))
```

abhinav (commented):
Mostly LGTM.

To address coverage drop, check out [gen/quick_test.go](https://github.com/thriftrw/thriftrw-go/blob/dev/gen/quick_test.go). We need to add newly declared types there.

abhinav (approved):

Issue #465, opened 2021-02-16
Fix possible misuse of reflect.SliceHeader vet error


robbertvanginkel:
A new check was added to vet in go1.16 (https://github.com/golang/go/issues/40701), lint now fails with:
```
wire/unsafe.go:38:35: possible misuse of reflect.SliceHeader
```
https://github.com/thriftrw/thriftrw-go/blob/b7d2a80efe388226633db82edcbb3b832d67be92/wire/unsafe.go#L28-L39
Currently failing https://github.com/thriftrw/thriftrw-go/pull/464. Should fix and rebase.

PR #466, opened 2021-02-17
Rewrite unsafeStringToBytes to satisfy 1.16 vet check

robbertvanginkel:
Fixes https://github.com/thriftrw/thriftrw-go/issues/465.

According to https://golang.org/pkg/unsafe/#Pointer (6), one of the only safe ways to use pointers:
> Conversion of a reflect.SliceHeader or reflect.StringHeader Data field to or from Pointer.
...
> In general, reflect.SliceHeader and reflect.StringHeader should be used only as *reflect.SliceHeader and *reflect.StringHeader pointing at actual slices or strings, never as plain structs. A program should not declare or allocate variables of these struct types.

This rewrites the unsafe to only modify to not allocate a `reflect.SliceHeader ` manually making the vet check happy.

### DIFF
--- a/wire/unsafe.go
+++ b/wire/unsafe.go
@@ -29,13 +29,14 @@ import (
 // new memory for it with the assumption that the resulting byte slice will not
 // be mutated.
 func unsafeStringToBytes(s string) []byte {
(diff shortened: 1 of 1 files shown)


abhinav (approved):
Thanks for fixing!

PR #469, opened 2021-04-02
[WIP] Encode and decode structs lazily

abhinav:
@kanads found out that some services were spending a lot of time
allocating and growing slices of `[]wire.Field` during deserialization.
Specifically, [on this line][1], where we append to the slice, we spend
a fair bit of CPU growing the slice.

    fields = append(fields, wire.Field{ID: fid, Value: val})

@kanads is experimenting with pre-allocating these slices to some sane
value in production. That's a good short-term fix, but the long-term
fix is likely to address this TODO in that function:

    // TODO(abg) add a lazy FieldList type instead of []Field.

[1]: https://github.com/thriftrw/thriftrw-go/blob/9c16419a8cced33850d40e7620e5dfe96e7b1a4f/protocol/binary/reader.go#L304

This change adds a lazy wire.FieldList type that behaves similarly to
the wire.ValueList and wire.MapItemList types.

    type FieldList interface {
        ForEach(func(Field) error) error
        Close()
    }

Using this, during deserialization we can return an implementation that
doesn't deserialize from the struct until ForEach is called.

---

I'm not completely sold on this yet because of nested structs: for
nested structs, we'll end up skipping over them several times. However,
this might still end up being more efficient than the many allocations
we make. Benchmarks pending.

```
name                                        old time/op    new time/op    delta
RoundTrip/PrimitiveOptionalStruct/Encode-4    3.14µs ±22%    2.21µs ±12%   -29.72%  (p=0.008 n=5+5)
RoundTrip/PrimitiveOptionalStruct/Decode-4    4.79µs ±17%    3.38µs ±19%   -29.49%  (p=0.016 n=5+5)
RoundTrip/Graph/Encode-4                      7.16µs ±23%    3.37µs ±28%   -52.92%  (p=0.008 n=5+5)
RoundTrip/Graph/Decode-4                      9.59µs ±16%    8.64µs ±11%      ~     (p=0.222 n=5+5)
RoundTrip/ContainersOfContainers/Encode-4     34.5µs ±11%    27.9µs ±10%   -19.11%  (p=0.008 n=5+5)
RoundTrip/ContainersOfContainers/Decode-4     80.6µs ±16%    78.6µs ±33%      ~     (p=0.548 n=5+5)

name                                        old alloc/op   new alloc/op   delta
RoundTrip/PrimitiveOptionalStruct/Encode-4      704B ± 0%        0B       -100.00%  (p=0.008 n=5+5)
[...]

PR #471, opened 2021-04-05
[WIP] Pre-allocate slices to prevent resizing at runtime

kanads:
It was noticed that some services were spending a lot of time calling `runtime.growSlice` on the `wire.Field` slice. Additionally, most of the time on these slice resize calls were to `runtime.mallocgc`, and not to `runtime.memmove`. This hints the extra calls are commonly due to resizing smaller slice lengths.

This patch pre-allocates the slice to optimize the common case (until PR #469 is merged). Did the same in the case of reading large fields in `readBytes`.

Benchmark results:
```
name                                           old time/op    new time/op    delta
_readStruct/small_struct_with_64_fields-12     8.19µs ± 2%    6.86µs ± 4%    -16.16%  (p=0.008 n=5+5)
_readStruct/large_struct_with_512_fields-12    65.6µs ± 3%    57.2µs ± 1%    -12.74%  (p=0.008 n=5+5)

name                                           old alloc/op   new alloc/op   delta
_readStruct/small_struct_with_64_fields-12     12.0kB ± 0%    12.3kB ± 0%     +2.81%  (p=0.008 n=5+5)
_readStruct/large_struct_with_512_fields-12    98.0kB ± 0%    86.0kB ± 0%    -12.20%  (p=0.008 n=5+5)

name                                           old allocs/op  new allocs/op  delta
_readStruct/small_struct_with_64_fields-12       7.00 ± 0%      1.00 ± 0%    -85.71%  (p=0.008 n=5+5)
_readStruct/large_struct_with_512_fields-12      10.0 ± 0%       3.0 ± 0%    -70.00%  (p=0.008 n=5+5)
```

### DIFF
--- a/protocol/binary/reader.go
+++ b/protocol/binary/reader.go
@@ -309,6 +309,11 @@ func (br *Reader) readStruct(off int64) (wire.Struct, int64, error) {
 			return wire.Struct{}, off, err
 		}
 	}
+
+	if len(fields) == 0 {
+		// Don't let unused slice escape.
+		return wire.Struct{}, off, err
+	}
 	return wire.Struct{Fields: fields}, off, err
 }
 
--- a/protocol/binary/reader.go
+++ b/protocol/binary/reader.go
@@ -279,7 +279,8 @@ func (br *Reader) readString(off int64) (string, int64, error) {
 }
 
 func (br *Reader) readStruct(off int64) (wire.Struct, int64, error) {
-	var fields []wire.Field
+	// Pre-allocate slice to prevent frequesnt resizing at shorter slice lengths.
+	fields := make([]wire.Field, 0, 128)
 	// TODO(abg) add a lazy FieldList type instead of []Field.
 
 	typ, off, err := br.readByte(off)
--- a/protocol/binary/reader_benchmark_test.go
+++ b/protocol/binary/reader_benchmark_test.go
@@ -9,22 +9,19 @@ import (
 )
 
 func Benchmark_readStruct(b *testing.B) {
-	for _, bench := range []struct {
+	benchmarks := []struct {
 		name      string
 		numFields int
 	}{
-		{
-			name:      "small struct with 64 fields",
-			numFields: 64,
-		},
-		{
-			name:      "large struct with 512 fields",
-			numFields: 512,
-		},
-	} {
-		reader := NewReader(constructThriftStruct(bench.numFields))
+		{name: "small struct with 64 fields", numFields: 64},
+		{name: "large struct with 512 fields", numFields: 512},
+	}
 
-		b.Run(bench.name, func(b *testing.B) {
+	for _, bb := range benchmarks {
+		b.Run(bb.name, func(b *testing.B) {
+			reader := NewReader(constructThriftStruct(bb.numFields))
+			b.ResetTimer()
+
 			for i := 0; i < b.N; i++ {
 				reader.readStruct(0)
 			}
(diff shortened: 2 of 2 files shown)


protocol/binary/reader.go
```diff
@@ -261,6 +261,8 @@ func (br *Reader) readBytes(off int64) ([]byte, int64, error) {
 	// bytesAllocThreshold. We don't want bad requests to lock the system up.
 	if length > bytesAllocThreshold {
 		var buff bytes.Buffer
+		buff.Grow(int(length))
```
abhinav, line 264: See comment right above this block.
This is a slow path specifically for requests larger than 1MB.
Requests smaller than 1MB will be pre-allocated as desired.

protocol/binary/reader_benchmark_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"testing"
+
+	"go.uber.org/thriftrw/wire"
+)
+
+func Benchmark_readStruct(b *testing.B) {
+	for _, bench := range []struct {
+		name      string
+		numFields int
+	}{
```
abhinav, line 15: stylistic nit: For table tests and benchmarks, please separate the test cases from the body, and place an empty line between them:

```
benchmarks := []struct{
}{
  {
    name: "...",
    numFields: ...,
  },
  {
    // ...
  },
}

for _, bb := range benchmarks {
    b.Run(bb.name, func(b *testing.B) {
        // ...
    })
}
```

Merging the table into the for loops makes it difficult to easily scan where test cases end and where test body begins.

protocol/binary/reader_benchmark_test.go
```diff
@@ -0,0 +1,48 @@
...
+		},
+		{
+			name:      "large struct with 512 fields",
+			numFields: 512,
+		},
+	} {
+		reader := NewReader(constructThriftStruct(bench.numFields))
+
+		b.Run(bench.name, func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
```
abhinav, line 28: When using subtests/subbenchmarks, setup specific to that benchmark/test should normally be inside the new function.

You can call b.ResetTimer to not have `NewReader(..)` count against the performance:
```suggestion
		b.Run(bench.name, func(b *testing.B) {
			reader := NewReader(constructThriftStruct(bench.numFields))
			b.ResetTImer()
			for i := 0; i < b.N; i++ {
```

protocol/binary/reader_benchmark_test.go
```diff
@@ -0,0 +1,48 @@
...
+			}
+		})
+	}
+}
+
+// constructThriftStruct constructs a message containing a struct with n bool fields.
+func constructThriftStruct(n int) io.ReaderAt {
+	var buf bytes.Buffer
+	buf.Grow(n*4 + 1)
+	boolField := []byte{byte(wire.TBool), 0, 0, 0}
```
abhinav, line 39: The fields should typically have different field IDs.
I think this function would be more readable if we generated a wire.Struct manually and used binary.Writer to serialize it.

abhinav (changes requested):

protocol/binary/reader.go
abhinav: Not resolved. Comment still valid.

kanads:
Fixed. Thanks.

abhinav:
Per discussion with @kanads, this ended up not actually helping. We're exploring other options.

PR #473, opened 2021-05-18
Add AST/IDL support for the 'cpp_include' header

jparise:
The CppInclude (`cpp_include`) header is a C++-specific header. From
https://thrift.apache.org/docs/idl#c-include:

    A C++ include adds a custom C++ include to the output of the C++
    code generator for this Thrift document.

This change allows thriftrw to parse Thrift IDL files that use this
header for C++ code generation targets. Otherwise, the parser fails with
an error like: `syntax error: unexpected IDENTIFIER`

For completeness, this change also adds a `CppInclude` AST node (rather
than just ignoring the parsed characters).

Because this is a C++-specific header, it is otherwise unused by the
higher-level Go-specific packages.


abhinav (approved):
Sounds reasonable. Thanks!
I added a changelog entry for this.
We'll merge this shortly.

PR #477, opened 2021-05-19
idl: Handle escape sequences in string literals

jparise:
We pass string literals through strconv.Unquote(), but we need to
perform a little more preprocessing to handle escape quote characters.
Specifically, we need to "unescape" the "opposite" quote characters to
the kind of string we're about to "unquote". That is, for a single-
quoted strings, we need to unescape `\"` sequences, and for double-quoted
strings, we unescape `\'` sequences.

### DIFF
--- a/idl/internal/lex.rl
+++ b/idl/internal/lex.rl
@@ -308,7 +308,7 @@ func (lex *lexer) Lex(out *yySymType) int {
                 if len(bs) > 0 && bs[0] == '\'' {
                     str, err = UnquoteSingleQuoted(bs)
                 } else {
-                    str, err = strconv.Unquote(string(bs))
+                    str, err = UnquoteDoubleQuoted(bs)
                 }
 
                 if err != nil {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 - `double` constants with exponents but without decimal components are now supported.
+- Fix handling of escaped quotes inside string literals.
 
 ## [1.26.0] - 2021-02-18
 ### Changed
--- a/idl/internal/lex.go
+++ b/idl/internal/lex.go
@@ -932,7 +932,7 @@ func (lex *lexer) Lex(out *yySymType) int {
 			if len(bs) > 0 && bs[0] == '\'' {
 				str, err = UnquoteSingleQuoted(bs)
 			} else {
-				str, err = strconv.Unquote(string(bs))
+				str, err = UnquoteDoubleQuoted(bs)
 			}
 
 			if err != nil {
--- a/idl/internal/quote.go
+++ b/idl/internal/quote.go
@@ -20,14 +20,17 @@
 
 package internal
 
-import "strconv"
+import (
+	"bytes"
+	"strconv"
+)
 
 // UnquoteSingleQuoted unquotes a slice of bytes representing a single quoted
 // string.
 //
 // 	UnquoteSingleQuoted([]byte("'foo'")) == "foo"
 func UnquoteSingleQuoted(in []byte) (string, error) {
-	out := string(swapQuotes(in))
+	out := string(swapQuotes(unescapeQuotes(in, '"')))
 	str, err := strconv.Unquote(out)
 	if err != nil {
 		return str, err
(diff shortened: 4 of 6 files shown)


jparise:
Here is the related code from the Apache Thrift compiler for reference: https://github.com/apache/thrift/blob/8101302c568bf7631448f11f7a5f6a22287d5cbe/compiler/cpp/src/thrift/thriftl.ll#L312-L337

idl/internal/lex.rl
```diff
@@ -301,10 +302,14 @@ func (lex *lexer) Lex(out *yySymType) int {
 
                 var str string
                 var err error
-                if len(bs) > 0 && bs[0] == '\'' {
-                    str, err = UnquoteSingleQuoted(bs)
-                } else {
-                    str, err = strconv.Unquote(string(bs))
+
+                str, err = Unescape(bs)
+                if err == nil && !strings.ContainsAny(str, "\r\n") {
```
jparise, line 307: I added this new condition to prevent strings containing these control characters from being passed to `strconv.Unquote()`, which will always error in these cases.

This is a behavior change. If that's not desired, we have two options:

1. Roll our own `Unquote()` that can handle these cases.
2. Stop expanding `\r` and `\n` in `Unescape()`, meaning we'd only have partial support for escape sequences.

abhinav: For (2), would it be partial support given that strconv.Unquote will still expand `\r`, `\n`, `\t`, etc.?
Our issue seems to be that strconv.Unquote doesn't handle `\'`, so really, that's all we need to unescape.

abhinav (commented):
Thanks for the contribution! It took me a second to understand the exact issue, but it's this: ThriftRW doesn't handle escaping of the opposite kind of quote inside a string.

That is, the following fail:

```
const string s1 = '\"foo\"'
const string s2 = "\'bar\'"
```

These both turn into the following because of the implementation of UnquoteSingleQuoted:

```
"\'foo\'"
"\'bar\'"
```

So really, the issue is that strconv.Unquote doesn't handle `\'`, so we should be able to handle that in Unescape and let strconv.Unquote do the rest, right? strconv.Unquote will handle `\r`, `\n`, and `\t` just fine (https://play.golang.org/p/8HJ_bvsTsn0).

jparise:
> So really, the issue is that strconv.Unquote doesn't handle `\'`, so we should be able to handle that in Unescape and let strconv.Unquote do the rest, right? strconv.Unquote will handle `\r`, `\n`, and `\t` just fine ([play.golang.org/p/8HJ_bvsTsn0](https://play.golang.org/p/8HJ_bvsTsn0)).

I don't think `strconv.Unquote` handles all of those escape sequences. In your linked example, the raw string contains the literal control characters (i.e. `\n` or `EOL`), which `strconv.Unquote` passes along without modification. However, in the case we need to support in the parser, we're receiving the two-character literal `[]byte{'\', \n'}`, which we need to "expand" into the control character equivalent (which is what `Unescape` does).

It appears that `strconv.Unquote` _does_ handle `\t` (https://play.golang.org/p/-BnEzUUqMJT), although supporting that sequence in `Unescape` isn't a burden given we already need to process the other characters.

abhinav:
Sorry, I must be misunderstanding something. The string literal in my example contains the bytes `[]byte{'\\', 'n'}`, not `[]byte{'\n'}`. (I used backticks to construct the string literal.) Here's a version that prints the characters separately: https://play.golang.org/p/o-Pa8gyHx45.

jparise:
> Sorry, I must be misunderstanding something. The string literal in my example contains the bytes `[]byte{'\\', 'n'}`, not `[]byte{'\n'}`. (I used backticks to construct the string literal.) Here's a version that prints the characters separately: [play.golang.org/p/o-Pa8gyHx45](https://play.golang.org/p/o-Pa8gyHx45).

You're right. I had my wires crossed. I'll work on narrowing this down now that I have a (slightly) better understanding of `strconv.Unquote`'s behavior. I think we're going to end up with just an `UnescapeQuotes` first pass followed by the existing `strconv.Unquote` behavior.

jparise:
The "trick" ending up being unescaping the "opposite" quote character based on the overall quoting of the string. I cleaned up the latest version of the diff and added some additional test cases.

abhinav (approved):
Looks great. Thanks!

abhinav:
@jparise Thanks for these fixes!

Do you need a release tagged soon or are you okay waiting?

jparise:
> @jparise Thanks for these fixes!

You're welcome! Our internal IDL repository makes a good Thrift test suite. 😉 

> Do you need a release tagged soon or are you okay waiting?

If you're able to make a release soon, that would be appreciated. I'm also fine pointing to a commit in the meantime.

PR #478, opened 2021-05-19
idl: decimal components of doubles are optional

jparise:
Doubles can optionally have decimals and/or exponents. Adjust the
`double` pattern to recognize constants like `1e-5`.


abhinav (approved):
Nice. Thanks!

PR #479, opened 2021-05-19
Release v1.27.0

abhinav:
This releases v1.27.0 with the following changes:

- Support parsing Thrift files that contain a `cpp_include` headers
- Support double constants without decimal components
- Handle escaped quotes inside string literals

Thanks to @jparise for these fixes.

I'm considering this a minor release, even though we could make a case for all
of these changes being bugfixes, largely because it contains changes to the
parser which is ever so slightly riskier than a usual bugfix.


mh-park (approved):
LGTM other than the coverage drops failing the checks, but they look pretty minor.

PR #481, opened 2021-05-20
idl: support for auto-assigned field identifiers

jparise:
Thrift allows implicit assignment of field identifiers. This is legacy
syntax, but it's useful to support it for compatibility with old files.

Auto-assigned fields start at -1 and get increasingly negative. If a
negative field identifier is explicitly given, it will be used as the
new basis for any additional auto-assigned fields. This behavior is
consistent with the Apache Thrift implementation.

### DIFF
--- a/ast/definition.go
+++ b/ast/definition.go
@@ -297,7 +297,8 @@ const (
 // 	3: i64 baz (go.name = "qux")
 //
 type Field struct {
-	ID           int
+	ID int
+	// IDUnset indicates that a field identifier wasn't provided.
 	IDUnset      bool
 	Name         string
 	Type         Type
--- a/ast/definition.go
+++ b/ast/definition.go
@@ -298,6 +298,7 @@ const (
 //
 type Field struct {
 	ID           int
+	IDUnset      bool
 	Name         string
 	Type         Type
 	Requiredness Requiredness
--- a/compile/field.go
+++ b/compile/field.go
@@ -85,9 +85,13 @@ func (r fieldRequiredness) isRequired(src *ast.Field) (bool, error) {
 //
 // disallowDefaultValue specifies whether the field is allowed to have a default
 // value.
+//
+// allowNegativeIDs controls whether negative field identifiers are allowed.
+// This also enables auto-assigning (negative) values for unset field IDs.
 type fieldOptions struct {
 	requiredness         fieldRequiredness
 	disallowDefaultValue bool
+	allowNegativeIDs     bool
 }
 
 // FieldSpec represents a single field of a struct or parameter list.
@@ -103,7 +107,7 @@ type FieldSpec struct {
 
 // compileField compiles the given Field source into a FieldSpec.
 func compileField(src *ast.Field, options fieldOptions) (*FieldSpec, error) {
-	if src.ID < 1 || src.ID > math.MaxInt16 {
+	if (src.ID < 1 && !options.allowNegativeIDs) || src.ID > math.MaxInt16 {
 		return nil, fieldIDOutOfBoundsError{ID: src.ID, Name: src.Name}
 	}
 
@@ -177,6 +181,7 @@ type FieldGroup []*FieldSpec
 func compileFields(src []*ast.Field, options fieldOptions) (FieldGroup, error) {
 	fieldsNS := newNamespace(caseSensitive)
 	usedIDs := make(map[int16]string)
+	nextNegativeID := -1
 
 	fields := make([]*FieldSpec, 0, len(src))
 	for _, astField := range src {
@@ -185,6 +190,15 @@ func compileFields(src []*ast.Field, options fieldOptions) (FieldGroup, error) {
 				Target: astField.Name,
 				Line:   astField.Line,
 				Reason: err,
+			}
+		}
+
+		if options.allowNegativeIDs {
+			if astField.ID < 1 {
+				nextNegativeID = astField.ID - 1
+			} else if astField.IDUnset {
+				astField.ID = nextNegativeID
+				nextNegativeID--
 			}
 		}
 
--- a/compile/field.go
+++ b/compile/field.go
@@ -194,7 +194,7 @@ func compileFields(src []*ast.Field, options fieldOptions) (FieldGroup, error) {
 		}
 
 		if options.allowNegativeIDs {
-			if astField.ID < 1 {
+			if astField.ID < 0 {
 				nextNegativeID = astField.ID - 1
 			} else if astField.IDUnset {
 				astField.ID = nextNegativeID
--- a/idl/internal/thrift.y
+++ b/idl/internal/thrift.y
@@ -24,6 +24,7 @@ import "go.uber.org/thriftrw/ast"
     fieldType ast.Type
     structType ast.StructureType
     baseTypeID ast.BaseTypeID
+    fieldIdentifier int
     fieldRequired ast.Requiredness
 
     field *ast.Field
@@ -64,6 +65,7 @@ import "go.uber.org/thriftrw/ast"
 %type <prog> program
 %type <fieldType> type
 %type <baseTypeID> base_type_name
+%type <fieldIdentifier> field_identifier
 %type <fieldRequired> field_required
 %type <structType> struct_type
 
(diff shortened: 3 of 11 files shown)


jparise:
This legacy compatibility comes with a (small-ish) complexity cost in the parser, but I think it's manageable to enable greater compatibility with existing Thrift files.

abhinav:
Hey @jparise, thanks for the contribution(s)!

For this one, I'd like to add some context on ThriftRW first.

We originally aimed for ThriftRW to limit the kinds of Thrift files it would accept to align more with our high-level internal Thrift style guide. Among the requirements were:

1. `optional` or `required` must be specified for all `struct` and `exception` fields —we found inconsistent behavior between languages and implementations for the default optional/required
2. field identifiers must always be specified

This is true today: ThriftRW will refuse to generate code for a file that doesn't have optional/required specifiers on all struct and exception fields. Our Thrift linter has enforced these requirements on all Thrift files almost since we first adopted Thrift, so by the time we wrote ThriftRW, this wasn't a problem.

(We added support for loosening (1) in the compiler layer with the [`NonStrict` option](https://pkg.go.dev/go.uber.org/thriftrw/compile#NonStrict) for tools like [yab](https://github.com/yarpc/yab) that need to accept more Thrift files, but we haven't yet exposed this in the CLI.)

With that background, about this change:
I think we would be okay with exposing support for legacy files as an opt-in flag rather than the default—I'd like our Thrift files to remain healthy per the linter's definition. How would you feel about a `--legacy` flag (or a different name) that opts into this behavior? It can also flip `NonStrict` in the compiler because I suspect you'll run into that too once you start trying to generate code.

Separately, about the implementation:
I don't love that we have to set the implicit identifiers at parse time rather than compile time. There's some loss of information of exactly what the user wrote in their Thrift file if we go this route. So one couldn't, for example, write a Thrift file reformatting tool that parses the IDL and reproduces the original (mostly).

We've tried to do "interpretation" of the AST in the compile layer. Check out [EnumItem](https://pkg.go.dev/go.uber.org/thriftrw/ast#EnumItem), for example, which records whether an enum item had an explicit value specified or not. Then the [compile layer assigns the implicit value](https://github.com/thriftrw/thriftrw-go/blob/v1.27.0/compile/enum.go#L46).

Ideally, we could do something like that here, but we're limited by the fact that [ast.Field](https://pkg.go.dev/go.uber.org/thriftrw/ast#Field) has a non-pointer ID field so we can't differentiate between "ID is missing" and "ID is zero". There are couple options here (in order of preference):

1. Deprecate `ID int` in favor of `Identifier *int` (or better name) -- set both for backwards compatibility, with ID taking zero for unset. Have the compiler generate the automatic ID.
2. Add an `IDUnset bool` field to `ast.Field`. This field indicates whether the ID was set. If unset, the compiler generates the automatic ID.
3. Continue generating the ID at parse time and add an `AutoID bool` field to `ast.Field` that indicates that the ID was auto-assigned, not user-specified.

(1) and(2) have the advantage that the compiler would implement the `--legacy` behavior, so the parser can remain simple-ish. With (3), the parser would have to be aware of the `--legacy` flag and enforce it.

What do you think?

jparise:
That all makes good sense to me. I'm actually using `thriftrw/ast` and `thriftrw/idl` to build our own Thrift linter, and for that use case, it's useful to be able to parse the full Thrift grammar even if some input forms will be rejected by higher-level rules.

Given that, I also prefer option (1) or (2). (1) would be more "correct" because it would match what you're already doing for `EnumItem`, but (2) is just as explicit and wouldn't be as disruptive.

compile/field.go
```diff
@@ -103,7 +107,7 @@ type FieldSpec struct {
 
 // compileField compiles the given Field source into a FieldSpec.
 func compileField(src *ast.Field, options fieldOptions) (*FieldSpec, error) {
-	if src.ID < 1 || src.ID > math.MaxInt16 {
+	if (src.ID < 1 && !options.allowNegativeIDs) || src.ID > math.MaxInt16 {
```
jparise, line 110: This is a small misnomer because `0` is/was considered "negative" here.

idl/internal/thrift.y
```diff
@@ -268,38 +270,65 @@ enum_item
...
+                IDUnset: true,
+                Name: $5,
+                Type: $4,
+                Requiredness: $3,
+                Annotations: $6,
                 Line: $1,
                 Doc: ParseDocstring($2),
             }
         }
+    | lineno docstring field_identifier field_required type IDENTIFIER '=' const_value type_annotations
```
jparise, line 302: I think we could reduce this from four cases to two by introducing a `default_value` pattern that just returns `nil` when `= const_value` isn't there. I didn't want to make that change in the middle of the rest of these changes, though.

jparise:
The latest commit implements Option 2 (by adding `IDUnset` to `ast.Field`) along with a new `fieldOptions` flag to control negative values and auto-assignment in the compiler.

I didn't see an existing pattern where `fieldOptions` are tested, so some of the new compiler paths are untested. If the overall approach looks promising, we can add test coverage for those cases.

ast/definition.go
```diff
@@ -298,6 +298,7 @@ const (
 //
 type Field struct {
 	ID           int
+	IDUnset      bool
```
abhinav, line 301: Since this one is unique, it might be worth adding a quick comment here.

```suggestion
	// IDUnset specifies that a field identifier was not
	// provided for this field.
	IDUnset      bool
```

idl/internal/thrift.y
```diff
@@ -268,38 +270,65 @@ enum_item
...
+                Default: $7,
+                Annotations: $8,
+                Line: $1,
+                Doc: ParseDocstring($2),
+            }
+        }
+    ;
+
+field_identifier
+    : INTCONSTANT ':' { $$ = int($1) }
```
abhinav, line 331: I'm not sure if we can do this, but would it makes sense to have this handle the empty case and return both, the field ID and whether it's unset? Something like this:

```
type fieldIdentifier struct {
  ID int
  Unset bool
}
```

And then,

```suggestion
field_identifier
    : INTCONSTANT ':' { $$ = fieldIdentifier{ID: int($1)} }
    | /* na */        { $$ = fieldIdentifier{Unset: true} }
```

That might let us half the number of cases above as well.

```
$$ = &ast.Field{
  // ...,
  ID: $3.ID,
  IDUnset: $3.Unset,
}
```

idl/internal/thrift.y
abhinav: Yeah, that's a reasonable change and it's okay to leave it out of this.
I'm happy to do that in a follow up to this once this lands if you don't have the time to.

compile/field.go
```diff
@@ -85,9 +85,13 @@ func (r fieldRequiredness) isRequired(src *ast.Field) (bool, error) {
 //
 // disallowDefaultValue specifies whether the field is allowed to have a default
 // value.
+//
+// allowNegativeIDs controls whether negative field identifiers are allowed.
+// This also enables auto-assigning (negative) values for unset field IDs.
 type fieldOptions struct {
 	requiredness         fieldRequiredness
 	disallowDefaultValue bool
+	allowNegativeIDs     bool
```
abhinav, line 94: We'll need to expose the public Option for this inside [compile/options.go](https://github.com/thriftrw/thriftrw-go/blob/dev/compile/options.go), which will set it on the `*compiler`, and that'll plumb it to fieldOptions for compileStruct.

abhinav (commented):
The approach makes sense to me. Thanks!

The simplification you suggest for default_value: that seems like it could apply to field_identifier too?

idl/internal/thrift.y
jparise: I didn't realize this syntax would be so straightforward. Thanks!

compile/field.go
jparise: Would you like these to be part of the existing `NonStrict` option's behavior, or as a new (to be named) option?

abhinav: The existing NonStrict flag should do. Thanks!

(Apologies for missing this yesterday.)

jparise:
I think I have everything covered now, but let me know if you have any remaining feedback.

abhinav (approved):
LGTM!

Thanks for addressing all the comments and the test cases.

PR #486, opened 2021-06-03
Move Stream-based interfaces into their own package

witriew:
The implementation for these `Stream`-based interfaces will likely lie in the `binary` package under the protocol directory.  Because of this, the `binary` package will need to import `protocol` causing a circular dependency.

Move these interfaces into a separate package so that `binary`'s import can happen without issue.


abhinav (approved):
Would you mind adding more details about the motivation for this to the commit message before we merge? (The whole thing with cycles, specifically.)

PR #487, opened 2021-06-04
Make Streaming interfaces private to allow for safe experimentation

dianale31:
The streaming Protocol/Reader/Writer interfaces are made "private" to allow for experimentation before merging onto dev. 

This PR also updates the streamdev branch to include the latest commit from thriftrw-go, where the streaming interfaces were moved into their own package.


internal/interface/interface.go
```diff
@@ -0,0 +1,7 @@
+package _interface
```
witriew, line 1: I've not seen this pattern with package names before, should this just be `interface`?

Also, maybe provide the copyright boilerplate and a docstring briefly talking about the package, since it's the only file here?

dianale31: I used `iface` since interface is a keyword.  Also, good catch on copywrite boilerplate.

internal/interface/interface.go
```diff
@@ -0,0 +1,45 @@
+package iface
```
witriew, line 1: nit: Let's also just name the directory/file to `iface` as well.  The boilerplate should also probably go above any declarations.

PR #488, opened 2021-06-04
Make Streaming interfaces private to allow for safe experimentation

dianale31:
The new streaming Reader/Writer/Protocol interfaces will be experimented on in the `streamdev` branch. The interfaces will be made private for now to disallow custom implementations.

### DIFF
--- a/internal/iface/iface.go
+++ b/internal/iface/iface.go
@@ -1,0 +1,47 @@
+// Copyright (c) 2021 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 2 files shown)


internal/interface/interface.go
```diff
@@ -0,0 +1,46 @@
+package iface
```
witriew, line 1: nit: Should the directory/filename also match this package name now that it's shortened?

The boilerplate should also go before any declarations.

internal/iface/iface.go
```diff
@@ -0,0 +1,46 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package iface
+
+// Interface will be used to make interfaces defined in ThriftRW private.
+// Example usage:
+//
+// import ".../internal/interface"
```
witriew, line 26: very nitty: Change example to match new renames.

witriew (approved):

protocol/stream/stream.go
```diff
@@ -25,12 +25,15 @@ package stream
 import (
 	"io"
 
+	internal "go.uber.org/thriftrw/internal/iface"
```
abhinav, line 28: We should avoid named import/aliasing if there isn't a conflict.

See also https://github.com/uber-go/guide/blob/master/style.md#import-aliasing

abhinav (commented):
just nits

abhinav (approved):

PR #490, opened 2021-06-08
Add streaming Writer implementation

dianale31:
This implementation extends the existing Binary writer struct to include the new streaming methods. Some refactoring is done in the main `writer.go` file to reduce common code. 

Tests have also been added to test for correctness of the new streaming methods, but we'll need to write extensive tests for code generation, where we will actually assert that container types have the proper elements.

### DIFF
--- a/protocol/binary/stream_writer.go
+++ b/protocol/binary/stream_writer.go
@@ -1,0 +1,233 @@
+// Copyright (c) 2021 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package binary
+
+import (
+	"io"
+	"math"
+	"sync"
+
+	"go.uber.org/thriftrw/internal/iface"
+	"go.uber.org/thriftrw/protocol/stream"
+)
+
+var streamWriterPool = sync.Pool{
+	New: func() interface{} {
+		return &StreamWriter{}
+	}}
+
+// StreamWriter implements basic logic for writing the Thrift Binary Protocol
+// to an io.Writer.
+type StreamWriter struct {
+	// Private implementation to satisfy the private Writer interface.
+	iface.Impl
+
+	writer io.Writer
+
+	// This buffer is re-used every time we need a slice of up to 8 bytes.
+	buffer [8]byte
+}
+
+// BorrowStreamWriter fetches a StreamWriter from the system that will write
+// its output to the given io.Writer.
+//
+// This StreamWriter must be returned back using ReturnStreamWriter.
+func BorrowStreamWriter(w io.Writer) *StreamWriter {
+	streamWriter := streamWriterPool.Get().(*StreamWriter)
+	streamWriter.writer = w
+	return streamWriter
+}
+
+// ReturnStreamWriter returns a previously borrowed StreamWriter back to the
+// system.
+func ReturnStreamWriter(sw *StreamWriter) {
+	sw.writer = nil
+	streamWriterPool.Put(sw)
+}
+
+func (sw *StreamWriter) write(bs []byte) error {
+	_, err := sw.writer.Write(bs)
+	return err
+}
+
+func (sw *StreamWriter) writeByte(b byte) error {
+	bs := sw.buffer[0:1]
+	bs[0] = b
+	return sw.write(bs)
+}
+
+// WriteBool encodes a boolean
+func (sw *StreamWriter) WriteBool(b bool) error {
+	if b {
+		return sw.writeByte(1)
+	}
+	return sw.writeByte(0)
+}
+
+// WriteInt8 encodes an int8
+func (sw *StreamWriter) WriteInt8(i int8) error {
+	return sw.writeByte(byte(i))
+}
+
+// WriteInt16 encodes an int16
+func (sw *StreamWriter) WriteInt16(i int16) error {
+	bs := sw.buffer[0:2]
+	bigEndian.PutUint16(bs, uint16(i))
+	return sw.write(bs)
+}
+
+// WriteInt32 encodes an int32
+func (sw *StreamWriter) WriteInt32(i int32) error {
+	bs := sw.buffer[0:4]
+	bigEndian.PutUint32(bs, uint32(i))
+	return sw.write(bs)
+}
+
+// WriteInt64 encodes an int64
+func (sw *StreamWriter) WriteInt64(i int64) error {
+	bs := sw.buffer[0:8]
+	bigEndian.PutUint64(bs, uint64(i))
+	return sw.write(bs)
+}
+
+// WriteString encodes a string
+func (sw *StreamWriter) WriteString(s string) error {
+	if err := sw.WriteInt32(int32(len(s))); err != nil {
+		return err
+	}
+
+	_, err := io.WriteString(sw.writer, s)
+	return err
+}
+
+// WriteDouble encodes a double
+func (sw *StreamWriter) WriteDouble(d float64) error {
+	value := math.Float64bits(d)
+	return sw.WriteInt64(int64(value))
+}
+
+// WriteBinary encodes binary
+func (sw *StreamWriter) WriteBinary(b []byte) error {
+	if err := sw.WriteInt32(int32(len(b))); err != nil {
+		return err
+	}
+	return sw.write(b)
+}
+
+// WriteFieldBegin marks the beginning of a new field in a struct. The first
+// byte denotes the type and the next two bytes denote the field id.
+func (sw *StreamWriter) WriteFieldBegin(f stream.FieldHeader) error {
+	// type:1
+	if err := sw.writeByte(byte(f.Type)); err != nil {
+		return err
+	}
+
+	// id:2
+	if err := sw.WriteInt16(f.ID); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// WriteFieldEnd denotes the end of a field. No-op.
+func (sw *StreamWriter) WriteFieldEnd() error {
+	return nil
+}
+
+// WriteStructBegin denotes the beginning of a struct. No-op.
+func (sw *StreamWriter) WriteStructBegin() error {
+	return nil
+}
+
+// WriteStructEnd uses the zero byte to mark the end of a struct.
+func (sw *StreamWriter) WriteStructEnd() error {
+	return sw.writeByte(0) // end struct
+}
+
+// WriteListBegin marks the beginning of a new list. The first byte denotes
+// the type of the items and the next four bytes denote the length of the list.
+func (sw *StreamWriter) WriteListBegin(l stream.ListHeader) error {
+	// vtype:1
+	if err := sw.writeByte(byte(l.Type)); err != nil {
+		return err
+	}
+
+	// length:4
+	if err := sw.WriteInt32(int32(l.Length)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// WriteListEnd marks the end of a list. No-op.
+func (sw *StreamWriter) WriteListEnd() error {
+	return nil
+}
+
+// WriteSetBegin marks the beginning of a new set. The first byte denotes
+// the type of the items and the next four bytes denote the length of the set.
+func (sw *StreamWriter) WriteSetBegin(s stream.SetHeader) error {
+	// vtype:1
+	if err := sw.writeByte(byte(s.Type)); err != nil {
+		return err
+	}
+
+	// length:4
+	if err := sw.WriteInt32(int32(s.Length)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// WriteSetEnd marks the end of a set. No-op.
+func (sw *StreamWriter) WriteSetEnd() error {
+	return nil
+}
+
+// WriteMapBegin marks the beginning of a new map. The first byte denotes
+// the type of the keys, the second byte denotes the type of the values,
+// and the next four bytes denote the length of the map.
+func (sw *StreamWriter) WriteMapBegin(m stream.MapHeader) error {
+	// ktype:1
+	if err := sw.writeByte(byte(m.KeyType)); err != nil {
+		return err
+	}
+
+	// vtype:1
+	if err := sw.writeByte(byte(m.ValueType)); err != nil {
+		return err
+	}
+
+	// length:4
+	if err := sw.WriteInt32(int32(m.Length)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// WriteMapEnd marks the end of a map. No-op.
+func (sw *StreamWriter) WriteMapEnd() error {
+	return nil
+}
--- a/protocol/binary/writer.go
+++ b/protocol/binary/writer.go
@@ -23,26 +23,24 @@ package binary
 import (
 	"fmt"
 	"io"
-	"math"
 	"sync"
 
+	"go.uber.org/thriftrw/protocol/stream"
 	"go.uber.org/thriftrw/wire"
 )
 
-var writerPool = sync.Pool{New: func() interface{} {
-	writer := &Writer{}
-	writer.writeValue = writer.WriteValue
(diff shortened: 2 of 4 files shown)


protocol/binary/stream_writer.go
```diff
@@ -0,0 +1,174 @@
...
+
+// WriteString encodes a string
+func (bw *Writer) WriteString(s string) error {
+	return bw.writeString(s)
+}
+
+// WriteDouble encodes a double
+func (bw *Writer) WriteDouble(d float64) error {
+	value := math.Float64bits(d)
+	return bw.writeInt64(int64(value))
```
witriew, line 64: Could make this an unexported `writeDouble` that's also used by
```
case wire.TDouble:
```
since it's the same logic.

protocol/binary_test.go
```diff
@@ -132,6 +133,30 @@ func TestBoolEOFFailure(t *testing.T) {
...
+
+		// Encode with Streaming protocol
+		w := binary.BorrowWriter(&buff)
+		err := w.WriteBool(test.value)
+		assert.NoError(t, err)
+		binary.ReturnWriter(w)
+
+		// Decode with wire representation and assert that the value is the same
+		val, err := Binary.Decode(bytes.NewReader(buff.Bytes()), wire.TBool)
+		assert.NoError(t, err)
```
witriew, line 155: `assert.NoError`'s like these should probably be from `require`.  The thought here is that `val` must be set (it could be `nil`) before continuing since any usage following this could get a "nil pointer exception".

witriew (commented):
We'll probably be able to improve upon test coverage once both the `stream.Writer` and the `stream.Reader` land, then we can repeat tests similar to those done with "encode <-> decode".

protocol/value_test.go
```diff
@@ -76,3 +81,166 @@ func vmap(kt, vt wire.Type, items ...wire.MapItem) wire.Value {
...
+// testing.
+type testValue struct {
+	val wire.Value
+}
+
+func testvalue(val wire.Value) testValue {
+	return testValue{val: val}
+}
+
+func (v testValue) streamEncodeStruct(sw *binary.StreamWriter) error {
```
dianale31, line 95: I couldn't add these stream methods directly onto the `wire.Value` object because it would have caused an import cycle. Wrapped it here and made it explicit that this is only for testing since objects using the wire.Value intermediate representation should continue to use the binary.Encode and binary.Decode methods.

protocol/binary_test.go
```diff
@@ -152,6 +184,22 @@ func TestI8EOFFailure(t *testing.T) {
...
 }
 
+func TestStreamI8(t *testing.T) {
+	tests := []streamEncodeDecodeTest{
+		{"0", testvalue(vi8(0)), []byte{0x00}},
+		{"1", testvalue(vi8(1)), []byte{0x01}},
+		{"-1", testvalue(vi8(-1)), []byte{0xff}},
+		{"127", testvalue(vi8(127)), []byte{0x7f}},
+		{"-128", testvalue(vi8(-128)), []byte{0x80}},
+	}
```
witriew, line 194: Instead of duplicating the test data, is it possible to add the `description` to `encodeDecodeStruct` and call `testvalue` upon passing `test.value` into `checkStreamEncode`?

Alternatively, was there no way to use `wire.Value.Get()`?

protocol/binary_test.go
```diff
@@ -68,6 +75,18 @@ func checkEncodeDecode(t *testing.T, typ wire.Type, tests []encodeDecodeTest) {
 	}
 }
 
+func checkStreamEncode(t *testing.T, val testValue, expected []byte) {
+	var buff bytes.Buffer
+
+	// Encode with Streaming protocol
+	w := binary.BorrowStreamWriter(&buff)
+	err := val.StreamEncode(w)
+	require.NoError(t, err)
```
witriew, line 84: The pattern under `StreamEncode` feels a bit weird...

Should a `*testing.T` be plumbed through?  Since the error checking is done here, the test failure would mark this line as errant making debugging the test difficult (it's unclear exactly where the error is since all of the tested calls are hidden away).

Alternatively, could you pass the specific function call in, e.g., `w.WriteInt32` to `checkStreamEncode` and execute that way?  This should be okay since a new `BorrowStreamWriter` is declared for every sub-test execution.

protocol/binary_test.go
```diff
@@ -23,6 +23,7 @@ package protocol
 import (
 	"bytes"
 	"fmt"
+	"go.uber.org/thriftrw/protocol/stream"
```
witriew, line 26: nit: import order

protocol/value_test.go
```diff
@@ -20,7 +20,12 @@
 
 package protocol
 
-import "go.uber.org/thriftrw/wire"
+import (
+	"fmt"
+	"go.uber.org/thriftrw/protocol/binary"
+	"go.uber.org/thriftrw/protocol/stream"
+	"go.uber.org/thriftrw/wire"
```
witriew, line 27: nit: import order

protocol/binary/stream_writer.go
```diff
@@ -0,0 +1,250 @@
...
+)
+
+var streamWriterPool = sync.Pool{New: func() interface{} {
+	writer := &StreamWriter{}
+	return writer
+}}
+
+// StreamWriter implements basic logic for writing the Thrift Binary Protocol
+// to an io.Writer.
+type StreamWriter struct {
```
abhinav, line 39: There's a tonne of duplication between this and binary.Writer.
Can we make Writer use StreamWriter? Like,

```go
type Writer struct {
  sw *StreamWriter
  // ...
}

func (w *Writer) writeStruct(s wire.Struct) error {
  w.sw.WriteStructBegin()
  // ...
  w.sw.WriteStructEnd()
}
```

That way we're not implementing this twice.

protocol/binary_test.go
abhinav: goimports should fix this grouping

protocol/value_test.go
```diff
@@ -76,3 +81,166 @@ func vmap(kt, vt wire.Type, items ...wire.MapItem) wire.Value {
 func vitem(k, v wire.Value) wire.MapItem {
 	return wire.MapItem{Key: k, Value: v}
 }
+
+// testValue wraps around wire.Value to allow for stream encoding/decoding
+// testing.
+type testValue struct {
```
abhinav, line 87: This seems like binary.Writer implemented in terms of binary.StreamWriter. We should just do that in the actual implementation and then we don't necessarily need this.

abhinav (commented):
This is a good first stab. I'd like to reduce the duplication in the implementation and the tests.
We should be able to reuse most of the existing test cases to exercise both code paths, especially if binary.Writer is implemented in terms of binary.StreamWriter.

protocol/binary/stream_writer.go
```diff
@@ -0,0 +1,250 @@
...
+func (bw *StreamWriter) write(bs []byte) error {
+	_, err := bw.writer.Write(bs)
+	return err
+}
+
+func (bw *StreamWriter) writeByte(b byte) error {
+	bs := bw.buffer[0:1]
+	bs[0] = b
+	return bw.write(bs)
+}
```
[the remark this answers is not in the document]
dianale31: Embedded the StreamWriter within Writer to avoid deduplication!

protocol/binary_test.go
dianale31: Decide to remove StreamEncode for testing in general. Once you embed the StreamWriter into Writer, you test the code for the streaming implementation anyway.

protocol/binary/stream_writer.go
```diff
@@ -0,0 +1,251 @@
...
+// system.
+func ReturnStreamWriter(sw *StreamWriter) {
+	sw.writer = nil
+	streamWriterPool.Put(sw)
+}
+
+func (bw *StreamWriter) write(bs []byte) error {
+	_, err := bw.writer.Write(bs)
+	return err
+}
```
witriew, line 71: nit: Have consistent var name for the `StreamWriter` - recommend using `sw` as the shorthand?

protocol/binary/writer.go
```diff
@@ -201,32 +168,32 @@ func (bw *Writer) WriteValue(v wire.Value) error {
 	switch v.Type() {
 	case wire.TBool:
 		if v.GetBool() {
-			return bw.writeByte(1)
+			return bw.sw.writeByte(1)
 		}
-		return bw.writeByte(0)
+		return bw.sw.writeByte(0)
```
witriew, line 173: This should be able to leverage`bw.sw.WriteBool` now, I think.

protocol/binary/stream_writer.go
```diff
@@ -0,0 +1,251 @@
...
+
+// WriteInt8 encodes an int8
+func (bw *StreamWriter) WriteInt8(i int8) error {
+	return bw.writeByte(byte(i))
+}
+
+// WriteInt16 encodes an int16
+func (bw *StreamWriter) WriteInt16(i int16) error {
+	return bw.writeInt16(i)
+}
```
witriew, line 122: Do we need these exported wrappers?  Could we export the underlying write instead, the `Writer` would then call the exported methods?

protocol/binary_test.go
```diff
@@ -29,15 +29,17 @@ import (
...
 	"go.uber.org/thriftrw/wire"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 type encodeDecodeTest struct {
-	value   wire.Value
-	encoded []byte
+	description string
```
witriew, line 40: We can probably plumb this through now in `checkEncodeDecode` with a `t.Run`.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- idl.Parse() now returns structured ParseError upon error.
```
abhinav, line 9: It looks like two commits that are on dev made it into this PR?

protocol/binary/writer.go
```diff
@@ -59,63 +58,29 @@ type Writer struct {
 //
 // This Writer must be returned back using ReturnWriter.
 func BorrowWriter(w io.Writer) *Writer {
+	streamWriter := streamWriterPool.Get().(*StreamWriter)
+	streamWriter.writer = w
```
abhinav, line 62: Should this just BorrowStreamWriter instead of accessing the pool directly?

```suggestion
	streamWriter := BorrowStreamWriter(w)
```

protocol/binary/writer.go
```diff
@@ -59,63 +58,29 @@ type Writer struct {
...
+	writer.sw = streamWriter
 	return writer
 }
 
 // ReturnWriter returns a previously borrowed Writer back to the system.
 func ReturnWriter(w *Writer) {
-	w.writer = nil
+	w.sw.writer = nil
+	streamWriterPool.Put(w.sw)
+	w.sw = nil
```
abhinav, line 72: We should maybe use ReturnStreamWriter here. And for safety, we should do that after setting sw to nil because that drops the last reference we have to it.

```suggestion
	sw := w.sw
	w.sw = nil
	ReturnStreamWriter(sw)
	writerPool.Put(w)
```

protocol/binary/writer.go
```diff
@@ -59,63 +58,29 @@ type Writer struct {
...
 func (bw *Writer) writeField(f wire.Field) error {
 	// type:1
-	if err := bw.writeByte(byte(f.Value.Type())); err != nil {
+	if err := bw.sw.writeByte(byte(f.Value.Type())); err != nil {
 		return err
 	}
 
 	// id:2
-	if err := bw.writeInt16(f.ID); err != nil {
+	if err := bw.sw.writeInt16(f.ID); err != nil {
```
abhinav, line 83: We should use WriteFieldBegin and WriteFieldEnd here.

```
if err := bw.sw.WriteFieldBegin(FieldHeader{...}); err != nil { return err }

if err := bw.WriteValue(f.Value); err != nil { ... }

if err := bw.sw.WriteFieldEnd(); err != nil { return err }
```

protocol/binary/writer.go
```diff
@@ -138,7 +103,7 @@ func (bw *Writer) writeStruct(s wire.Struct) error {
 			return err
 		}
 	}
-	return bw.writeByte(0) // end struct
+	return bw.sw.writeByte(0) // end struct
```
abhinav, line 106: Struct should use WriteStructBegin and WriteStructEnd.

protocol/binary/writer.go
abhinav: +1

    return bw.sw.WriteBool(v.GetBool())

protocol/binary/writer.go
```diff
@@ -201,32 +168,32 @@ func (bw *Writer) WriteValue(v wire.Value) error {
...
 		if v.GetBool() {
-			return bw.writeByte(1)
+			return bw.sw.writeByte(1)
 		}
-		return bw.writeByte(0)
+		return bw.sw.writeByte(0)
 
 	case wire.TI8:
-		return bw.writeByte(byte(v.GetI8()))
+		return bw.sw.writeByte(byte(v.GetI8()))
```
abhinav, line 176: This should use bw.sw.WriteInt8

```suggestion
		return bw.sw.WriteInt8(v.GetI8())
```

protocol/binary/writer.go
```diff
@@ -201,32 +168,32 @@ func (bw *Writer) WriteValue(v wire.Value) error {
...
 
 	case wire.TDouble:
 		value := math.Float64bits(v.GetDouble())
-		return bw.writeInt64(int64(value))
+		return bw.sw.writeInt64(int64(value))
```
abhinav, line 180: This should be bw.sw.WriteDouble

protocol/binary/writer.go
```diff
@@ -201,32 +168,32 @@ func (bw *Writer) WriteValue(v wire.Value) error {
...
 
 	case wire.TI16:
-		return bw.writeInt16(v.GetI16())
+		return bw.sw.writeInt16(v.GetI16())
```
abhinav, line 183: We should avoid using private functionality of sw here. This can be implemented solely in terms of the exported streaming API.


```suggestion
		return bw.sw.WriteInt16(v.GetI16())
```

protocol/binary/writer.go
```diff
@@ -201,32 +168,32 @@ func (bw *Writer) WriteValue(v wire.Value) error {
...
 
 	case wire.TI32:
-		return bw.writeInt32(v.GetI32())
+		return bw.sw.writeInt32(v.GetI32())
```
abhinav, line 186:
```suggestion
		return bw.sw.WriteInt32(v.GetI32())
```

protocol/binary/writer.go
```diff
@@ -201,32 +168,32 @@ func (bw *Writer) WriteValue(v wire.Value) error {
...
 
 	case wire.TI64:
-		return bw.writeInt64(v.GetI64())
+		return bw.sw.writeInt64(v.GetI64())
```
abhinav, line 189:
```suggestion
		return bw.sw.WriteInt64(v.GetI64())
```

protocol/binary/writer.go
```diff
@@ -201,32 +168,32 @@ func (bw *Writer) WriteValue(v wire.Value) error {
...
 
 	case wire.TBinary:
 		b := v.GetBinary()
-		if err := bw.writeInt32(int32(len(b))); err != nil {
+		if err := bw.sw.writeInt32(int32(len(b))); err != nil {
 			return err
 		}
-		return bw.write(b)
+		return bw.sw.write(b)
```
abhinav, line 196: This should let WriteBinary do the work.

```go
return bw.sw.WriteBinary(v.GetBinary())
```

protocol/binary_test.go
```diff
@@ -376,6 +379,22 @@ func TestStruct(t *testing.T) {
 	checkEncodeDecode(t, wire.TStruct, tests)
 }
 
+func TestStructBeginAndEndEncode(t *testing.T) {
+	var streamBuff bytes.Buffer
+	var err error
```
abhinav, line 384: nit: please group these:
```suggestion
	var (
		streamBuff bytes.Buffer
		err error
	)
```

(https://github.com/uber-go/guide/blob/master/style.md#group-similar-declarations)

protocol/binary_test.go
```diff
@@ -430,6 +450,24 @@ func TestMap(t *testing.T) {
 	checkEncodeDecode(t, wire.TMap, tests)
 }
 
+func TestMapBeginEncode(t *testing.T) {
+	var streamBuff bytes.Buffer
+	var err error
```
abhinav, line 455:
```suggestion
	var (
		streamBuff bytes.Buffer
		err error
	)
```

abhinav (commented):
Thanks for updating! This is generally looking great, except:

- Looks like a couple commits from dev unintentionally ended up on this PR
- There are a couple other spots in binary.Writer that it should use the exported StreamWriter methods rather than unexported
- I left some optional minor stylistic comments for the tests

protocol/binary_test.go
```diff
@@ -376,6 +381,24 @@ func TestStruct(t *testing.T) {
...
+		streamBuff bytes.Buffer
+		err        error
+	)
+
+	// Encode with Streaming protocol
+	w := binary.BorrowStreamWriter(&streamBuff)
+	err = w.WriteStructBegin()
+	assert.NoError(t, err)
+	err = w.WriteStructEnd()
+	assert.NoError(t, err)
```
abhinav, line 395: optional: can simplify:
```suggestion
	require.NoError(t, w.WriteStructBegin())
	require.NoError(t, w.WriteStructEnd())
```

protocol/binary_test.go
```diff
@@ -452,16 +496,33 @@ func TestMapEOFFailure(t *testing.T) {
...
 			[]byte{0x02, 0x00, 0x00, 0x00, 0x03, 0x01, 0x00, 0x01},
 		},
 	}
 
 	checkEncodeDecode(t, wire.TSet, tests)
 }
 
+func TestSetBeginEncode(t *testing.T) {
+	var streamBuff bytes.Buffer
+	var err error
```
abhinav, line 511: optional nit: can group
```suggestion
	var (
		streamBuff bytes.Buffer
		err error
	)
```

abhinav (approved):
LGTM! Thanks for making the changes.

PR #491, opened 2021-06-08
Add a "StreamReader" and plumb it for use in the standard "binary.Reader"

witriew:
Adds a "StreamReader", implementing a "stream.Reader", and plumbs it into
"binary.Protocol" allowing it to be used as a "stream.Protocol".

Also de-dupe a lot of code that was in "binary.Reader" by using the
"StreamReader" for a lot of its underlying implementation.

### DIFF
--- a/protocol/binary.go
+++ b/protocol/binary.go
@@ -24,13 +24,18 @@ import (
 	"fmt"
 	"io"
 
+	"go.uber.org/thriftrw/internal/iface"
 	"go.uber.org/thriftrw/protocol/binary"
+	"go.uber.org/thriftrw/protocol/stream"
 	"go.uber.org/thriftrw/wire"
 )
 
 // Binary implements the Thrift Binary Protocol.
 // Binary can be cast up to EnvelopeAgnosticProtocol to support DecodeRequest.
 var Binary Protocol
+
+// BinaryStreamer implements a streaming version of the Thrift Binary Protocol.
+var BinaryStreamer stream.Protocol
 
 // EnvelopeAgnosticBinary implements the Thrift Binary Protocol, using
 // DecodeRequest for request bodies that may or may not have an envelope.
@@ -104,10 +109,13 @@ func (e errUnexpectedEnvelopeType) Error() string {
 
 func init() {
 	Binary = binaryProtocol{}
+	BinaryStreamer = binaryProtocol{}
 	EnvelopeAgnosticBinary = binaryProtocol{}
 }
 
-type binaryProtocol struct{}
+type binaryProtocol struct {
+	iface.Private
+}
 
 func (binaryProtocol) Encode(v wire.Value, w io.Writer) error {
 	writer := binary.BorrowWriter(w)
@@ -133,6 +141,15 @@ func (binaryProtocol) DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error) {
 	reader := binary.NewReader(r)
 	e, err := reader.ReadEnveloped()
 	return e, err
+}
+
+func (binaryProtocol) Writer(w io.Writer) stream.Writer {
+	return binary.BorrowStreamWriter(w)
+}
+
+func (binaryProtocol) Reader(r io.Reader) stream.Reader {
+	sReader := binary.NewStreamReader(r)
+	return &sReader
 }
 
 // DecodeRequest specializes Decode and replaces DecodeEnveloped for the
--- a/protocol/binary/envelope.go
+++ b/protocol/binary/envelope.go
@@ -89,11 +89,12 @@ func (bw *Writer) WriteLegacyEnveloped(e wire.Envelope) error {
 // version numbers such that the value will always be negative.
 func (bw *Reader) ReadEnveloped() (wire.Envelope, error) {
 	var e wire.Envelope
-	initial, off, err := bw.readInt32(0)
+	val, off, err := bw.ReadValue(wire.TI32, 0)
 	if err != nil {
 		return wire.Envelope{}, err
 	}
 
+	initial := val.GetI32()
 	if initial > 0 {
 		e, off, err = bw.readNonStrictNameType()
 	} else {
@@ -103,7 +104,8 @@ func (bw *Reader) ReadEnveloped() (wire.Envelope, error) {
 		return e, err
 	}
 
-	e.SeqID, off, err = bw.readInt32(off)
+	val, off, err = bw.ReadValue(wire.TI32, off)
+	e.SeqID = val.GetI32()
 	if err != nil {
 		return e, err
 	}
@@ -126,25 +128,25 @@ func (bw *Reader) readStrictNameType(initial int32, off int64) (wire.Envelope, i
 	// This will truncate the bits that are not required.
 	e.Type = wire.EnvelopeType(initial)
 
-	var err error
-	e.Name, off, err = bw.readString(off)
+	name, off, err := bw.ReadValue(wire.TBinary, off)
+	e.Name = name.GetString()
 	return e, off, err
 }
 
 func (bw *Reader) readNonStrictNameType() (wire.Envelope, int64, error) {
 	var e wire.Envelope
 
-	name, off, err := bw.readString(0)
+	name, off, err := bw.ReadValue(wire.TBinary, 0)
 	if err != nil {
 		return e, off, err
 	}
-	e.Name = name
+	e.Name = name.GetString()
 
-	typeID, off, err := bw.readByte(off)
+	val, off, err := bw.ReadValue(wire.TI8, off)
 	if err != nil {
 		return e, off, err
 	}
-	e.Type = wire.EnvelopeType(typeID)
+	e.Type = wire.EnvelopeType(val.GetI8())
 
 	return e, off, nil
 }
--- a/protocol/binary/reader.go
+++ b/protocol/binary/reader.go
@@ -21,24 +21,220 @@
 package binary
 
 import (
-	"bytes"
 	"io"
-	"math"
 
 	"go.uber.org/thriftrw/wire"
 )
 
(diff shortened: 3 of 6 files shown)


protocol/binary/stream_reader.go
```diff
@@ -0,0 +1,417 @@
...
+
+// ReadMapEnd reads the "end" of a Thrift encoded list.  Since there is no
+// encoding for the end of a map, this is a noop.
+func (sr *StreamReader) ReadMapEnd() error {
+	return nil
+}
+
+// Skip skips fully over the provided Thrift type.
+func (sr *StreamReader) Skip(t wire.Type) error {
+	if w := fixedWidth(t); w > 0 {
```
dianale31, line 302: maybe just add a comment that `fixedWidth` allows you to skip over primitives?

protocol/binary_test.go
```diff
@@ -68,11 +68,23 @@ func checkEncodeDecode(t *testing.T, typ wire.Type, tests []encodeDecodeTest) {
 	}
 }
 
-type failureTest []byte
+func getStreamReader(t *testing.T, encoded []byte) stream.Reader {
```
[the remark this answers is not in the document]
witriew: Good catch, thanks!

Integrated this into the existing `TestStructs` test.

protocol/binary/stream_reader.go
```diff
@@ -0,0 +1,419 @@
...
+
+// StreamReader provides an implementation of a "stream.Reader".
+type StreamReader struct {
+	iface.Private
+
+	reader io.Reader
+	buffer [8]byte
+}
+
+// NewStreamReader returns a new StreamReader.
```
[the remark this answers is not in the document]
witriew: Cool, thanks for the review!

Made that modification as a separate commit on top of this one, detailing the perf impacts as well.

protocol/binary/reader.go
```diff
@@ -21,465 +21,232 @@
...
+	return Reader{reader: r}
+}
+
+// ReadValue reads a value off the given type off the wire starting at the
+// given offset.
+//
+// Returns the Value, the new offset, and an error if there was a decode error.
+func (br *Reader) ReadValue(t wire.Type, off int64) (wire.Value, int64, error) {
+	realReader := newReader(br.reader)
+	return realReader.ReadValue(t, off)
```
abhinav, line 251: nit: naming: realReader vs just reader? Or inline the variable?
```suggestion
	return newReader(br.reader).ReadValue(t, off)
```

protocol/binary/stream_reader.go
```diff
@@ -0,0 +1,446 @@
...
+	}
+
+	return err
+}
+
+// ReadBool reads a Thrift encoded bool value, returning a bool.
+func (sr *StreamReader) ReadBool() (bool, error) {
+	bs := sr.buffer[0:1]
+	_, err := sr.read(bs)
+	if err != nil {
```
abhinav, line 103: nit: can inline
```suggestion
	if _, err := sr.read(bs); err != nil {
```

protocol/binary/reader.go
```diff
@@ -21,465 +21,232 @@
...
-	if err != nil {
-		return off, err
-	}
-	vt := wire.Type(vtByte)
+func (r *reader) readStructStream() (wire.Struct, error) {
+	var fields []wire.Field
 
-	count, off, err := br.readInt32(off)
+	r.sr.ReadStructBegin()
+	fh, hasVal, err := r.sr.ReadFieldBegin()
```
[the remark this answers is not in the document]
witriew: Done - wasn't sure if `ok` was a good use here, but good to know that it is.

protocol/binary/reader.go
```diff
@@ -21,465 +21,232 @@
...
-	vtByte, off, err := br.readByte(off)
-	if err != nil {
-		return off, err
-	}
-	vt := wire.Type(vtByte)
+func (r *reader) readStructStream() (wire.Struct, error) {
+	var fields []wire.Field
 
-	count, off, err := br.readInt32(off)
+	r.sr.ReadStructBegin()
```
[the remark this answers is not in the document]
witriew: Not technically, but since it returns some `error` type, wrapped it with a check.

protocol/binary/reader.go
```diff
@@ -21,465 +21,232 @@
...
-		return off + w, nil
-	}
-
-	switch t {
-	case wire.TBinary:
-		length, off, err := br.readInt32(off)
-		if err != nil {
-			return off, err
+		fields = append(fields, wire.Field{ID: fh.ID, Value: val})
+		if err = r.sr.ReadFieldEnd(); err != nil {
```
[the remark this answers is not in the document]
witriew: Done - did my best to replace these across the diff too.

protocol/binary/reader.go
witriew: Inlined.

protocol/binary.go
```diff
@@ -135,6 +142,16 @@ func (binaryProtocol) DecodeEnveloped(r io.ReaderAt) (wire.Envelope, error) {
...
 }
 
+// stub
+func (binaryProtocol) Writer(w io.Writer) stream.Writer {
+	return nil
+}
+
+func (binaryProtocol) Reader(r io.Reader) stream.Reader {
+	sReader := binary.NewStreamReader(r)
+	return &sReader
```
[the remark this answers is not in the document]
witriew: Yeah, I'd like to tackle this in a separate PR, if possible.  There's a lot of movement in this PR that other diffs are waiting on so finishing up this one for now would be great.

protocol/binary/stream_reader.go
```diff
@@ -0,0 +1,445 @@
...
+
+func (sr *StreamReader) read(bs []byte) (int, error) {
+	n, err := sr.reader.Read(bs)
+
+	// For consistency with the non-streaming Reader, return an
+	// io.ErrUnexpectedEOF if number of bytes read is smaller than what was
+	// expected, https://golang.org/pkg/io/#ReaderAt
+	if err == io.EOF || n < len(bs) {
+		// All EOFs are unexpected when streaming
+		err = io.ErrUnexpectedEOF
```
abhinav, line 83: Should we use https://golang.org/pkg/io/#ReadFull here or make our own version of it with a `for` loop?
Otherwise this won't handle non-EOF partial reads correctly.

protocol/binary/stream_reader.go
```diff
@@ -0,0 +1,445 @@
...
+		return false, err
+	}
+
+	switch bs[0] {
+	case 0:
+		return false, nil
+	case 1:
+		return true, nil
+	default:
+		return false, decodeErrorf("invalid bool value: %q", bs[0])
```
abhinav, line 112: `%q` will treat the byte as a character. we should use `%d` or `%v` here.

protocol/binary/stream_reader.go
```diff
@@ -0,0 +1,445 @@
...
+	return nil
+}
+
+// ReadFieldBegin reads off a Thrift encoded field-header returning that and a
+// 'bool' representing whether or not a field-value follows.
+// A 'false' without any error means that it has reached the stop-field.  There
+// is no guarantee that the field-header is valid in this case.
+func (sr *StreamReader) ReadFieldBegin() (stream.FieldHeader, bool, error) {
+	fh := stream.FieldHeader{}
+
```
abhinav, line 207: optional: given how we're using `fh`, named returns might be more shorter, and also document the bool:
```suggestion
func (sr *StreamReader) ReadFieldBegin() (fh stream.FieldHeader, ok bool, err error) {
```

protocol/binary/stream_reader.go
```diff
@@ -0,0 +1,445 @@
...
+// ReadFieldEnd reads the "end" of a Thrift encoded field  Since there is no
+// encoding for the end of a field, this is a noop.
+func (sr *StreamReader) ReadFieldEnd() error {
+	return nil
+}
+
+// ReadListBegin reads off the list header of a Thrift encoded list.
+func (sr *StreamReader) ReadListBegin() (stream.ListHeader, error) {
+	lh := stream.ListHeader{}
+
```
abhinav, line 237:
```suggestion
func (sr *StreamReader) ReadListBegin() (lh stream.ListHeader, err error) {
```

protocol/binary/stream_reader.go
```diff
@@ -0,0 +1,445 @@
...
+// ReadSetEnd reads the "end" of a Thrift encoded list.  Since there is no
+// encoding for the end of a set, this is a noop.
+func (sr *StreamReader) ReadSetEnd() error {
+	return nil
+}
+
+func (sr *StreamReader) readTypeSizeHeader() (wire.Type, int, error) {
+	elemType, err := sr.ReadInt8()
+	if err != nil {
+		return wire.Type(0), 0, err
```
abhinav, line 277: . Constant literals are untyped, so just 0 should work as a wire.Type here.

```suggestion
		return 0, 0, err
```

abhinav (approved):
LGTM minus nits

protocol/binary/envelope.go
```diff
@@ -117,7 +119,11 @@ func (bw *Reader) ReadEnveloped() (wire.Envelope, error) {
 }
 
 func (bw *Reader) readStrictNameType(initial int32, off int64) (wire.Envelope, int64, error) {
-	var e wire.Envelope
+	var (
+		e       wire.Envelope
+		rawName wire.Value
+		err     error
+	)
```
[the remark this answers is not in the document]
witriew: Good catch!  Thanks.

protocol/binary/stream_reader.go
witriew: TIL about `ReadFull`, thanks - integrated that in.

PR #495, opened 2021-06-11
Generate "Decode" for "enums" that will directly decode

witriew:
Add templating for generating code for "enums" that will utilize the new
`stream.Reader`, performing the same duties as `binary.Decode` and `FromWire`,
but without going through an intermediary format.

### DIFF
--- a/gen/enum.go
+++ b/gen/enum.go
@@ -70,6 +70,7 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 
 		<$stream := import "go.uber.org/thriftrw/protocol/stream">
 		<$wire := import "go.uber.org/thriftrw/wire">
+		<$stream := import "go.uber.org/thriftrw/protocol/stream">
 
 		<$enumName := goName .Spec>
 		<formatDoc .Spec.Doc>type <$enumName> int32
@@ -199,6 +200,26 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 		//   return <$v>, nil
 		func (<$v> *<$enumName>) FromWire(<$w> <$wire>.Value) error {
 			*<$v> = (<$enumName>)(<$w>.GetI32());
+			return nil
+		}
+
+		<$sr := newVar "sr">
+		// Decode reads off the encoded <$enumName> directly off of the wire.
+		//
+		//   sReader := BinaryStreamer.Reader(reader)
+		//
+		//   var <$v> <$enumName>
+		//   if err := <$v>.Decode(sReader); err != nil {
+		//     return <$enumName>(0), err
+		//   }
+		//   return <$v>, nil
+		func (<$v> *<$enumName>) Decode(<$sr> <$stream>.Reader) error {
+			<- $i := newVar "i" ->
+			<$i>, err := <$sr>.ReadInt32()
+			if err != nil {
+				return err
+			}
+			*<$v> = (<$enumName>)(<$i>)
 			return nil
 		}
 
--- a/gen/internal/tests/enum_conflict/enum_conflict.go
+++ b/gen/internal/tests/enum_conflict/enum_conflict.go
@@ -133,6 +133,24 @@ func (v *RecordType) FromWire(w wire.Value) error {
 	return nil
 }
 
+// Decode reads off the encoded RecordType directly off of the wire.
+//
+//   sReader := BinaryStreamer.Reader(reader)
+//
+//   var v RecordType
+//   if err := v.Decode(sReader); err != nil {
+//     return RecordType(0), err
+//   }
+//   return v, nil
+func (v *RecordType) Decode(sr stream.Reader) error {
+	i, err := sr.ReadInt32()
+	if err != nil {
+		return err
+	}
+	*v = (RecordType)(i)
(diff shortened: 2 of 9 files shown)


gen/enum.go
```diff
@@ -190,6 +191,32 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+		//     return <$enumName>(0), err
+		//   }
+		//   return <$v>, nil
+		func (<$v> *<$enumName>) Decode(<$sr> <$stream>.Reader) error {
+			<- $rVal := newVar "rVal" ->
+			<$rVal>, err := <$sr>.ReadInt32()
+			if err != nil {
+				return err
+			}
+
```
abhinav, line 215: small nit: delete this empty line please.
```suggestion
```

gen/roundtrip_test.go
```diff
@@ -81,3 +82,47 @@ func assertBinaryRoundTrip(t *testing.T, w wire.Value, message string) (wire.Val
 
 	return newV, true
 }
+
+// assertStreamingRoundTrip checks that the streaming Encode/Decode returns the same
+// value.
+//
+// Note: Leverages 'ToWire' and regular binary.Encode until a 'stream.Writer'
+// implementation is plumbed through.
+func assertStreamingRoundTrip(t *testing.T, x thriftType, v wire.Value, msg string, args ...interface{}) bool {
```
abhinav, line 91: So, this is probably fine for now, but possibly once we have the writer, an alternative is that this is the new assertRoundTrip, and does a cross-product of streaming/non-streaming encode/decode.

Like,

```
streams := []struct{ encode, decode bool }{
  {false, false},
  {false, true},
  {true, false},
  {true, true}
}

for _, stream := range streams {
  _, canStream := x.(streamingThriftType)
  
  // Skip cases where we can't stream and streaming is requested.
  if !canStream && (stream.encode || stream.decode) {
    continue
  }
  
  name := ...
  t.Run(name, func(t *testing.T) {
    var buff bytes.Buffer
    if stream.encode {
        x.Encode(binary.Writer(&buff))
    } else {
        binary.Encode(x.ToWire(), &buff)
    }
    
    if stream.decode {
        // ...
    } else {
        // ...
    }
  })
}
```

witriew: Yep, I was imagining this as well.

witriew: Added a comment to reflect that we want to do that eventually.

gen/enum.go
```diff
@@ -190,6 +191,32 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 			return nil
 		}
 
+		<$sr := newVar "sr">
+		// Decode reads off the encoded <$enumName> directly off of the wire.
+		//
+		//   sp, err := binaryProtocol.(stream.Protocol)
+		//   if err != nil {
+		//     return <$enumName>(0), err
+		//   }
```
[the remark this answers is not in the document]
witriew: Cleaned this up a bit, lemme know how it looks.

gen/internal/tests/enum_conflict/enum_conflict.go
```diff
@@ -122,6 +123,30 @@ func (v *RecordType) FromWire(w wire.Value) error {
 	return nil
 }
 
+// Decode reads off the encoded RecordType directly off of the wire.
+//
+//   sp, err := binaryProtocol.(stream.Protocol)
+//   if err != nil {
+//     return RecordType(0), err
+//   }
```
abhinav, line 131: outdated generated code

abhinav (commented):
Generated code is out of date, and needs to be rebased, but lgtm otherwise.

gen/enum.go
```diff
@@ -190,6 +191,26 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
+		//
+		//   sReader := BinaryStreamer.Reader(reader)
+		//
+		//   var <$v> <$enumName>
+		//   if err := <$v>.Decode(sReader); err != nil {
+		//     return <$enumName>(0), err
+		//   }
+		//   return <$v>, nil
+		func (<$v> *<$enumName>) Decode(<$sr> <$stream>.Reader) error {
+			<- $rVal := newVar "rVal" ->
```
abhinav, line 205: optional nitty nit: I would have just named this "x", "i", or "val". There's not much ambiguity here; it's a really small function to require the "rVal" disambiguation.

abhinav (approved):
LGTM

gen/util_for_test.go
```diff
@@ -37,3 +38,15 @@ type thriftType interface {
 	ToWire() (wire.Value, error)
 	FromWire(wire.Value) error
 }
+
+// streamingThriftType is implemented by all generated types that know how to
+// write and read themselves to the Thrift Protocol, skipping over the
+// intermediary wire.Type
+//
+// Only requires Decode until Encode is plubmed through to a stream.Writer.
```
mh-park, line 46: spelling nit:
```suggestion
// Only requires Decode until Encode is plumbed through to a stream.Writer.
```

mh-park (approved):
LGTM

Issue #349, opened 2018-03-30
ast.DefinitionInfo should include the location of the source code


DAddYE:
Hi!

Now each `ast.Definition` has `func Info() DefinitionInfo()` which returns a struct with `Name` and `Line` of the `Definition`, however, would be amazing if instead of the line we have a `Location[int, int]` (i.e [token.Position](https://golang.org/pkg/go/token/#Position)) this would allow me to select back the text range in the source file and do extra parsing.

abhinav:
Hey, it's not quite clear what you're asking. Did you want column numbers as
well? Or do you want the start and end positions?

The first would be possible by doing some extra book keeping in the lexer and
passing it on to the constructed nodes similar to the line number.

The latter is possible but significantly more difficult because it requires a
lot more changes to the parser and lexer.

If you expand on the use case a bit, we can maybe figure out a work around.

jparise:
@abhinav would you prefer to see this added as distinct `Line` and `Column` fields:

```go
type DefinitionInfo struct {
	Name   string
	Line   int
    Column int
}
```

... or by moving to an embedded `Position`?

```go
type DefinitionInfo struct {
    Position
	Name string
}
```

abhinav:
Hey, sorry for the delay. 
Switching to embedding might be considered a breaking change. 

I would lean with the first option, but an alternative is a new Pos field, with Line deprecated in favor of it. We'd still fill Line for backwards compatibility, but Pos is what should be used going forward. Something like this?

```go
type DefintiionInfo struct {
  Name string

  // Deprecated: Use Pos.
  Line int

  // Position of the definition in the source file.
  Pos Position
}
```

jparise:
I think it will probably end up being more trouble maintaining both `Line` and `Pos` together. The parser tests will become pretty verbose.

We can always produce a `Position` from `Line` and `Column` fields (which I'll offer through a `nodeWithPosition` interface and an `ast.Pos(ast.Node)` function).

At a future point when we can break some backwards compatibility, we could drop `Line` and `Column` in favor of a single `Pos`, or decide to embed `Position`.

Issue #493, opened 2021-06-10
Line numbers for primitive constants


jparise:
These primitive constant AST types don't carry line number information (i.e. they don't implement the `nodeWithLine` interface):

```go
// ConstantBoolean is a boolean value specified in the Thrift file.
//
//   true
//   false
type ConstantBoolean bool

// ConstantInteger is an integer value specified in the Thrift file.
//
//   42
type ConstantInteger int64

// ConstantString is a string literal specified in the Thrift file.
//
//   "hello world"
type ConstantString string

// ConstantDouble is a floating point value specified in the Thrift file.
//
//   1.234
type ConstantDouble float64
```

This is/was intentional, but I wanted to ask if there might be a path to providing `lineNumber()` implementations for these types.

abhinav:
Hey there! Unfortunately, this was less of an intentional decision and
more of an oversight.

By the time we added support for line number tracking, we had already
released 1.0 with those type definitions. We decided that losing losing
line number tracking on constant values was okay versus making a
breaking change by converting them to structs.

Today, we're still opposed to a breaking change, but there might be a
path forward. Unfortunately, not all `ast.Node` types are hashable, so
we can't just generate, say, a `map[ast.Node]Position`.

One possibility is to track line numbers for those and other hashable
types on a map on `ast.Program` on an unexported field. Something like:

```go
type Program struct {
    // ...
    
    otherLineNums map[Node]int
}
```

If a type is hashable, then we stick its location in there while
parsing. Types that are not hashable (and so, can't go there) are all
struct types, so they have their own Line fields.

We then expose a method on Program to access line information for any
node that uses the node's lineNumber method, if any, and the map if
not. Something like:

```go
func (p *Program) LineFor(n ast.Node) int {
    if n, ok := n.(nodeWithLine); ok {
        return n.lineNumber()
    }
    return p.otherLineNums[n] // zero if unknown
}
```

---

One disadvantage of this is that the added internal state may break
large, complex deep-equals between parsed `ast.Program`s and hand-
written ones.

```go
p, err := idl.Parse(...)
...
assert.Equal(t, &Program{...}, p)
// This will begin failing because Program.otherLineNums isn't set
// in the hand-written one.
```

So an alteration to the idea above is that we don't add this state to
Program, but to a new `idl.Info` type:

```go
package idl

type Info struct {
    otherLineNums map[ast.Node]int
}

func (*Info) LineFor(ast.Node) int
```

[...]

PR #497, opened 2021-06-16
idl: add a Position struct to wrap reported lines

jparise:
Line-based data types are too limiting should we add support for other
positional information, such as offsets/columns (#349).

### DIFF
--- a/idl/error.go
+++ b/idl/error.go
@@ -29,30 +29,33 @@ import (
 
 // ParseError is an error type listing parse errors and the positions
 // that caused them.
-type ParseError struct{ Lines []LineError }
+type ParseError struct{ Errors []Error }
 
-// LineError holds an error and the line number that caused it.
-type LineError struct {
-	Line int
-	Err  error
+// Error holds an error and the position that caused it.
+type Error struct {
+	Pos Position
+	Err error
 }
 
-func newParseError(errors []internal.LineError) error {
+func newParseError(errors []internal.ParseError) error {
 	if len(errors) == 0 {
 		return nil
 	}
-	lines := make([]LineError, len(errors))
+	errs := make([]Error, len(errors))
 	for i, err := range errors {
-		lines[i] = LineError{Line: err.Line, Err: err.Err}
+		errs[i] = Error{
+			Pos: Position{Line: err.Pos.Line},
+			Err: err.Err,
+		}
 	}
-	return &ParseError{Lines: lines}
+	return &ParseError{Errors: errs}
 }
 
 func (pe *ParseError) Error() string {
 	var buffer bytes.Buffer
 	buffer.WriteString("parse error\n")
-	for _, line := range pe.Lines {
-		buffer.WriteString(fmt.Sprintf("  line %d: %s\n", line.Line, line.Err))
+	for _, pe := range pe.Errors {
+		buffer.WriteString(fmt.Sprintf("  line %d: %s\n", pe.Pos.Line, pe.Err))
 	}
 	return buffer.String()
 }
--- a/idl/internal/error.go
+++ b/idl/internal/error.go
@@ -20,8 +20,8 @@
 
 package internal
 
-// LineError holds a parse error and the line number that caused it.
-type LineError struct {
-	Line int
-	Err  error
+// ParseError holds a parse error and the position that caused it.
+type ParseError struct {
+	Pos Position
+	Err error
 }
--- a/idl/internal/lex.go
+++ b/idl/internal/lex.go
@@ -48,7 +48,7 @@ type lexer struct {
 	lastDocstring       string
 	linesSinceDocstring int
 
-	errors      []LineError
+	errors      []ParseError
 	parseFailed bool
 
 	// Ragel:
@@ -16605,7 +16605,7 @@ func (lex *lexer) Error(e string) {
 
 func (lex *lexer) AppendError(err error) {
 	lex.parseFailed = true
-	lex.errors = append(lex.errors, LineError{Line: lex.line, Err: err})
+	lex.errors = append(lex.errors, ParseError{Pos: Position{Line: lex.line}, Err: err})
 }
 
 func (lex *lexer) LastDocstring() string {
--- a/idl/internal/lex.rl
+++ b/idl/internal/lex.rl
@@ -28,7 +28,7 @@ type lexer struct {
     lastDocstring string
     linesSinceDocstring int
 
-    errors []LineError
+    errors []ParseError
     parseFailed bool
 
     // Ragel:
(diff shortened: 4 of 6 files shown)


jparise:
This relates to both #349 and #493. I'd like to get (something like) this in place before the line-based interface exported by #492 makes it into a published release.

jparise:
A few questions:

- Is `Position` or `Location` the right word? I think either is fine, but now is the time to choose.
- Should `Position` be embedded in the `PositionError` struct or be a distinct `Pos` field?

abhinav:
Thanks for the PR!

> Is Position or Location the right word? I think either is fine, but now is the time to choose.

Agreed. Either is fine by me. I think location reads nicer than position in this context, but there's [prior art on position representing locations](https://golang.org/pkg/go/token/#Position). I'll defer this choice to you.

> Should Position be embedded in the PositionError structs or be a distinct Pos field?

I think I'd prefer a Pos field (or Loc if we name the type Location).

jparise:
> > Is Position or Location the right word? I think either is fine, but now is the time to choose.
> 
> Agreed. Either is fine by me. I think location reads nicer than position in this context, but there's [prior art on position representing locations](https://golang.org/pkg/go/token/#Position). I'll defer this choice to you.

I looked at some prior art. The [Language Server Protocol (LSP) specification](https://microsoft.github.io/language-server-protocol/specifications/specification-current) uses both words with specific meanings:

- [`Position`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#position) is essential a (line, offset) pair, like a coordinate.
- [`Location`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#location) is a higher-level representation that can include a range of positions (start, end) as well as a URI (such as a filename).

Given the buffer-based design of the IDL module, we can't provide a URI (filename), although we could provide a way to inject one such as via the `idl.Config` + `ast.Info` concepts proposed in #493. We could also provide ranges as an advanced implementation of #349.

So I think `Position` is accurate for what we're probably going to be able to provide in the near term, but `Location` would buy is more room to maneuver in the future. The complication is that by introducing `Location` now, we'd be doing so knowing that we'd want to extend it in the future, because it might be too limiting to ship:

```go
type Location struct {
    Line int
    // ... and then add:
    Offset int
    // ... and then realize we want (start, end) ranges
}
```

All that being said, I think we could commit to doing the following before we release a new version (so we have the flexibility to iterate a little bit within the release window):

1. Implement #349, and as part of that implementation, decide if we'll support ranges or just (line, offset).
2. Based on (1), we'll know the general shape of the `Location` struct.
3. Choose whether to also add a `Filename` field to the `Location` struct as well. (I don't think this is especially useful unless we offer recursive document parsing, though.)

My current instinct is that we'll probably _not_ implement ranges in (1) or filenames in (3), and then we'll end up with:

```go
type Location struct {
    Line int
    Offset int
}
```

... which is effectively just a `Position`.

Not to just turn this back around at you, but do you land in the same place?

abhinav:
Those are valid points. Yeah, I suspect we won't end up doing ranges,
in which case Position is fine as-is.

(Although, naming nit: `Offset`, if we add it, would be byte offset
within a file. The column number in a line would be on a `Column`
field.)

---

If we really want to track ranges, there's the possibility of an
alternative, slightly more complicated design inspired by go/token and
go/ast. The gist of it is:

- track position ranges as byte offsets in the file `(start, end)`
- expose them via `Info.Start(Node)` and `Info.End(Node)` methods
  returning offsets, similar to how we were going to do line numbers
- separately in Info, track a list of byte offsets in the file at which
  newlines occur (a sorted `[]Offset`)
- `Info.Position(Offset) Position` returns a `Position{Line, Column}`,
  computed dynamically from the knowledge of newline offsets
  (approximately: binary search in `[]Offset` to get line number,
  subtract offset for that line to get column number).
  
So the rough surface for this is:

```go
type Offset int
// To make the zero offset mean "unknown" (to account for existing code
// and struct literals), we can make offset 1-indexed in the file.

type Position struct {
    Line int // 1-indexed line number
    // In the future,
    //   File string  // if known
    //   Column int
}

type Info struct {
    newlines []Offset
    otherStarts map[ast.Node]Offset
}

func (*Info) Start(Node) Offset

// In the future,
//   func (*Info) End(Node) Offset

func (*Info) Position(Offset) Position

// Optional convenience methods:
//   func (*Info) StartPosition(Node) Position{ i.Position(i.Start(n)) }
//   func (*Info) EndPosition(Node) Position{ i.Position(i.End(n)) }

type Config struct {
    Info *Info

    // In the future,
    //  File string // name of the file, if known
}

func (*Config) Parse([]byte) (*ast.Program, error)
```

Used like so:

```go
var info Info
cfg := idl.Config{Info: &info, File: "foo.thrift"}

prog, err := cfg.Parse(..)
if err != nil {
  // ...
}

n := someNode(prog)

startPos := info.Position(info.Start(n))
fmt.Println(startPos.Line)  // 42

// In the future:
//   fmt.Println(startPos)  // foo.thrift:42:5
//
// Also in the future:
//   fmt.Println(info.EndPosition(n))  // foo.thrift:45:1
```

This is similar to how `go/token.FileSet` and `go/token.Pos` work to
track positions of Go AST nodes.

Advantages: this can support column numbers, file names, and ranges,
and costs only one integer per node (two if/when we add end positions).

Disadvantages: it's a lot more API surface area, it's more complicated
to implement and use, and it leaves a redundant `Line` field on all
existing AST nodes that have it.

---

I'm throwing the second design idea mostly as a maybe. Right now, I
don't have the time to implement it, and we don't have the need for
that functionality. I don't want to volunteer your time for anything
you don't need from this library.

So, if you really want column tracking and positional ranges, we'd
welcome an implementation of the above, or of an alternative/better
proposal.

Equally valid: if you don't have need of this functionality in a
timeline that matters, I'm also okay with a simple `Position` struct
that leaves room for a Column field in the future.

jparise:
I think I'm good sticking with what's in this change for now. I'm motivated to get these two things done next:

1. Line numbers from #493 (including the proposed `Config` concept)
2. Column numbers from #349

... but I don't have a need for ranges at this point.

idl/error.go
```diff
@@ -29,30 +29,33 @@ import (
...
-type ParseError struct{ Lines []LineError }
+type ParseError struct{ Errors []Error }
 
-// LineError holds an error and the line number that caused it.
-type LineError struct {
-	Line int
-	Err  error
+// Error holds an error and the position that caused it.
+type Error struct {
+	Pos Position
```
abhinav, line 36: Note to other reviewers: Not a breaking change. We haven't released this yet.

abhinav (approved):
Sounds good, thanks!

PR #496, opened 2021-06-16
Provide "decode" code generation for the streaming variants for all other types

witriew:
#495 provided the code generation for streaming reads of enums off of the wire.
Provide the streaming read generators for all of the other types.

Doing so meant providing a `StreamGenerator` that plumbs into the existing
generators.  This `StreamGenerator` only provides the "decoding" mechanism,
leveraging #491 for reading the raw Thrift encoding off of the wire.

In addition to providing a `StreamGenerator`, the list, map, set, enum, typdef,
and struct generators all added a `Decoder` method that will appropriately
recurse and iterate to generate the proper, mirrored readers for the raw wire
representation.

The `Decode` and `DecodePtr` methods for `StreamGenerator` then hook into the
templated generator itself, providing the templated `decode` and `decodePtr`
calls that will be used where necessary.

The existing tests were leveraged to make sure that the streaming reads were
compatible with the current binary writes and that no data was lost.  In doing
so, a 'genericized' function that will perform the cross-products of
encoding/writing and decoding/reading.

A benchmark was also added to evaluate the new streaming reads.  The following
represents benchmarks of the original code vs the dedup'd code (the binary
decode/encode leveraging the streaming implementations):
```
name                                        old time/op    new time/op    delta
RoundTrip/PrimitiveOptionalStruct/Encode-8    1.76µs ± 2%    1.86µs ± 5%    +5.93%  (p=0.008 n=5+5)
RoundTrip/PrimitiveOptionalStruct/Decode-8    2.55µs ± 3%    2.82µs ± 5%   +10.68%  (p=0.008 n=5+5)
RoundTrip/Graph/Encode-8                      3.15µs ± 2%    3.16µs ± 1%      ~     (p=1.000 n=5+5)
RoundTrip/Graph/Decode-8                      5.02µs ± 1%    8.95µs ±11%   +78.05%  (p=0.008 n=5+5)
RoundTrip/ContainersOfContainers/Encode-8     19.1µs ± 1%    19.8µs ± 3%    +3.76%  (p=0.008 n=5+5)
RoundTrip/ContainersOfContainers/Decode-8     45.2µs ± 3%   110.8µs ± 1%  +144.98%  (p=0.008 n=5+5)

name                                        old alloc/op   new alloc/op   delta
RoundTrip/PrimitiveOptionalStruct/Encode-8      704B ± 0%      704B ± 0%      ~     (all equal)
[...]

PR #498, opened 2021-06-16
Add code generation for all wire types for stream encoding

dianale31:
Creates the code generation for all the primitive types, enums, typedefs, structs, and container types. We use the same testing utils as the reader implementation in #496 (hopefully that will reduce rebase conflicts).

### DIFF
--- a/gen/benchmark_test.go
+++ b/gen/benchmark_test.go
@@ -9,6 +9,8 @@ import (
 	tc "go.uber.org/thriftrw/gen/internal/tests/containers"
 	ts "go.uber.org/thriftrw/gen/internal/tests/structs"
 	"go.uber.org/thriftrw/protocol"
+	"go.uber.org/thriftrw/protocol/binary"
+	"go.uber.org/thriftrw/protocol/stream"
 	"go.uber.org/thriftrw/ptr"
 	"go.uber.org/thriftrw/wire"
 )
@@ -16,6 +18,12 @@ import (
 type thriftType interface {
 	ToWire() (wire.Value, error)
 	FromWire(wire.Value) error
+}
+
+type streamingThriftType interface {
+	thriftType
+
+	Encode(stream.Writer) error
 }
 
 func BenchmarkRoundTrip(b *testing.B) {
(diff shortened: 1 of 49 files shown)


gen/benchmark_test.go
```diff
@@ -9,13 +9,16 @@ import (
...
+	"go.uber.org/thriftrw/protocol/binary"
+	"go.uber.org/thriftrw/protocol/stream"
 	"go.uber.org/thriftrw/ptr"
 	"go.uber.org/thriftrw/wire"
 )
 
 type thriftType interface {
 	ToWire() (wire.Value, error)
 	FromWire(wire.Value) error
+	Encode(stream.Writer) error
```
witriew, line 21: Should be in a separate interface declaration - like you currently have in `gen/utils_for_test.go`

gen/field.go
```diff
@@ -467,6 +471,108 @@ func (f fieldGroupGenerator) FromWire(g Generator) error {
...
+	return g.DeclareFromTemplate(
+		`
+		<$stream := import "go.uber.org/thriftrw/protocol/stream">
+		<$v := newVar "v">
+		<$sw := newVar "sw">
+		// Decode deserializes a <.Name> struct directly from its Thrift-level
+		// representation, without going through an intemediary type.
+		//
+		// An error is returned if a <.Name> struct could not be generated from the wire
+		// representation.
```
witriew, line 484: Comment :D

gen/struct.go
```diff
@@ -51,6 +51,27 @@ func (s *structGenerator) Reader(g Generator, spec *compile.StructSpec) (string,
...
 
+func (s *structGenerator) Encoder(g Generator, spec *compile.StructSpec) (string, error) {
+	name := encoderFuncName(g, spec)
+	err := g.EnsureDeclared(
+		`
+		<$stream := import "go.uber.org/thriftrw/protocol/stream">
+		<$val := newVar "val">
+		<$sw := newVar "sw">
+		<$t := typeReference .Spec>
+		func <.Name>(<$val> <$t>, <$sw> <$stream>.Writer) error {
```
witriew, line 62: I think you should be able to inline the `typeReference .Spec`
```suggestion
		func <.Name>(<$val> <typeReference .Spec>, <$sw> <$stream>.Writer) error {
```

gen/field.go
```diff
@@ -95,6 +95,10 @@ func (f fieldGroupGenerator) Generate(g Generator) error {
 		return err
 	}
 
+	if err := f.Encode(g); err != nil {
+		return err
+	}
```
witriew, line 100: Don't forget to add "Encode" to the list of `reservedIdentifiers`.

protocol/binary/stream_writer.go
```diff
@@ -65,6 +65,14 @@ func ReturnStreamWriter(sw *StreamWriter) {
 	streamWriterPool.Put(sw)
 }
 
+func NewStreamWriter(w io.Writer) *StreamWriter {
+	return &StreamWriter{
+		Impl:   iface.Impl{},
+		writer: w,
+		buffer: [8]byte{},
+	}
+}
```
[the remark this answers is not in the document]
dianale31: Decided to go with the BorrowStreamWriter and ReturnStreamWriter approach which was added in #490

PR #499, opened 2021-06-16
Support "<" in the templating language

witriew:
The templating language for code generation uses `<` and `>` as its
delimiters.  This results in syntax errors when trying to use the beginning
delimiter as an actual character , not as invoking a template.  Add an explicit
command (`<lessthan>`) to inject `<` as needed.


abhinav (approved):
In hindsight, "<", ">" may not have been the best choices for the template delimiters.

PR #502, opened 2021-06-23
gen: conditionally return err on enum marshal/unmarshal of unknown values 

aloksinghaluber:
Refs EDGE-577


abhinav:
Please include a lot more information in the PR description, including references to relevant tickets.

abhinav:
I'm sorry, an issue reference isn't enough; check out the rest of the history for this repo. By "a lot more information", I mean that the PR or the commit must at least answer:

- What is the change trying to do
- Why is this change needed
- Are there any other things we need to be aware of

See [this section of the How to Write a Git Commit Message post](https://chris.beams.io/posts/git-commit/#why-not-how) for what information we want in the PR description. Separately, [A Branch in Time](https://tekin.co.uk/2019/02/a-talk-about-revision-histories) talks about the importance of a good commit history and why references to JIRA tickets aren't enough.

aloksinghaluber:
Generate code to return error if we are trying to marshal an unknown enum value in marshal Json function. Current behavior is to send an integer value. 

We have added a flag to feature gate these changes and keep the default behaviour unchanged. 

These changes are done since some mobile clients crash when we send an integer when they are expecting a string value for the Enum. This led to an L5H incident: https://incidents.uberinternal.com/incident/5e7ad5cc-2636-4584-8705-6286dffa6b30

main.go
```diff
@@ -60,6 +60,7 @@ type genOptions struct {
 	NoEmbedIDL        bool   `long:"no-embed-idl" description:"Do not embed IDLs into the generated code."`
 	NoZap             bool   `long:"no-zap" description:"Do not generate code for Zap logging."`
 	OutputFile        string `long:"output-file" value-name:"FILENAME" description:"Generates a single .go file as an output. Specifying an OutputFile prevents code generation for included Thrift Files."`
+	ErrOnUnknownEnum  bool    `long:"error-on-unknown-enum" value-name:"ERRUNKNOWNENUM" description:"Send error on unknown enum marshal"`
```
rpatali, line 63: **nit**:alignment

gen/internal/tests/err_on_unknown_enum_value/err_on_unknown_enum_value.go
```diff
@@ -0,0 +1,211 @@
...
+		if x > math.MaxInt32 {
+			return fmt.Errorf("enum overflow from JSON %q for %q", text, "EnumDefault")
+		}
+		if x < math.MinInt32 {
+			return fmt.Errorf("enum underflow from JSON %q for %q", text, "EnumDefault")
+		}
+		*v = (EnumDefault)(x)
+		return nil
+	case string:
+		return v.UnmarshalText([]byte(w))
```
rpatali, line 196: I think it makes sense to also throw an error during unmarshal of unknown values. The flag is not specific to marshaling (and probably shouldn't be).

gen/internal/tests/err_on_unknown_enum_value/err_on_unknown_enum_value.go
```diff
@@ -0,0 +1,219 @@
...
+	case 0:
+		return []byte("Foo"), nil
+	case 1:
+		return []byte("Bar"), nil
+	case 2:
+		return []byte("Baz"), nil
+	case 3:
+		return []byte("Bat"), nil
+	}
+	return []byte(strconv.FormatInt(int64(v), 10)), nil
```
rpatali, line 78: what about this? Also, how is (Un) MarshalText actually used?

gen/enum.go
```diff
@@ -110,19 +114,27 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
 			}
 		}
 
 		// MarshalText encodes <$enumName> to text.
 		//
-		// If the enum value is recognized, its name is returned. Otherwise,
-		// its integer value is returned.
+		// If the enum value is recognized, its name is returned.
+		<if checkErrOnUnknownEnum ->
+		//Otherwise, error is thrown
```
rpatali, line 134: *nit*: space after `//` for consistency with the rest of the generated comments.

rpatali (approved):
LGTM

gen/enum.go
```diff
@@ -66,7 +66,11 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 		<$fmt := import "fmt">
 		<$json := import "encoding/json">
 		<$math := import "math">
-		<$strconv := import "strconv">
+		<if checkErrOnUnknownEnum ->
+			<$errors := import "errors">
+		<else ->
+			<$strconv := import "strconv">
+		<end ->
```
abhinav, line 73: We can just drop this and do the imports at the site where the package is needed.

```
  <$errors := import "errors">
  <$errors>.New(..)
```

gen/enum.go
```diff
@@ -110,19 +114,27 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 					return nil
 			<end ->
 				default:
-					<$val>, err := <$strconv>.ParseInt(<$s>, 10, 32)
+					<if checkErrOnUnknownEnum ->
+						return <$fmt>.Errorf("unknown enum value %q for %q", <$s>, "<$enumName>")
+					<else ->
+					<$val>, err := strconv.ParseInt(<$s>, 10, 32)
```
abhinav, line 120: If we're importing strconv with `<$strconv := import "..">`, we should use `<$strconv>` here to reference the package. The variable accounts for conflicting imports.
```suggestion
					<$val>, err := <$strconv>.ParseInt(<$s>, 10, 32)
```

gen/enum.go
abhinav: This comment is still valid.

gen/enum.go
```diff
@@ -134,7 +146,11 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 				<end ->
 				}
 			<end ->
-			return []byte(<$strconv>.FormatInt(int64(<$v>), 10)), nil
+			<if checkErrOnUnknownEnum ->
+				return nil, errors.New(fmt.Sprintf("unknown value for enum %s", <$v>))
+			<else ->	
+				return []byte(strconv.FormatInt(int64(<$v>), 10)), nil
```
abhinav, line 152: As before, we either need to use `<$errors>` and `<$strconv>` here, or we can do:

    <import "errors">.New(...)

That will handle ensuring that the package is imported and not import it twice.

gen/enum.go
```diff
@@ -134,7 +146,11 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 				<end ->
 				}
 			<end ->
-			return []byte(<$strconv>.FormatInt(int64(<$v>), 10)), nil
+			<if checkErrOnUnknownEnum ->
+				return nil, errors.New(fmt.Sprintf("unknown value for enum %s", <$v>))
```
abhinav, line 150: `errors.New(fmt.Sprintf(..))` is redundant. Please use `fmt.Errorf(..)` instead.

staticcheck will usually catch this except it's part of the template. https://staticcheck.io/docs/checks#S1028

main.go
```diff
@@ -60,6 +60,7 @@ type genOptions struct {
 	NoEmbedIDL        bool   `long:"no-embed-idl" description:"Do not embed IDLs into the generated code."`
 	NoZap             bool   `long:"no-zap" description:"Do not generate code for Zap logging."`
 	OutputFile        string `long:"output-file" value-name:"FILENAME" description:"Generates a single .go file as an output. Specifying an OutputFile prevents code generation for included Thrift Files."`
+	ErrOnUnknownEnum  bool   `long:"error-on-unknown-enum" value-name:"ERRUNKNOWNENUM" description:"Send error on unknown enum marshal"`
```
abhinav, line 63: I'm reluctant to advertise this flag with this description. Marshaling is part of Thrift's problem statement, and this flag has no effect on the *Thrift* marshaling. It only affects the unsupported JSON/text marshaling of enums.

How about:

- make the flag hidden (`hidden:"true"` like the GeneratePluginAPI flag above) so that only people like yourself who know about it can use it
- include "text-marshal" somewhere in the name (`--enum-text-marshal-strict`?)

Also, the `value-name:"..."` option isn't needed for a boolean flag—it specifies the name for the argument of the flag, if any, and booleans don't usually take arguments.

abhinav (changes requested):

gen/generate.go
```diff
@@ -85,6 +85,9 @@ type Options struct {
 
 	// Name of the file to be generated by ThriftRW.
 	OutputFile string
+
+	//Generate Error on Unknown Enum Marshal
```
mh-park, line 89: nit: space after // and the comment should be like a sentence (no need to capitalize every word)

gen/golden_test.go
```diff
@@ -42,6 +42,10 @@ var noZapFiles = map[string]struct{}{
 	"nozap": {},
 }
 
+var errOnUnknownEnumFiles = map[string]struct{}{
+	"err_on_unknown_enum_value" : {},
```
mh-park, line 46: I don't see a `internal/tests/thrift/err_on_unknown_enum_value` directory.

gen/enum.go
```diff
@@ -110,19 +109,27 @@ func enum(g Generator, spec *compile.EnumSpec) error {
...
 			}
 		}
 
 		// MarshalText encodes <$enumName> to text.
 		//
-		// If the enum value is recognized, its name is returned. Otherwise,
-		// its integer value is returned.
+		// If the enum value is recognized, its name is returned.
+		<if checkEnumTextMarshalStrict ->
+		// Otherwise, error is thrown
```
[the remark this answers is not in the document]
abhinav:
```suggestion
		// Otherwise, an error is returned.
```

gen/enum.go
```diff
@@ -134,7 +141,11 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 				<end ->
 				}
 			<end ->
-			return []byte(<$strconv>.FormatInt(int64(<$v>), 10)), nil
+			<if checkEnumTextMarshalStrict ->
+				return nil, <$fmt>.Errorf("unknown value for enum %s", <$v>)
```
abhinav, line 145: we should include the value and the enum name like above:

```suggestion
				return nil, <$fmt>.Errorf("unknown enum value %q for %q", <$v>, "<$enumName>")
```

gen/enum.go
```diff
@@ -213,8 +224,12 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 
 		// MarshalJSON serializes <$enumName> into JSON.
 		//
-		// If the enum value is recognized, its name is returned. Otherwise,
-		// its integer value is returned.
+		// If the enum value is recognized, its name is returned.
+		<if checkEnumTextMarshalStrict ->
+		// Otherwise, error is thrown
```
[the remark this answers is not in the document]
abhinav:
```suggestion
		// Otherwise, an error is returned.
```

gen/enum.go
```diff
@@ -226,7 +241,11 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 				<end ->
 				}
 			<end ->
-			return ([]byte)(<$strconv>.FormatInt(int64(<$v>), 10)), nil
+			<if checkEnumTextMarshalStrict ->
+				return nil, <$fmt>.Errorf("unknown value for enum %s", <$v>)
```
abhinav, line 245:
```suggestion
				return nil, <$fmt>.Errorf("unknown enum value %q for %q", <$v>, "<$enumName>")
```

gen/generate.go
abhinav:
```suggestion
	// Generates an error on MarshalText and MarshalJSON if the enum value is
	// unrecognized.
```

gen/golden_test.go
abhinav: Yeah should this be `enum-text-marshal-strict`?

gen/enum.go
```diff
@@ -110,19 +109,27 @@ func enum(g Generator, spec *compile.EnumSpec) error {
 					return nil
 			<end ->
 				default:
-					<$val>, err := <$strconv>.ParseInt(<$s>, 10, 32)
+					<if checkEnumTextMarshalStrict ->
+						return <$fmt>.Errorf("unknown enum value %q for %q", <$s>, "<$enumName>")
+					<else ->
+						<$val>, err := <import "strconv">.ParseInt(<$s>, 10, 32)
```
abhinav, line 115: Why do we need to break UnmarshalText? Doesn't it make sense to be liberal about what we accept here?

And if we do need to break it, why only UnmarshalText and not UnmarshalJSON which also accepts numbers?

abhinav (commented):
Besides the nits, which should be easy to fix:

- Why are we breaking UnmarsahlText? That seems out of scope for what this is meant to accomplish. And if we really need to break UnmarshalText with this flag (which I'd still like an explanation for), would we also not want UnmarshalJSON to replicate the beahvior?
- Please add some tests for the new logic. Just looking over the generated code is not enough. The way tests for these templates work is: we generate code in gen/internal/tests (be sure to update the gen/internal/tests/Makefile), and then we write tests against the generated code. For this code, we would need tests for the MarshalText and MarshalJSON methods of the strict enum, for both success and failure cases.

gen/enum_test.go
```diff
@@ -501,6 +502,69 @@ func TestEnumLabelValid(t *testing.T) {
...
+	}{
+		{"marshalText success",
+			1,
+			false,
+		},
+		{
+			"marshalText unknown value",
+			5,
+			true,
+		},
```
abhinav, line 519: please use named fields here and fix the alignment:

```
{
  testName: "marshalText succeeds",
  enumStrict: ...,
  hasError: false,
},
```

gen/enum_test.go
```diff
@@ -501,6 +502,69 @@ func TestEnumLabelValid(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		t.Run(string(tt.testName), func(t *testing.T) {
+			marshalledValue, err := tt.enumStrict.MarshalText()
+			if tt.hasError {
```
abhinav, line 524: Instead of hasError which replicates the logic for generating an error message, would you mind changing this to a hard-coded string in the test cases that specifies an approximate or exact error message?

Something like:

```
tests := []struct{
  desc string
  give tems.EnumMarshalStrict
  
  want    string // expected MarshalText value
  wantErr string // error message if any
}{

    // ...
}
```

Where only one of `want` and `wantErr` is set per test case. In the test logic itself, we can do something like:

```
got, err := tt.give.MarshalText()
if len(tt.wantErr) > 0 {
    require.Error(t, err)
    assert.Equal(t, tt.wantErr, err.Error())
} else {
    require.NoError(t, err)
    assert.Equal(t, tt.want, string(got))
}
```

This will allow us to validate all cases properly.

And I want to clarify that for `wantErr`, we should *hard-code* the exact string we want to see in the error message. Otherwise we're just replicating the code we wrote in the template, which doesn't verify that the output is legible. For example, if there's a mistake in the fmt.Sprintf template, then both, the test and code being tested will agree, but it won't actually catch the issue.

So for example, something like:

    wantErr: `unknown enum value "1" for "EnumFoo"`

gen/enum_test.go
```diff
@@ -501,6 +502,69 @@ func TestEnumLabelValid(t *testing.T) {
...
+	tests := []struct {
+		testName   string
+		enumStrict tems.EnumMarshalStrict
+		hasError   bool
+		expectedValue string
+	}{
+		{"marshalJSON success",
+			1,
+			false,
+			"\"Bar\"",
```
abhinav, line 545: Please use backticks to avoid escaping quotes in these strings.

```suggestion
			`"Bar"`,
```

gen/enum_test.go
```diff
@@ -501,6 +502,69 @@ func TestEnumLabelValid(t *testing.T) {
...
+				assert.Equal(t, fmt.Sprintf("unknown enum value %q for %q", tt.enumStrict, "EnumMarshalStrict"), err.Error())
+			} else {
+				assert.Nil(t, err)
+				assert.Equal(t, []byte(tt.enumStrict.String()), marshalledValue)
+			}
+		})
+	}
+}
+
+func TestEnumStrictMarshalJSON(t *testing.T) {
```
abhinav, line 535: This entire test case should also look roughly like the MarshalText test case. We might even be able to merge the two by having `wantText` and `wantJSON` instead of just `want` above.

gen/enum_test.go
```diff
@@ -501,6 +502,69 @@ func TestEnumLabelValid(t *testing.T) {
...
+			false,
+		},
+		{
+			"marshalText unknown value",
+			5,
+			true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(string(tt.testName), func(t *testing.T) {
```
abhinav, line 522: testName is already a string. the cast is not needed.
```suggestion
		t.Run(tt.testName, func(t *testing.T) {
```

abhinav (changes requested):
The test could do with some cleanup. Please also fix the linter issues:

```
Checking gofmt
gen/golden_test.go
gen/generate.go
gen/enum_test.go
```

You can verify this locally with `make lint`.

gen/enum_test.go
```diff
@@ -501,6 +502,46 @@ func TestEnumLabelValid(t *testing.T) {
...
+			wantText: "Bar",
+			wantJSON: `"Bar"`,
+			wantErr:  "",
+		},
+		{
+			desc:     "marshalText unknown value",
+			give:     5,
+			wantText: "",
+			wantJSON: "",
+			wantErr:  "unknown enum value \"EnumMarshalStrict(5)\" for \"EnumMarshalStrict\"",
```
abhinav, line 525:
```suggestion
			wantErr:  `unknown enum value "EnumMarshalStrict(5)" for "EnumMarshalStrict"`,
```

gen/enum_test.go
```diff
@@ -501,6 +502,46 @@ func TestEnumLabelValid(t *testing.T) {
...
+			give:     1,
+			wantText: "Bar",
+			wantJSON: `"Bar"`,
+			wantErr:  "",
+		},
+		{
+			desc:     "marshalText unknown value",
+			give:     5,
+			wantText: "",
+			wantJSON: "",
```
abhinav, line 524:
```suggestion
```

gen/enum_test.go
```diff
@@ -501,6 +502,46 @@ func TestEnumLabelValid(t *testing.T) {
...
+		wantText string
+		wantJSON string
+		wantErr  string
+	}{
+		{
+			desc:     "marshalText success",
+			give:     1,
+			wantText: "Bar",
+			wantJSON: `"Bar"`,
+			wantErr:  "",
```
abhinav, line 518:
```suggestion
```

abhinav (approved):

PR #503, opened 2021-06-25
idl: record document positions on constant nodes

### LINKED ISSUE #493: Line numbers for primitive constants

jparise:
This change does two things:

First, it records document positions for parsed nodes that cannot store
their own position data. This primarily applies to primitive constants.

Second, it makes this information available through a new Config-based
parsing interface using an optional idl.Info structure. Node positions
are available though idl.Info.Pos(), which first attempts to return a
node's "native" position (line) value before falling back to its
internal nodePositions map.

Closes #493

### DIFF
--- a/idl/config_test.go
+++ b/idl/config_test.go
@@ -31,7 +31,7 @@ import (
 func TestParse(t *testing.T) {
 	c := &Config{}
 	prog, err := c.Parse([]byte{})
-	if assert.NoError(t, err, "%v", err) {
+	if assert.NoError(t, err) {
 		assert.Equal(t, &ast.Program{}, prog)
 	}
 }
--- a/idl/config_test.go
+++ b/idl/config_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2015 Uber Technologies, Inc.
+// Copyright (c) 2021 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
--- a/idl/config_test.go
+++ b/idl/config_test.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 Uber Technologies, Inc.
+// Copyright (c) 2015 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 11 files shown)


idl/internal/parser.go
```diff
@@ -27,13 +27,13 @@ func init() {
 }
 
 // Parse parses the given Thrift document.
-func Parse(s []byte) (*ast.Program, []ParseError) {
+func Parse(s []byte) (*ast.Program, NodePositions, []ParseError) {
```
jparise, line 30: I don't believe the package is consider "public API", but we could also think about returning a struct value here to avoid ballooning the number of return values over time.

idl/config_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"go.uber.org/thriftrw/ast"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParse(t *testing.T) {
+	c := &Config{}
+	prog, err := c.Parse([]byte{})
+	if assert.NoError(t, err, "%v", err) {
```
abhinav, line 34: nit: the default error generated by testify will include the error message.

```suggestion
	if assert.NoError(t, err) {
```

(I suspect the old code in ThriftRW has some anti-patterns like this one.)

idl/internal/thrift.y
```diff
@@ -386,13 +386,12 @@ base_type_name
...
-    : INTCONSTANT { $$ = ast.ConstantInteger($1) }
-    | DUBCONSTANT { $$ = ast.ConstantDouble($1) }
-    | TRUE        { $$ = ast.ConstantBoolean(true) }
-    | FALSE       { $$ = ast.ConstantBoolean(false) }
-    | LITERAL     { $$ = ast.ConstantString($1) }
+    : INTCONSTANT { $$ = ast.ConstantInteger($1); yylex.(*lexer).RecordPosition($$) }
+    | DUBCONSTANT { $$ = ast.ConstantDouble($1); yylex.(*lexer).RecordPosition($$) }
+    | TRUE        { $$ = ast.ConstantBoolean(true); yylex.(*lexer).RecordPosition($$) }
+    | FALSE       { $$ = ast.ConstantBoolean(false); yylex.(*lexer).RecordPosition($$) }
+    | LITERAL     { $$ = ast.ConstantString($1); yylex.(*lexer).RecordPosition($$) }
```
abhinav, line 394: I suspect this will record the line number right *after* the token.

This is probably not an issue for constant primitives because they'll never be split across multiple lines, but if we want accurate column numbers later, the position would have to be recorded before the value (like with `lineno IDENTIFIER` below). I'm okay doing that when we add support for column numbers, though.

abhinav (approved):
Pretty straightforward. Thanks!

idl/config_test.go
jparise: You'll find this in `idl/parser_test.go`, too.

idl/info.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// Pos returns a node's position in the parsed document.
+func (i *Info) Pos(n ast.Node) Position {
+	if line := ast.LineNumber(n); line != 0 {
+		return Position{Line: line}
+	}
+	pos := i.nodePositions[n]
+	return Position{Line: pos.Line}
+}
```
jparise, line 40: @abhinav I've just realized this will panic if passed an unhashable node that doesn't support `ast.LineNumber` (like `ast.MapType`). Can you think of a pattern to avoid that here aside from using reflection or `recover()`?

abhinav: Oh, yeah, that's a problem. I think we should not ignore the `!= 0` case here: if the type records its own line number, we should return that as-is. That would narrow down the possibility of panic to only those types that are unhashable, but I don't think we have anything that is unhashable that doesn't record its own line number.

To respect the line number of AST nodes that know their own line number, we have two options: `Info` moves to the AST package (which I suspect is not possible without exposing Info.nodePositions), or we add an `ast.LookupLineNumber(ast.Node) (n int, ok bool)` which reports the line number only if available.

jparise: Perhaps this is an opportunity to move `Position` to the `ast` package with an accessor that returns `pos, ok`?

abhinav: That sounds reasonable!

jparise: #504 504

PR #504, opened 2021-06-25
ast: move idl.Position to the ast package

jparise:


ast/position.go
```diff
@@ -18,9 +18,17 @@
...
-package idl
+package ast
 
 // Position represents a position in the parsed document.
 type Position struct {
 	Line int
 }
+
+// Pos attempts to return the position of a Node in the parsed document.
+func Pos(n Node) (Position, bool) {
```
abhinav, line 29: Worth pointing to Info here?


```suggestion
// Pos attempts to return the position of a Node in the parsed document.
// For most use cases, prefer to use idl.Info to access positional information.
func Pos(n Node) (Position, bool) {
```

abhinav (approved):

PR #505, opened 2021-06-25
idl: replace internal.Position with ast.Position

jparise:


idl/internal/position.go
```diff
@@ -22,10 +22,5 @@ package internal
...
 import "go.uber.org/thriftrw/ast"
 
-// Position represents a position in the parsed document.
-type Position struct {
-	Line int
-}
-
 // NodePositions maps (hashable) nodes to their document positions.
-type NodePositions map[ast.Node]Position
+type NodePositions map[ast.Node]ast.Position
```
jparise, line 26: I could also move this elsewhere so we could retire this entire file.

abhinav (approved):

PR #506, opened 2021-06-28
Expose stream protocol method to close Writer

dianale31:
Provides a stream protocol method to close out the resources of the given Writer. 

The corresponding Close() method on Reader will be implemented in a follow-up PR.


protocol/stream/stream.go
```diff
@@ -95,6 +95,7 @@ type Writer interface {
 	WriteSetEnd() error
 	WriteListBegin(l ListHeader) error
 	WriteListEnd() error
+	Close()
```
abhinav, line 98: Thoughts on allowing an `error` return here?

abhinav (approved):

PR #507, opened 2021-06-28
idl: add column numbers to parse error positions

jparise:
Column numbers for parse errors are computed by tracking a new lineStart
value in the lexer that tracks the start position (document byte offset)
of the current line. We can then subtract this value from the existing
token start (`ts`) value to compute the column (+1 for 1-based numbers).

### DIFF
--- a/idl/error.go
+++ b/idl/error.go
@@ -56,7 +56,7 @@ func (pe *ParseError) Error() string {
 	var buffer bytes.Buffer
 	buffer.WriteString("parse error\n")
 	for _, pe := range pe.Errors {
-		buffer.WriteString(fmt.Sprintf("  line %d:%d: %s\n", pe.Pos.Line, pe.Pos.Column, pe.Err))
+		buffer.WriteString(fmt.Sprintf("  line %s: %s\n", pe.Pos, pe.Err))
 	}
 	return buffer.String()
 }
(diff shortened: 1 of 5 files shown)


idl/error.go
```diff
@@ -56,7 +56,7 @@ func (pe *ParseError) Error() string {
 	var buffer bytes.Buffer
 	buffer.WriteString("parse error\n")
 	for _, pe := range pe.Errors {
-		buffer.WriteString(fmt.Sprintf("  line %d: %s\n", pe.Pos.Line, pe.Err))
+		buffer.WriteString(fmt.Sprintf("  line %d:%d: %s\n", pe.Pos.Line, pe.Pos.Column, pe.Err))
```
abhinav, line 59: Consider implementing `String()` on `Position` to print `$line[:$column]`.

```
func (p Position) String() string {
  s := strconv.Itoa(p.Line)
  if c := p.Column; c > 0 {
    s += ":" + strconv.Itoa(c)
  }
  return s
}
```

idl/internal/lex.rl
```diff
@@ -340,7 +344,7 @@ func (lex *lexer) Lex(out *yySymType) int {
     }%%
 
     if lex.cs == thrift_error {
-        lex.AppendError(fmt.Errorf("unknown token at index %d", lex.p))
+        lex.Error("unknown token")
```
abhinav, line 347: nice

abhinav (approved):
Nice, that was pretty straightforward!

idl/error.go
jparise: Funny. I had a note to do that by didn't come back to it over the weekend. I'll add that now.

abhinav (approved):

abhinav:
Thanks!

PR #508, opened 2021-06-29
idl: record full positions for constants

jparise:
We now have enough information to record full positions (line+column)
for constant values.


idl/info.go
```diff
@@ -35,6 +35,5 @@ func (i *Info) Pos(n ast.Node) ast.Position {
 	if pos, ok := ast.Pos(n); ok {
 		return pos
 	}
-	pos := i.nodePositions[n]
-	return ast.Position{Line: pos.Line}
+	return i.nodePositions[n]
```
jparise, line 38: This change should have been part of d7c4d71abe562dc7852dac633a84e1c8130f0985 (#505).

jparise:
The coverage drop doesn't seem accurate. It appears to be reacting to the fact that I replaced two existing lines with one, which is skewing the math.

abhinav (approved):
Thanks!

When we add position information to the other ast.Nodes, we might be able to drop the separate `lineno` component completely. (Or have it use pos to get the line number.)

jparise:
> When we add position information to the other ast.Nodes, we might be able to drop the separate `lineno` component completely. (Or have it use pos to get the line number.)

Right. I've started to make progress in that direction locally. Just looking for a little direction here: https://github.com/thriftrw/thriftrw-go/issues/349#issuecomment-871003807

PR #510, opened 2021-06-30
Add pooling to the stream reader so that resources can be returned to the system

dianale31:
Adds a sync.Pool to the StreamReader implementation to allow callers to borrow a StreamReader and return it after finishing work. 

The binary.Reader implementation continues to instantiate the StreamReader because the binary.Reader.ReadValue method cannot return a StreamReader to the system since methods on container types like ForEach need access to the StreamReader to iterate through items. `newStreamReader` is unexported to make it only usable by internal ThriftRW callers. External callers of the StreamReader should use protocol.BinaryStreamer.Reader(io.Reader) and the StreamReader.Close() method.

### DIFF
--- a/protocol/binary/stream_reader.go
+++ b/protocol/binary/stream_reader.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"io/ioutil"
 	"math"
+	"sync"
 
 	"go.uber.org/thriftrw/internal/iface"
 	"go.uber.org/thriftrw/protocol/stream"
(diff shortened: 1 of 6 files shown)


protocol/binary/stream_reader.go
```diff
@@ -67,11 +68,33 @@ type StreamReader struct {
...
+var streamReaderPool = sync.Pool{
+	New: func() interface{} {
+		return &StreamReader{}
+	}}
+
+// BorrowStreamReader fetches a StreamReader from the system that will write
+// its output to the given io.Reader.
+//
+// This StreamReader must be returned back using ReturnStreamReader.
+func BorrowStreamReader(r io.Reader) *StreamReader {
```
abhinav, line 85: Note to whoever takes over this PR: 
Now that there's a Close method on the interface, we might be able to hide the Borrow-Return as an implementation detail. We can rename BorrowStreamReader to `NewStreamReader` and hide `ReturnStreamReader` (since Close will call it).

abhinav (commented):
this is still in draft mode, but leaving a suggestion for when it's picked up again

abhinav:
Superseded by #514

PR #511, opened 2021-06-30
Define enveloping interfaces for Thrift streaming protocol

usmyth:
Define enveloping interfaces and types for Thrift streaming protocol

Defines a "RequestReader" interface type for the protocol/stream package. The RequestReader interface exposes a single method, "ReadRequest", to provide Thrift handlers a way to transparently stream data off the request (stream.Reader) and stream data to the response (stream.ResponseWriter) while being agnostic to the enveloping style of a given caller (i.e. no-envelope, strict envelope, non-strict envelope).

Defines a "ResponseWriter" interface type for the protocol/stream package. The ResponseWriter interface exposes a single method, WriteResponse, to handle writing the response envelope with the envelope style of the corresponding request. It returns a stream.Writer (borrowed) to the caller to write the response bytes.

Exposes new methods on the "stream.Writer" interface, WriteEnvelope{Begin,End} to support writing an envelope in a strict, versioned style.

### DIFF
--- a/protocol/stream/envelope.go
+++ b/protocol/stream/envelope.go
@@ -1,0 +1,49 @@
+// Copyright (c) 2021 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package stream
+
+import (
+	"io"
+
+	"go.uber.org/thriftrw/wire"
+)
+
+// RequestReader captures how to read from a request in a streaming fashion.
+type RequestReader interface {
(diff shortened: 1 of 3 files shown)


witriew (approved):
Looks fine to me, let's get @abhinav's opinion too.

protocol/stream/envelope.go
```diff
@@ -0,0 +1,61 @@
...
+// metadata about the method, the type of data in the envelope, and the value.
+// It is equivalent of `wire.Envelope`, but for streaming purposes.
+type EnvelopeHeader struct {
+	Name  string
+	Type  wire.EnvelopeType
+	SeqID int32
+}
+
+// EnvelopeAgnosticProtocol is the streaming equivalent of protocol.EnvelopeAgnosticProtocol
+type EnvelopeAgnosticProtocol interface {
```
abhinav, line 39: We put the original DecodeRequest on the new interface (EnvelopeAgnosticProtocol) in the main protocol package for backwards compatibility: we couldn't add the method to the existing Protocol interface because that would constitute a breaking change.

This package does not suffer from this limitation because it's entirely new code, so we can name and align things closer to desirable. Just spitballing here, but how do you folks feel about this?

```
type RequestReader interface {
  ReadRequest(wire.EnvelopeType, io.Reader) (body Reader, res ResponseWriter, err error)
}

type ResponseWriter interface {
  WriteResponse(wire.EnvelopeType, io.Writer) (body Writer, err error)
}
```

It's roughly the same but with the following changes:

- No embedding
- Interface names closer align with https://golang.org/doc/effective_go#interface-names (we don't do this always)

The net effect of this that the low level Protocol interface has a concept of envelopes and envelope headers but the high level RequestReader/ResponseWriter interface doesn't. This theoretically allows for a standard basic catch-all RequestReader/ResponseWriter implementation based on any stream.Protocol. (We wouldn't do this, but we could.)

Binary protocol, which is trying to support three different wire formats (not enveloped, strict envelope, and non-strict envelope), will implement RequestReader/ResponseWriter itself instead of the theoretical catch-all implementation because it needs to have binary-protocol-specific knowledge.

protocol/stream/stream.go
```diff
@@ -95,6 +95,12 @@ type Writer interface {
 	WriteSetEnd() error
 	WriteListBegin(l ListHeader) error
 	WriteListEnd() error
+
+	WriteEnvelopeBegin(eh EnvelopeHeader) error
```
abhinav, line 99: nit: Given that the other Header types that are part of the protocol are defined in this interface, and EnvelopeHeader is also directly used by this interface, can we put EnvelopeHeader in this file?

protocol/stream/stream.go
```diff
@@ -95,6 +95,12 @@ type Writer interface {
...
+	WriteEnvelopeEnd() error
+	WriteLegacyEnvelopeBegin(eh EnvelopeHeader) error
+	WriteLegacyEnvelopeEnd() error
```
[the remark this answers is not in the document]
usmyth: Agreed. Nice catch

protocol/stream/envelope.go
usmyth: Ah, I see. Yes, this is much cleaner. Thanks!

abhinav (approved):

PR #513, opened 2021-07-02
Define an Enveloper interface type for thrift types that can be stream encoded with an envelope

usmyth:
The enveloper/stream Enveloper interface is implemented by a thrift type that can be written with an envelope via streaming (stream.Writer).


dianale31 (approved):
LGTM!

witriew (approved):
Makes sense to me, please wait for one of the core members to approve as well.

abhinav (approved):
Stamping because it aligns with the existing non-streaming interface, but it's not clear where/how this is expected to be used.

PR #515, opened 2021-07-07
Rename 'BorrowStreamWriter' to 'NewStreamWriter'; unexport 'ReturnStreamWriter'

witriew:
With #506 implementing a `Close()` for the `StreamWriter`, change the concepts
around the {con,de}structors to not use 'borrow'ing - hide these behind New and Close.


abhinav (approved):
thanks!

PR #516, opened 2021-07-07
binary: Move Responders and Protocol into package

abhinav:
In #512 it came up that the `*Responder` types in the protocol package
belong in the protocol/binary package because they're binary-specific.
We found that we cannot move the types into the `binary` package because
`EnvelopeAgnosticProtocol.DecodeRequest` returns a `protocol.Responder`,
which causes an import cycle between protocol and protocol/binary.

This change moves the implementation of the binary protocol, all its
methods, and all responder implementations into a binary/protocol.go
file, including the DecodeRequest method and exposes the raw Protocol
type so that we're not limited by an interface.

To get around the import cycle, we redeclare the Responder interface in
the binary package, and adapt the type from a binary.Protocol into a
protocol.Protocol using an unexported type.

Per apidiff, the net effect of these changes is:

```
--- go.uber.org/thriftrw/protocol/binary ---
Compatible changes:
- Default: added
- EnvelopeV0Responder: added
- EnvelopeV1Responder: added
- NoEnvelopeResponder: added
- Protocol: added
- Responder: added
```

So nothing changes in the top-level `protocol` package (besides some
deprecations) and the `binary` package exports a bunch of new types.

Caveat: I would have preferred not to export the `*Responder` types and
return a `Responder` struct instead of the interface, but unfortunately,
ur hands are tied because of exporting the responder implementations
originally. So `binary.Protocol.ReadRequest` has to be able to return
one of the three Responder types, which is what necessitates the
interface.

### DIFF
--- a/protocol/binary.go
+++ b/protocol/binary.go
@@ -31,7 +31,7 @@ import (
 // Binary implements the Thrift Binary Protocol.
 // Binary can be cast up to EnvelopeAgnosticProtocol to support DecodeRequest.
 //
-// Deprecated: binary.Default.
+// Deprecated: Don't use this directly. Use binary.Default.
 var Binary Protocol
 
 // BinaryStreamer implements a streaming version of the Thrift Binary Protocol.
--- a/protocol/binary.go
+++ b/protocol/binary.go
@@ -21,10 +21,8 @@
 package protocol
 
 import (
-	"fmt"
 	"io"
 
-	"go.uber.org/thriftrw/internal/iface"
 	"go.uber.org/thriftrw/protocol/binary"
 	"go.uber.org/thriftrw/protocol/stream"
 	"go.uber.org/thriftrw/wire"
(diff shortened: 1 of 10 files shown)


usmyth (approved):
LGTM

usmyth:
@abhinav should we update the base branch for this PR to be streamdev so that PR #512 can use this?

protocol/binary.go
```diff
@@ -30,6 +29,8 @@ import (
 
 // Binary implements the Thrift Binary Protocol.
 // Binary can be cast up to EnvelopeAgnosticProtocol to support DecodeRequest.
+//
+// Deprecated: binary.Default.
```
witriew, line 33:
```suggestion
// Deprecated: Don't use this directly. Use binary.Default.
```

witriew (approved):

abhinav:
> @abhinav should we update the base branch for this PR to be streamdev so that PR #512 can use this?

Sounds good. I figured we would cherry pick it into streamdev but either works.

abhinav:
@usmyth rebased on streamdev and merged. There were enough conflicts that cherry-picking would have been annoying.

PR #521, opened 2021-07-08
idl: account for tabs in column calculations

jparise:
Now that we're tracking column position, we need to deal with tabs. This
takes a little extra work because tabs can't simply be translated to
e.g. 8 additional characters; they instead advance the "virtual" column
forward to the next tab stop.

We use a 8-character tab width by default, but this could be made
configurable in the future.

### DIFF
--- a/idl/internal/lex.rl
+++ b/idl/internal/lex.rl
@@ -25,6 +25,8 @@ type lexer struct {
 
     line int
     lineStart int
+    tabExtra int
+    tabWidth int
 
     docstringStart int
     lastDocstring string
@@ -44,6 +46,7 @@ type lexer struct {
 func newLexer(data []byte) *lexer {
     lex := &lexer{
         line: 1,
+        tabWidth: 8,
         nodePositions: make(NodePositions, 0),
         parseFailed: false,
         data: data,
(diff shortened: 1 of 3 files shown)


idl/internal/lex.rl
```diff
@@ -71,17 +74,28 @@ func (lex *lexer) Lex(out *yySymType) int {
...
+            lex.tabExtra += lex.tabWidth - (lex.te - lex.lineStart + lex.tabExtra) % lex.tabWidth - 1
+        };
 
         # All uses of \n MUST use this instead if we want accurate line
-        # number tracking.
+        # and column tracking.
         newline = '\n' >{
             lex.line++
-            lex.lineStart = lex.p + 1
+            lex.lineStart = lex.te + 1
```
jparise, line 93: This doesn't change anything in practice, but it's slightly more accurate to refer to the "end of the newline character" here rather than the current "pointer" in the data buffer.

idl/parser_test.go
```diff
@@ -123,7 +131,7 @@ func TestParseErrors(t *testing.T) {
 				enum Foo {}
 				include "bar.thrift"
 			`,
-			wantErrors: []string{"line 3:5", "unexpected INCLUDE"},
+			wantErrors: []string{"line 3:33", "unexpected INCLUDE"},
```
jparise, line 134: These cases aren't are obvious because of the way Go's multiline string literals work in combination with the indentation we're using here.

I think this is fine, but we could either skip the indentation or use something like https://github.com/MakeNowJust/heredoc/.

idl/internal/lex.rl
```diff
@@ -44,6 +46,7 @@ type lexer struct {
 func newLexer(data []byte) *lexer {
     lex := &lexer{
         line: 1,
+        tabWidth: 8,
```
jparise, line 49: 8 feels like the right value because it's (a) the definition of a tab and (b) it's what most terminals will render.

I expect many people use 4-column tabs in their editors, though.

abhinav (commented):
I feel like the conversion of tabs here is lossy and I'm reluctant to lose that information. That is, we'll lose the real information that "this entity starts at byte offset n-1 on line X". Translation of tabs to spaces appears to be a presentation-level concern.

What do you think about making the TabWidth configurable from the start, and defaulting it to 1? We can use the existing idl.Config struct for this.

```
package idl

type Config struct {
  Info *ast.Info
  TabWidth int  // defaults to 1
}
```

(We'll have to modify internal.Parse to accept an additional parameter but that's fine.)

That way, the default behavior will be to provide positions based on treating tabs as single characters, which maps easily to byte offsets on individual lines. And for your case, where column numbers likely affect presentation, you can supply a `Config{TabWidth: 8}`. 

Does that sound reasonable?

jparise:
> I feel like the conversion of tabs here is lossy and I'm reluctant to lose that information. That is, we'll lose the real information that "this entity starts at byte offset n-1 on line X". Translation of tabs to spaces appears to be a presentation-level concern.

...

> That way, the default behavior will be to provide positions based on treating tabs as single characters, which maps easily to byte offsets on individual lines. And for your case, where column numbers likely affect presentation, you can supply a `Config{TabWidth: 8}`.
> 
> Does that sound reasonable?

I think I'm going to sit on this for now. After thinking it through some more, I agree that it makes sense to continue treating tabs as a single byte by default, and I'm not sure I'd change that at a higher level either (in which case we don't need this support at all).

abhinav:
@jparise If you don't mind, I'm going to close this PR for now since we're not sure about whether this change is desirable. If you end up needing tab expansion, we can revive this and incorporate it into idl.Config. Thanks!

PR #522, opened 2021-07-09
ast: add column values to the AST nodes

### LINKED ISSUE #349: ast.DefinitionInfo should include the location of the source code

jparise:
The nodeWithLine interface has been retired in favor of
nodeWithPosition. ast.Pos() is the generic way to retrieve
a node's position, and ast.LineNumber() has been revised
to work in terms of the nodeWithPosition interface.

Closes #349


idl/internal/thrift.y
```diff
@@ -358,18 +375,18 @@ throws
...
-        { $$ = ast.SetType{ValueType: $4, Annotations: $6, Line: $1} }
-    | lineno IDENTIFIER
-        { $$ = ast.TypeReference{Name: $2, Line: $1} }
+    | pos MAP '<' type ',' type '>' type_annotations
+        { $$ = ast.MapType{KeyType: $4, ValueType: $6, Annotations: $8, Line: $1.Line, Column: $1.Column} }
+    | pos LIST '<' type '>' type_annotations
+        { $$ = ast.ListType{ValueType: $4, Annotations: $6, Line: $1.Line, Column: $1.Column} }
+    | pos SET '<' type '>' type_annotations
+        { $$ = ast.SetType{ValueType: $4, Annotations: $6, Line: $1.Line, Column: $1.Column} }
+    | lineno pos IDENTIFIER
```
jparise, line 388: I could use some guidance here.

This is the last remaining occurrence of `lineno`. It's value isn't used by this expression, but when I remove it, I start getting incorrect column numbers in the resulting nodes. There's something "off" about the way the parser generates the code in that case, but I haven't been able to debug the root cause yet.

Any ideas would be welcome. I'm hopefully just overlooking something simple.

jparise: I think there's now a parsing ambiguity between `type: pos base_type_name type_annotations` and `type: pos IDENTIFIER` when the `lineno` term is removed from the match.

abhinav: This is weird. It appears to be picking the position of the field rather than the position of the type.

jparise: `pos IDENTIFIER` was always taking precedence over `pos base_type_name`. I was able to "fix" that by using `IDENTIFIER pos` in this one place, in the interest of practicality over purity.

I tried yacc's various precedence-controlling features without luck. I think something like [Selection Preferences](https://www.ibm.com/docs/en/zos/2.3.0?topic=topics-selection-preference-rules) would be ideal here, but goyacc doesn't support them.

```yacc
| ps IDENTIFIER [ ^ BOOL BYTE I8 I16 I32 I64 DOUBLE STRING BINARY ]
```

I couldn't think of another way to reorganize the grammar itself to avoid this situation, but I'm definitely open to ideas.

jparise: I also tried to "trick" the grammar using an intermediate type, but I (helpfully) collapsed it, resulting in the same original problem:

```yacc
typeref
    : pos IDENTIFIER
        { $$ = ast.TypeReference{Name: $2, Line: $1.Line, Column: $1.Column} }
```

abhinav (approved):
Apologies for the delay in reviewing. Thanks for the change! This is a reasonable workaround. I was unsure of how to address the weirdness, and didn't have too much time to dig into it.

PR #524, opened 2021-07-12
Responder: Deduplicate interface

abhinav:
In #516, we tried to move binary-specific concerns into the binary
subpackage. Unfortunately, this has the effect of declaring two
different `Responder` interfaces:

    protocol/
      envelope.go     // type Responder
      binary/
        protocol.go   // type Responder

Although these two interfaces are equivalent, and any type that
implements one implements the other, this introduced a problem: an
interface that has a method that *returns* one of these interfaces does
not automatically support the other interface. That is,
`*binary.Protocol`, which returns a `binary.Responder` does not
implement `protocol.EnvelopeAgnosticProtocol`, which returns a
`protocol.Responder`.

    // compilation error!
    var p protocol.EnvelopeAgnosticProtocol = binary.Default

To address this, we introduced an adapter in the `protocol` package that
adapts a `*binary.Default` into a `protocol.EnvelopeAgnosticProtocol` by
altering the return type of the `DecodeRequest` method.

This isn't great, especially because this means that we either need this
adapter everywhere, or we need to remember to use
`EnvelopeAgnosticBinary` in different places. In particular, it breaks
code like the following:

    var proto protocol.Protocol = binary.Default
    if p, ok := protocol.(protocol.EnvelopeAgnosticProtocol); ok {
        p.DecodeRequest(..)
    }

Because `binary.Default` (of type `*binary.Protocol`) does not implement
that interface.

More importantly, this breaks the following code as well, which
previously worked just fine:

    var proto protocol.Protocol = protocol.Binary
    if p, ok := protocol.(protocol.EnvelopeAgnosticProtocol); ok {
        p.DecodeRequest(..)
    }

To address these problems, declare the `Responder` interface in a new
subpackage of `protocol`, and alias the `protocol.Responder` to that
interface.

With this, `*binary.Protocol` now implements `EnvelopeAgnosticProtocol`

### DIFF
--- a/protocol/binary.go
+++ b/protocol/binary.go
@@ -21,11 +21,8 @@
 package protocol
 
 import (
-	"io"
-
 	"go.uber.org/thriftrw/protocol/binary"
 	"go.uber.org/thriftrw/protocol/stream"
-	"go.uber.org/thriftrw/wire"
 )
 
 // Binary implements the Thrift Binary Protocol.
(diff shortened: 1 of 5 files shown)


usmyth:
Lint failures:

`thrifttest/mock_protocol.go:170:109: protocol.Responder is deprecated: Use envelope.Responder.  (SA1019)
`

abhinav:
Derp. Fixing.

witriew (approved):

PR #526, opened 2021-07-20
stream: Implement Request and Response handling with Enveloping

witriew:
Based off of the implementation by @usmyth for enveloping, perform reading out a
request as well as providing an appropriate `stream.ReponseWriter`, based on the
enveloping that was included on the request itself.

In both cases of reading out the request as well as providing a
`stream.ResponseWriter`, the interfaces for these needed to change to support
doing so with enveloping.  Since the envelope's begin and end are meant to
encapsulate the request and response body, an object needs to be provided that
understands how to parse, or write out these bodies respectively.

For the request flow, this object implements a `stream.BodyReader` to parse out
the body and should be a thriftrw object that reflects the request's body.  For
the response flow, a `stream.Enveloper` that reflects the response's body and is
also a compatible thriftrw object, must be provided.

### DIFF
--- a/protocol/binary/protocol.go
+++ b/protocol/binary/protocol.go
@@ -195,24 +195,29 @@ func (p *Protocol) DecodeRequest(et wire.EnvelopeType, r io.ReaderAt) (wire.Valu
 // 15 valid types today).
 //
 // Callers must call Close() on the stream.Reader once finished.
-func (p *Protocol) ReadRequest(et wire.EnvelopeType, r io.Reader) (stream.Reader, stream.ResponseWriter, error) {
+func (p *Protocol) ReadRequest(et wire.EnvelopeType, r io.Reader, bodyFunc stream.ReadBodyFunc) (stream.ResponseWriter, error) {
 	var buf [2]byte
 
 	if count, _ := r.Read(buf[0:2]); count < 2 {
-		return p.Reader(bytes.NewReader(buf[:])), NoEnvelopeResponder, nil
+		return NoEnvelopeResponder, bodyFunc(p.Reader(bytes.NewReader(buf[:])))
 	}
 
-	// reset the Reader to properly read the envelope, if it exists;
+	// Reset the Reader to allow for properly reading the envelope, if it exists.
 	r = io.MultiReader(bytes.NewReader(buf[:]), r)
 	sr := p.Reader(r)
+	defer sr.Close()
 
 	if buf[0] == 0x00 {
 		eh, err := p.readEnvelopeHeader(sr, et)
 		if err != nil {
-			sr.Close()
-			return nil, NoEnvelopeResponder, err
+			return NoEnvelopeResponder, err
 		}
-		return sr, &EnvelopeV0Responder{
+
+		if err := bodyFunc(sr); err != nil {
+			return NoEnvelopeResponder, err
+		}
+
+		return &EnvelopeV0Responder{
 			Name:  eh.Name,
 			SeqID: eh.SeqID,
 		}, nil
@@ -221,18 +226,22 @@ func (p *Protocol) ReadRequest(et wire.EnvelopeType, r io.Reader) (stream.Reader
 	if buf[0]&0x80 > 0 {
 		eh, err := p.readEnvelopeHeader(sr, et)
 		if err != nil {
-			sr.Close()
-			return nil, NoEnvelopeResponder, err
+			return NoEnvelopeResponder, err
 		}
-		return sr, &EnvelopeV1Responder{
+
+		if err := bodyFunc(sr); err != nil {
+			return NoEnvelopeResponder, err
+		}
+
+		return &EnvelopeV1Responder{
 			Name:  eh.Name,
 			SeqID: eh.SeqID,
 		}, nil
 	}
 
 	// For anything else, the request is either not-enveloped or invalid, let the
-	// caller manage that data
-	return sr, NoEnvelopeResponder, nil
+	// bodyFunc manage that data.
+	return NoEnvelopeResponder, bodyFunc(sr)
 }
 
 func (p *Protocol) readEnvelopeHeader(sr stream.Reader, et wire.EnvelopeType) (stream.EnvelopeHeader, error) {
--- a/protocol/binary/protocol.go
+++ b/protocol/binary/protocol.go
@@ -21,6 +21,7 @@
 package binary
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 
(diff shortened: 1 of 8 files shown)


witriew:
> For the response flow, a `stream.Enveloper` is needed that can write
out the appropriate objects out to the wire.

Doing it this way is potentially controversial.  The signature of `stream.ResponseWriter` could potentially reflect the `Handle` call and take in a callback similarly that consumes a `stream.Writer`.  The concern here is that a `ResponseWriter` is specifically meant to handle writing out the body and and a `stream.Enveloper` reflects this exactly.  Doing so otherwise would introduce another layer of complexity by probably requiring a `ResponseHandle` interface that defines something similar to:
```
HandleResponse(context.Context, stream.Writer) error
```

protocol/binary/protocol.go
```diff
@@ -161,54 +166,105 @@ func (p *Protocol) DecodeRequest(et wire.EnvelopeType, r io.ReaderAt) (wire.Valu
...
+				Name:  eh.Name,
+				SeqID: eh.SeqID,
+			}
+		case buf[0]&0x80 > 0:
+			responder = &EnvelopeV1Responder{
+				Name:  eh.Name,
+				SeqID: eh.SeqID,
+			}
+		default:
+			return responder, fmt.Errorf("should never happen")
```
dianale31, line 239: Should we just add a proper error string, just in case?

protocol/binary/protocol.go
```diff
@@ -161,54 +166,105 @@ func (p *Protocol) DecodeRequest(et wire.EnvelopeType, r io.ReaderAt) (wire.Valu
...
+		}
 
-// EnvelopeV1Responder responds to requests with a strict, version 1 envelope.
-type EnvelopeV1Responder struct {
-	Name  string
-	SeqID int32
+		err = body.Decode(sr)
+		if err != nil {
```
dianale31, line 243: nit: collapse into `if err := body.Decode(sr); err != nil`

dianale31 (approved):
LGTM!

protocol/binary/protocol.go
```diff
@@ -161,54 +166,105 @@ func (p *Protocol) DecodeRequest(et wire.EnvelopeType, r io.ReaderAt) (wire.Valu
...
+	r = io.MultiReader(bytes.NewReader(buf[:]), r)
+	sr = p.Reader(r)
+	defer sr.Close()
 
-// EnvelopeV0Responder responds to requests with a non-strict (unversioned) envelope.
-type EnvelopeV0Responder struct {
-	Name  string
-	SeqID int32
-}
+	if buf[0] == 0x00 || buf[0]&0x80 > 0 {
```
abhinav, line 220: I refactored this a bit to match the shape of the DecodeRequest method that this is modeled after.

protocol/binary/protocol.go
```diff
@@ -161,54 +166,117 @@ func (p *Protocol) DecodeRequest(et wire.EnvelopeType, r io.ReaderAt) (wire.Valu
...
+	body stream.BodyReader,
+) (stream.ResponseWriter, error) {
+	var buf [2]byte
 
-func (noEnvelopeResponder) EncodeResponse(v wire.Value, t wire.EnvelopeType, w io.Writer) error {
-	return Default.Encode(v, w)
-}
+	sr := p.Reader(bytes.NewReader(buf[:]))
+	if count, _ := r.Read(buf[0:2]); count < 2 {
+		return NoEnvelopeResponder, body.Decode(sr)
```
abhinav, line 209: This bit is a bit off. We don't need to instantiate sr until we get inside the if block, and at then, we should still close the reader so that it gets returned to the pool.

Separately, `buf[:]` isn't right. If we read only one of the two bytes, that's all we're going to want to parse. I'll fix this bit.

protocol/binary/protocol.go
```diff
@@ -161,54 +166,121 @@ func (p *Protocol) DecodeRequest(et wire.EnvelopeType, r io.ReaderAt) (wire.Valu
...
+		defer sr.Close()
+		return NoEnvelopeResponder, body.Decode(sr)
+	}
 
-// NoEnvelopeResponder responds to a request without an envelope.
-var NoEnvelopeResponder = &noEnvelopeResponder{}
+	// Reset the Reader to allow for properly reading the envelope if it
+	// exists.
+	r = io.MultiReader(bytes.NewReader(buf[:]), r)
+	sr := p.Reader(r)
```
abhinav, line 218: There's a little room for optimization here. Fixing in #531.

protocol/binary/protocol.go
```diff
@@ -161,54 +166,121 @@ func (p *Protocol) DecodeRequest(et wire.EnvelopeType, r io.ReaderAt) (wire.Valu
...
+// and masking out the MSB indicates the version number.
+// At this time, there is only one version.
+//
+// 4. A message of length >1, where the first byte is >=0 can only be a bare
+// struct starting with that field identifier. Valid field identifiers today
+// are in the range 0x00-0x0f. There is some chance that a future version of
+// the protocol will add more field types, but it is very unlikely that the
+// field type will flow into the MSB (128 type identifiers, starting with the
+// 15 valid types today).
+func (p *Protocol) ReadRequest(
```
abhinav, line 199: I'm not 100% into this duplicating DecodeRequest logic, but after some exploration, it's not quite straightforward to resolve this. I might look into that separately, but doesn't need to be a blocker for this.

protocol/binary_test.go
```diff
@@ -1485,6 +1556,173 @@ func TestBinaryEnvelopeSuccessful(t *testing.T) {
 	}
 }
 
+func TestStreamingEnvelopeErrors(t *testing.T) {
```
abhinav, line 1559: good tests

protocol/binary_test.go
```diff
@@ -1485,6 +1556,173 @@ func TestBinaryEnvelopeSuccessful(t *testing.T) {
...
+			responder, err := binary.Default.ReadRequest(context.Background(), tt.inputEnvType, bytes.NewReader(tt.inputReqBytes), h)
+			require.NoError(t, err, "failed to read request with envelope")
+			require.True(t, tt.wantResponderType == reflect.TypeOf(responder), "read request should have responder want %v got %T", tt.wantResponderType, responder)
+
+			// Verify whether we can read the correct request field ID from the stream.Reader
+			// This is a basic test to ensure the stream.Reader is at the correct offset in the request.
+			// It is not intended to test the functionality of stream.Reader
+			//fh := readStructFieldHeader(t, reader)
+			//require.Equal(t, tt.wantReqFieldID, fh.ID, "request field ID mismatch")
+			//require.NoError(t, reader.Skip(fh.Type), "failed to skip")
```
abhinav, line 1712: ?

abhinav (approved):

protocol/binary_test.go
witriew: Cleaned it up, corrected the test.

protocol/binary/protocol.go
witriew: Fixed with https://github.com/thriftrw/thriftrw-go/pull/526/commits/6fc9df673c9f6564b0c25280e8a9dc8b9de29c61

PR #527, opened 2021-07-22
thrifttest: Add mocks for streaming interfaces

witriew:


thrifttest/stream/gen.go
```diff
@@ -0,0 +1,23 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package thrifttest
```
abhinav, line 21: Package name here doesn't match basename of the directory.

Also, did you mean to make this thrifttest/streamtest, BTW? The reason for the `*test` suffix is so that you can import both, `foo` and `footest` in the same code without import aliasing.

witriew: > Also, did you mean to make this thrifttest/streamtest, BTW? 

I did, yes, thanks for that catch.

abhinav (approved):

thrifttest/streamtest/mock_stream.go
```diff
@@ -0,0 +1,960 @@
...
+}
+
+// Writer indicates an expected call of Writer
+func (mr *MockProtocolMockRecorder) Writer(arg0 interface{}) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "Writer", reflect.TypeOf((*MockProtocol)(nil).Writer), arg0)
+}
+
+// private mocks base method
+func (m *MockProtocol) private() {
```
abhinav, line 88: So turns out this won't work. This MockProtocol does not implement the stream.Protocol interface because of the `private()` method.

witriew: You're correct and I'm ahead of you on this :).

There're currently two branches that I'm using to dev from:
https://github.com/thriftrw/thriftrw-go/tree/witriew/unprivate_ifaces
https://github.com/thriftrw/thriftrw-go/tree/witriew/embed_private_mock

The branch names are sort of self-explanatory.  If the interfaces that we decided here look good enough for releasing, I'll use the former branch.  Otherwise, I'll use the latter branch to make them implement `iface.Private`, but this means that we'll need to do this for the mock generation each time.

witriew: I created #532 to resolve this.

PR #529, opened 2021-07-26
Upgrade to golang.org/x/tools version 0.1.5

jparise:
This bumps the project to the latest golang.org/x/tools release (0.1.5).

There don't appear to be any functional changes that impact our usage,
but there are some goyacc formatting differences, so it seemed worth
getting this upgrade out of the way while it's convenient to do so.


abhinav (approved):
Thanks!

PR #533, opened 2021-08-13
streaming: Unembed iface.Private in streaming-based interfaces

witriew:
`iface.Private` was embedded in the newly defined interfaces for streaming to
prevent others from implementing them incase they needed to be modified.

Since the all of the development for streaming has occurred outside of 'dev',
and since it's unlikely for these interfaces to be merged to 'dev' without
already certifying and agreeing on them, unembed these to simplify mocking uses
and to clear out extra code.


abhinav (approved):
We can also delete the iface package now, yes?

PR #537, opened 2021-08-23
ast: add Annotations(Node) []*Annotation

jparise:
It's useful to have a single way to retrieve an arbitrary Node's
annotations without resorting to type switches or reflection.

This change introduces a nodeWithAnnotations interface and an exported
Annotations() function. All Node types which support annotations support
this interface.


abhinav (approved):
Sounds reasonable. Thanks!

PR #538, opened 2021-08-26
Introduce thriftbreak, a backwards compatibility linter

shirchen:
Here we introduce thriftbreak, a linter that takes two files and detects the following four cases:
- a required field was added to an existing struct
- a field was changed from optional to required for an existing struct
- a service was deleted
- one of service's methods was deleted

Note, one of the limitations of this linter is that we cannot yet detect changes if a struct was renamed and then a required field was added.

As a followup, we may introduce better integration with git, where for a Thrift file, we will try and lookup its prevision state in a past commit. 

Ref GO-835


abhinav (changes requested):
We need to consider the git situation sooner rather than later. Even if we put it in an cmd/x/ directory to indicate experimental, I'd like the CLI experience to be better fleshed out. As it stands, this won't help us find breaking changes across imports, for example.

If you'd like to work on the breaking change detection logic independently of the CLI experience, we can break this into two PRs: one internal package that attempts to find incompatibilities between `compile.Module` objects, and one that builds the full CLI around it, including the compilation.

For inspiration of API and CLI:

- https://pkg.go.dev/golang.org/x/exp/apidiff is a similar library-level component that finds incompatibilities in two compiled packages (compile.Module is our rough analog of Go's types.Package)
- https://cs.opensource.google/go/x/exp/+/b4e88ed8:cmd/apidiff/main.go is the CLI for it. apidiff has an interesting mode of operation: you invoke it on the before state with the `-w $FILE` flag to record the information it needs, then on the after state, and then with both files to compare:

    ```
    git checkout old
    apidiff -w before.data example.com/path/to/package
    
    git checkout new
    apidiff -w after example.com/path/to/package
    
    apidiff old new
    ```

  We probably won't have the same CLI experience, but we should definitely put some more thought into it.

cmd/thriftbreak/main.go
```diff
@@ -0,0 +1,78 @@
...
+	fromFile := fs.String("from_file", "", "original file")
+	if err := fs.Parse(args); err != nil {
+		return err
+	}
+
+	if *toFile == "" {
+		return fmt.Errorf("must provide an updated Thrift file")
+	}
+	// TODO: here we'll get into interesting stuff with git, but for now default to
+	// using originalFile as an argument.
```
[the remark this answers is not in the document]
shirchen: either provide `--to` and `--from` or `--repo` argument.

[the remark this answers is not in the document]
abhinav: How about `--base $hash` to specify the base to compare against? It could default to `HEAD^` (one commit before now) or alternatively `HEAD` (most recent commit). The latter could be useful to compare a dirty working tree against the most recent commit. If we call it `--base`, then the "destination" should probably be called `--head` (base and head), and that can default to "current working tree".

So like, locally:

```
$ git status
[clean]
$ vim foo.thrift
[modify]
$ thriftbreak
# Will compare dirty working tree to HEAD which was the last commit.
foo.thrift:2:great sadness
```

And in CI:

```
$ git status
[clean]

$ base=$(git rev-parse HEAD)

# apply the patch
$ arc patch D1234   # or
$ gh pr checkout 123

$ head=$(git rev-parse HEAD)

$ thriftbreak --base "$base" --head "$head"
foo.thrift:2:great sadness
```

abhinav (commented):
I think there may have been some miscommunication in the expectations for the Git integration.
It shouldn't special case current commit and previous commit.
It also doesn't need to look at diffs and changed files.

The way I see it, git should be:

```
thriftbreak --base $sha foo.thrift bar.thrift
# Finds breaking changes made to foo.thrift and bar.thrift and their imports
# as they exist in the working tree
# compared with their state at Git hash $sha.

thriftbreak --base $sha1 --head $sha2 foo.thrift bar.thrift
# Finds breaking changes made to foo.thrift and bar.thrift and their imports
# as they exist at commit $sha2
# compared with their state at commit $sha1.

thriftbreak -C path/to/repo --base $sha foo.thrift
# Same as the first one, except the files are not in the current directory.
```

So the options are:

```
-C dir
    path to the directory containing these files.
    Defaults to the current working directory.
--base SHA
    git hash of the base commit to compare against.
    Defaults to HEAD.
--head SHA
    git hash of the head commit to compare against.
    Uses the current working tree if unspecified.
```

So the `-C` flag is much like `tar` and `git`'s `-C` flag: go to that directory before doing anything, but default to current.

This alone unlocks a few basic uses cases:

- Local development: make a change and
  run thriftbreak without committing anything
  
    ```
    $ vim foo.thrift  # make changes
    $ thriftbreak  # no arguments = compare HEAD with working tree
    $ git commit
    ```

- CI: patch the change and compare it with the base revision. For Phabricator:

    ```
    $ arc patch D123
    $ thriftbreak --base origin/main --head arcpatch-D123
    ```
    
Whether we want to support a non-git path is up for discussion.
I don't see much value to it because "copy v1.thrift to v2.thrift and modify" is pretty atypical usage. But if we do support it, code organization wise, I would expect the bulk of the code to be exactly the same with a different interface injected at exactly one point in the code.

For example,

    var beforeFS, afterFS compile.FS
    if /* git path */ {
        beforeFS = ...  // FS based on --base
        afterFS = ...   // FS based on --head
    } else {
        beforeFS = ...  // FS based on --from
        afterFS = ...   // FS based on --to
    }
    compare(beforeFS, afterFS)

If you think that complexity is not worth it,
I would vote for dropping the non-git path until we need it.

---

Separately, for the lint messages:

- We should know the file name, line number, and possibly column number for each error.  Without this, surfacing these messages in CI will lead to a not-very-good user experience relying on posting comments and scraping through logs.
- I don't think lint messages should be based on `error` objects.
  That's a pretty lossy conversion. We should stick with rich structured objects "posted" to an abstraction responsible for catching lint messages.
  Prior art: [go/analysis.Pass.Report](https://pkg.go.dev/golang.org/x/tools/go/analysis#Pass.Reportf) posts    [Diagnostic](https://pkg.go.dev/golang.org/x/tools/go/analysis#Diagnostic) objects to the destination. In #294, I had a similar "lint catcher" type that would receive `Lint` objects.
  This would also let us make the position information part of the Lint/Diagonistic.
  The presentation component can decide how to surface that to the user.
  For the CLI tool, it would be "print to stderr in a line-delimited fashion".
  For, say, a CI integration, it would be "make API requests to GitHub/Phabricator with exact line/column numbers to annotate them in the PR/Diff."

internal/compare/compare.go
```diff
@@ -0,0 +1,177 @@
...
+					Err: changOptToReqError{
+						field: toField.ThriftName(),
+						struc: to.ThriftName(),
+					},
+				})
+			}
+		} else if toField.Required {
+			p.Report(Diagnostic{
+				File: file,
+				Err:  addReqError{toField.ThriftName(), to.ThriftName()},
```
abhinav, line 149: Another reason to not use `error` objects for these: readabilitiy.
I have no way of knowing what these issues are without going to the error definition.

internal/git/git.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/utils/merkletrie"
+	"go.uber.org/thriftrw/compile"
+	"go.uber.org/thriftrw/internal/compare"
+)
+
+// NewGitFS creates an implementation of FS to use git to discover
+// Thrift changes and previous version of a Thrift file.
+func NewGitFS(gitDir string, repo *git.Repository, from bool) *FS {
+	if repo == nil {
```
abhinav, line 37: Why would repo be nil? All usages appear to be invoking it with a valid repo.

internal/git/git.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/utils/merkletrie"
+	"go.uber.org/thriftrw/compile"
+	"go.uber.org/thriftrw/internal/compare"
+)
+
+// NewGitFS creates an implementation of FS to use git to discover
+// Thrift changes and previous version of a Thrift file.
+func NewGitFS(gitDir string, repo *git.Repository, from bool) *FS {
```
abhinav, line 36: ordering: this is effectively a constructor for `*FS`. Let's place it under the type declaration.

ref: https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering

internal/git/git.go
```diff
@@ -0,0 +1,234 @@
...
+		repo:   repo,
+		gitDir: gitDir,
+		from:   from,
+	}
+}
+
+// Compare takes a path to a git repository and returns errors between HEAD and HEAD~
+// for any incompatible Thrift changes between the two shas.
+func Compare(path string) (compare.Pass, error) {
+	pass := compare.Pass{}
```
abhinav, line 58: nit: use var form here

https://github.com/uber-go/guide/blob/master/style.md#use-var-for-zero-value-structs

internal/git/git.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+// findChangedThrift reads a git repo and finds any Thrift files that got changed
+// between HEAD and previous commit.
+func findChangedThrift(gitDir string) ([]*change, error) {
+	// Init git repo.
+	r, err := git.PlainOpenWithOptions(gitDir, &git.PlainOpenOptions{
+		DetectDotGit:          true,
+		EnableDotGitCommonDir: true,
+	})
```
abhinav, line 182: we already have a repo open by the time we call this. can't we pass that repository object here?

internal/git/git.go
```diff
@@ -0,0 +1,234 @@
...
+
+	// Get Repo's HEAD
+	refHead, err := r.Head() // *plumbing.Reference
+	if err != nil {
+		return nil, err
+	}
+	// Look at the log. Can't use filters as they need exact path which
+	// we don't know yet.
+	commitIter, err := r.Log(&git.LogOptions{From: refHead.Hash()})
+	if err != nil {
```
abhinav, line 195: Actually you don't need Log either.
Here's what you can do from the HEAD ref:

```
head, err := r.Head() 
// handleError(err)

// Use the ref to get the head commit:
commit, err := repo.CommitObject(head.Hash())
// handleError(err)

parentCommit, err := ...
// the rest of what you're already doing
```

internal/git/git.go
```diff
@@ -0,0 +1,234 @@
...
+	for _, o := range objects {
+		a, err := o.Action() // Insert, delete or modify.
+		if err != nil {
+			return nil, err
+		}
+		from, _, _ := o.Files()
+		if filepath.Ext(from.Name) == ".thrift" {
+			changed = append(changed, &change{
+				file:   o.From.Name,
+				change: a,
```
abhinav, line 228: if we're detecting renames, should changed have `from` and `to`?
or are we deferring that?

internal/git/git.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+// FS holds reference to components needed for git FS.
+type FS struct {
+	repo   *git.Repository
+	gitDir string
+	from   bool // Whether we are looking for previous version.
+}
+
+func (fs FS) findCommit() (*object.Commit, error) {
```
abhinav, line 107: This combination of `from` on FS + having to find the commit again seems like a missed opportunity.
The `FS` implementation should only need the `*object.Tree` you're working off of.
We already track down the tree. so we should construct the FS from that.
The findCommit can be dropped, and Read can be shortened.

internal/git/git.go
```diff
@@ -0,0 +1,234 @@
...
+
+	// p will have lints as a field which we can sort in cli.
+
+	return pass, errs
+}
+
+// FS holds reference to components needed for git FS.
+type FS struct {
+	repo   *git.Repository
+	gitDir string
```
abhinav, line 103: naming: this is called git Dir but it's actually the repository root.
In git plumbing contexts, git dir refers to the ".git" directory.

internal/git/git.go
```diff
@@ -0,0 +1,234 @@
...
+	commit, err := fs.findCommit()
+
+	if err != nil {
+		return nil, err
+	}
+
+	// filename is going to be the full path. We don't want that.
+	filename, err := filepath.Rel(fs.gitDir, path)
+	if err != nil {
+		return nil, err
```
abhinav, line 144: consider adding a comment that this is impossible

internal/git/git.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	// It's possible that file was deleted and it will not exist.
+	f, err := commit.File(filename)
+	if err != nil {
+		return nil, err
```
abhinav, line 149: might want to include more information here:

```
if err != nil {
  return nil, fmt.Errorf("open %q: %w", filename, err)
```

(because File will only return "file not found".
https://github.com/go-git/go-git/blob/v5.4.2/plumbing/object/tree.go#L76)

abhinav (commented):
good starting point if we're narrowing down to just the `-C` flag for now

internal/git/git.go
```diff
@@ -0,0 +1,234 @@
...
+	fsFrom := NewGitFS(path, r, true)
+
+	changed, err := findChangedThrift(path)
+	if err != nil {
+		return pass, err
+	}
+	var errs error
+	for _, c := range changed {
+		var toModule *compile.Module
+		if c.change == merkletrie.Modify {
```
r-hang, line 76: non-comment remark: nifty. For idl-observer, I literally parsed the output of git status to determine affected files. I didn't know about this merkletrie.

internal/git/git_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/go-git/go-git/v5/utils/merkletrie"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/thriftrw/internal/breaktest"
+)
+
+func TestOpenRepo(t *testing.T) {
+	t.Parallel()
+	tmpDir := t.TempDir()
+	breaktest.CreateRepoAndCommit(t, tmpDir)
```
r-hang, line 35: Looking at how CreateRepoAndCommit is used in this test, I think that CreateRepoAndCommit hides too much of the details of what files get changed. In this package's test, we need to look into the internals of CreateReportAndCommit, which are in a different package,  to know what values to assert.

It seems like CreateRepoAndCommit should take as inputs file modifications to commit. With this, all the knowledge of what get committed and asserted can live in the same test and package.

internal/git/git.go
shirchen: i agree, having a bool as a field is always a code smell. fixed.

internal/compare/compare.go
```diff
@@ -0,0 +1,141 @@
...
+		p.service(fromService, to.Services[name])
+	}
+
+	for n, fromType := range from.Types {
+		p.typ(fromType, to.Types[n], filepath.Base(from.ThriftPath))
+	}
+}
+
+func (p *Pass) typ(from, to compile.TypeSpec, file string) {
+	switch f := from.(type) {
```
sywhang, line 77: is this a singular case switch stmt instead of a type cast for a reason? Can't we do:
```suggestion
	if f, ok := from.(*compile.StructSpec); ok {
	//...
	}
```

internal/compare/compare.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// Modules looks for removed methods and added required fields.
+func (p *Pass) Modules(from, to *compile.Module) {
+	for name, fromService := range from.Services {
+		p.service(fromService, to.Services[name])
+	}
+
+	for n, fromType := range from.Types {
+		p.typ(fromType, to.Types[n], filepath.Base(from.ThriftPath))
```
sywhang, line 72: I think `to.Types[n]` can be nil here so we'd wanna add a nil check at top of `func (p *Pass) typ` or here

sywhang (approved):
Minus couple of fixes suggested, I think this looks good as a start, given that we'll address some of the comments here in follow-up refactors and further rules to detect breaking changes.

internal/compare/compare.go
shirchen: Yea, I think goal is to support other types hence setup as a switch, but not yet. Happy to switch to a simple if statement if you want me not to future proof.

internal/compare/compare.go
```diff
@@ -0,0 +1,177 @@
...
+type deleteServiceError struct {
+	service string
+}
+
+func (e deleteServiceError) Error() string {
+	return fmt.Sprintf("deleting service %s is not backwards compatible", e.service)
+}
+
+// Modules looks for removed methods and added required fields.
+func (p *Pass) Modules(fromModule, toModule *compile.Module) {
```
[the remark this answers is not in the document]
shirchen: ugh, nvm, I forgot that I added `Pass` object hence I'm wrong and you won't call it like `compare.Modules`. it'll be `pass.Modules` vs `pass.CompareModules`.

internal/compare/compare.go
sywhang: I see - in that case can we just add a TODO comment here for that?

r-hang (approved):
The current CLI API looks good, the rest of the things under discussion are all internal. I was thinking of the HEAD to HEAD~ behaviour. If there was ever a need to compare against other points in the commit history, compatible, additional flags could be added.

internal/compare/compare.go
shirchen: ok, it'll be easier to switch to if now.

internal/compare/compare.go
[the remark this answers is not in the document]
sywhang: yeah i mean it's internal so I don't feel super strong about one way or the other. It's a little confusing cause both `pass` and `Modules` are nouns and it doesn't sound that intuitive but it's not a huge deal.

internal/compare/compare.go
shirchen: if there's no key, then value will be the map's zero-value, `TypeSpec`, not nil which will then get handled in `typ`. if you are looking at other cases where we handle it, it's because for those maps the values are pointers. 

from flow point of view, if we dont have the type, in this version, we can't handle renames yet, so we aren't going to error. renaming a struct is backwards compatible.

internal/breaktest/repo.go
```diff
@@ -0,0 +1,135 @@
...
+	t.Helper()
+	// Create a new repo in temp directory.
+	repository, err := git.PlainInit(tmpDir, false)
+	require.NoError(t, err)
+	worktree, err := repository.Worktree()
+	require.NoError(t, err)
+
+	// Start writing files.
+	exampleThrifts := map[string]string{
+		"v1.thrift": "namespace rb v1\n" +
```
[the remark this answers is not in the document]
abhinav: Yeah, test fixtures is a thing @shirchen and I talked about offline too.
There are some ideas in https://github.com/thriftrw/thriftrw-go/pull/538#discussion_r708456367,
if we were to use a txtar-based approach, but that's not the only approach.

For these end-to-end tests, we basically need some way to easily define:

- before state
- after state
- expected issues (if any)

There are a few ways to go about that,
and specifying before/after states and expected stdout in txtar files is one way.

internal/compare/compare.go
abhinav: So if `to.Types[n]` returns nil because the value doesn't exist in the `to`, the following won't `ok`:

    t, ok := to.(*compile.StructSpec)

It will refuse to turn a `nil` `TypeSpec` into a `*compile.StructSpec`, and return `false`.
So the nil check is kinda already done.

internal/compare/compare.go
```diff
@@ -0,0 +1,140 @@
...
+		p.structSpecs(f, t, file)
+	}
+}
+
+// StructSpecs compares two structs defined in a Thrift file.
+func (p *Pass) structSpecs(from, to *compile.StructSpec, file string) {
+	fields := make(map[int16]*compile.FieldSpec, len(from.Fields))
+	// Assume that these two should be compared.
+	for _, f := range from.Fields {
+		// Capture state of all fields here.
```
abhinav, line 92: nit: can delete comment since the next line is self explanatory

```suggestion
```

internal/compare/compare.go
```diff
@@ -0,0 +1,140 @@
...
+	for n, fromType := range from.Types {
+		p.typ(fromType, to.Types[n], filepath.Base(from.ThriftPath))
+	}
+}
+
+func (p *Pass) typ(from, to compile.TypeSpec, file string) {
+	if f, ok := from.(*compile.StructSpec); ok {
+		t, ok := to.(*compile.StructSpec)
+		if !ok {
+			// This is a new Type, which is backwards compatible.
```
abhinav, line 80: Comment is incorrect. If this is a new type, `from` would be nil.
Well, not actually, because it wouldn't be in `from.Types`. So this won't be called for new types.
We'll hit this line when a type is deleted,
which isn't breaking if it's unused or the corresponding usage was deleted too.

internal/compare/compare.go
```diff
@@ -0,0 +1,140 @@
...
+		fields[f.ID] = f
+	}
+
+	for _, toField := range to.Fields {
+		if fromField, ok := fields[toField.ID]; ok {
+			fromRequired := fromField.Required
+			toRequired := toField.Required
+			if !fromRequired && toRequired {
+				p.Report(Diagnostic{
+					File: file,
```
abhinav, line 102: I would have moved the compile.FieldGroup comparison to its own function so you could use it for the function arguments check.

internal/compare/compare.go
```diff
@@ -0,0 +1,140 @@
...
+					Message: fmt.Sprintf(
+						"changing an optional field %s in %s to required is not backwards compatible",
+						toField.ThriftName(), to.ThriftName()),
```
abhinav, line 105: I would use `%q` to quote the field names here and elsewhere.

```suggestion
						"changing an optional field %q in %q to required is not backwards compatible",
						toField.ThriftName(), to.ThriftName()),
```

internal/compare/compare.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (p *Pass) function(to *compile.FunctionSpec, fn string, file string, service string) {
+	if to == nil {
+		p.Report(Diagnostic{
+			File:    file,
+			Message: fmt.Sprintf("removing method %s in service %s is not backwards compatible", fn, service),
+		})
+	}
+}
```
abhinav, line 140: In case you missed https://github.com/thriftrw/thriftrw-go/pull/538#discussion_r714340252,
we'll want to check the function parameters with the same checks as struct fields.

Deferral to a follow up diff okay, but I want to make sure you saw this.

internal/compare/compare_test.go
```diff
@@ -0,0 +1,214 @@
...
+			toStruct: &compile.StructSpec{
+				Name: "structA",
+				Fields: compile.FieldGroup{
+					&compile.FieldSpec{
+						Required: true,
+						Name:     "fieldA",
+					},
+				},
+			},
+			wantError: "foo.thrift:adding a required field fieldA to structA is not backwards compatible\n",
```
abhinav, line 75: nit: given that the program is for detecting backwards incompatible changes, we could skip that from the messages and just tell people what they're doing wrong:

    foo.thrift:adding a required field fieldA to structA

internal/git/git.go
```diff
@@ -0,0 +1,185 @@
...
+	r, err := git.PlainOpenWithOptions(path, &git.PlainOpenOptions{
+		DetectDotGit:          true,
+		EnableDotGitCommonDir: true,
+	})
+	if err != nil {
+		return pass, err
+	}
+
+	h, err := findChangedThrift(r)
+	if err != nil || h == nil {
```
abhinav, line 65: Why do we need `h == nil` here?
Would findChangedThrift return a nil value with an `err == nil` under any valid usage?
If not, we should skip the check and assume `h` is valid.

If `err == nil` and `h == nil`, that's probably a bug and should fail in a loud way.
(Which it will, two lines below when `h.to` panics.)
```suggestion
	if err != nil {
```

abhinav (approved):
reasonable starting point; bunch of nits.
assuming that the following are deferred to follow-ups:

- end to end test suite based on external fixtures
- refactoring to allow separation of lint checks
- line (and possibly, column) numbers in error messages (will require plumbing them to `compile.*Spec` from `ast.Node`s)
- more checks, including function parameters being checked like struct parameters

sywhang (approved):
LGTM!

PR #541, opened 2021-08-31
streaming: Handle nil items in containers

witriew:
`nil` items should be handled gracefully similar to the Wire-based
representation.

Ref #183

### DIFF
--- a/gen/container_test.go
+++ b/gen/container_test.go
@@ -1280,7 +1280,7 @@ func TestContainerValidate(t *testing.T) {
 					{4, 5, 6},
 				},
 			},
-			wantError: "invalid '[][]byte', index [2]: value is nil",
+			wantError: "invalid list '[][]byte', index [2]: value is nil",
 		},
 		{
 			desc: "nil second element of array",
@@ -1291,7 +1291,7 @@ func TestContainerValidate(t *testing.T) {
 					{StartPoint: &ts.Point{X: 5, Y: 6}, EndPoint: &ts.Point{X: 7, Y: 8}},
 				},
 			},
-			wantError: "invalid '[]*Edge', index [1]: value is nil",
+			wantError: "invalid list '[]*Edge', index [1]: value is nil",
 		},
 		{
 			desc: "nil set item",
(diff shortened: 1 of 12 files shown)


gen/container_test.go
```diff
@@ -1319,12 +1321,12 @@ func TestContainerValidate(t *testing.T) {
...
 				},
 			},
-			wantError: "invalid [foo]: value is nil",
+			wantError: "invalid map 'map[string][]byte', key [foo]: value is nil",
 		},
 	}
 
 	for _, tt := range tests {
-		t.Run(tt.desc, func(t *testing.T) {
+		t.Run(tt.desc+" wire", func(t *testing.T) {
```
abhinav, line 1329: optional nit: can use `/wire` here and `/streaming` below to have the test system treat them as subtests.

```suggestion
		t.Run(tt.desc+"/wire", func(t *testing.T) {
```

gen/container_test.go
```diff
@@ -1334,6 +1336,21 @@ func TestContainerValidate(t *testing.T) {
 				assert.Equal(t, tt.wantError, err.Error())
 			}
 		})
+
+		t.Run(tt.desc+" streaming", func(t *testing.T) {
```
abhinav, line 1340:
```suggestion
		t.Run(tt.desc+"/streaming", func(t *testing.T) {
```

gen/list.go
```diff
@@ -60,7 +60,7 @@ func (l *listGenerator) ValueList(g Generator, spec *compile.ListSpec) (string,
 				<- else ->
 				for <$i>, <$x> := range <$v> {
 					if <$x> == nil {
-						return <import "fmt">.Errorf("invalid [%v]: value is nil", <$i>)
+						return <import "fmt">.Errorf("invalid '<typeReference .Spec>', index [%v]: value is nil", <$i>)
```
abhinav, line 63: nit: map and set error messages say "invalid map '..'" or "invalid set '...'" but this says just "invalid '...'".

I don't have a strong preference between "invalid list '...'" for this, or "invalid '...'" for all three, so your call on which error message you like more.

abhinav (approved):
Thanks! Minor error message nit, but LGTM otherwise.

Can you also add a `### Fixed` entry to the changelog?

gen/list.go
witriew: Made the error message more explicit with "invalid list '...'".

abhinav (approved):

PR #544, opened 2021-09-09
streaming: Skip over unrecognized fields and fields not matching the expected type

witriew:
In the Wire-based implementation of reading the raw Thrift protocol, the reader
greedily reads over all of the fields in the structs, populating a Wire
intermediary format.  This format is then validated and parsed by the generated
code, populating the appropriate Thrift object with the proper values
associated with the corresponding field ID.

In the stream-based implementation, validation and parsing of the raw Thrift
protocol is done as the bytes are read.  Since unrecognized fields were not
skipped over, this caused issues in future reads as there is now data that
doesn't match the expected Thrift type.

Skip over fields that are unrecognized by the streaming reader and skip fields
that don't match the type expected by the Thrift definition.

### DIFF
--- a/gen/field.go
+++ b/gen/field.go
@@ -601,10 +601,9 @@ func (f fieldGroupGenerator) Decode(g Generator) error {
 			}
 
 			for <$ok> {
-				switch <$fh>.ID {
+				switch {
 				<range .Fields ->
-				case <.ID>:
-					if <$fh>.Type == <typeCode .Type> {
+				case <$fh>.ID == <.ID> && <$fh>.Type == <typeCode .Type>:
 						<- $lhs := printf "%s.%s" $v (goName .) ->
 						<- if .Required ->
 							<$lhs>, err = <decode .Type $sr>
@@ -617,11 +616,6 @@ func (f fieldGroupGenerator) Decode(g Generator) error {
 						<if .Required ->
 							<$isSet.Rotate (printf "%sIsSet" .Name)> = true
 						<- end>
-					} else {
-						if err := <$sr>.Skip(<$fh>.Type); err != nil {
-							return err
-						}
-					}
 				<end ->
 				default:
 					if err := <$sr>.Skip(<$fh>.Type); err != nil {
--- a/gen/field.go
+++ b/gen/field.go
@@ -617,8 +617,16 @@ func (f fieldGroupGenerator) Decode(g Generator) error {
 						<if .Required ->
 							<$isSet.Rotate (printf "%sIsSet" .Name)> = true
 						<- end>
+					} else {
+						if err := <$sr>.Skip(<$fh>.Type); err != nil {
+							return err
+						}
 					}
 				<end ->
+				default:
+					if err := <$sr>.Skip(<$fh>.Type); err != nil {
+						return err
+					}
 				}
 
 				if err := <$sr>.ReadFieldEnd(); err != nil {
--- a/gen/struct_test.go
+++ b/gen/struct_test.go
@@ -762,14 +762,16 @@ func TestStructFromWireUnrecognizedField(t *testing.T) {
 				buf bytes.Buffer
 			)
 
-			protocol.Binary.Encode(tt.give, &buf)
+			require.NoError(t, binary.Default.Encode(tt.give, &buf))
 
-			sr := protocol.BinaryStreamer.Reader(bytes.NewReader(buf.Bytes()))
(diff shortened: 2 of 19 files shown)


gen/struct_test.go
```diff
@@ -729,20 +731,55 @@ func TestStructFromWireUnrecognizedField(t *testing.T) {
...
+			} else {
+				if assert.NoError(t, err, tt.desc) {
+					assert.Equal(t, tt.want, o)
+				}
 			}
-		}
+
+			_, err = sr.ReadBool()
+			require.Error(t, err)
+			assert.Contains(t, err.Error(), "unexpected EOF", "exepected error for continued reads")
```
abhinav, line 781: This seems a bit janky in that it's a random `ReadBool`. I think it might be cleaner to retain a reference to the `bytes.NewReader` and [check that it has no unread bytes](https://pkg.go.dev/bytes#Reader.Len).

gen/struct_test.go
```diff
@@ -793,17 +830,44 @@ func TestUnionFromWireInconsistencies(t *testing.T) {
...
+			} else {
+				if assert.Error(t, err, tt.desc) {
+					assert.Contains(t, err.Error(), tt.failure, tt.desc)
+				}
 			}
-		}
+
+			_, err = sr.ReadBool()
+			require.Error(t, err)
+			assert.Contains(t, err.Error(), "unexpected EOF", "exepected error for continued reads")
```
abhinav, line 869: same comments as above

gen/struct_test.go
```diff
@@ -793,17 +830,44 @@ func TestUnionFromWireInconsistencies(t *testing.T) {
...
-				assert.Contains(t, err.Error(), tt.failure, tt.desc)
+		})
+
+		t.Run(tt.desc+"/streaming", func(t *testing.T) {
+			var (
+				o   tu.Document
+				buf bytes.Buffer
+			)
+
+			protocol.Binary.Encode(tt.input, &buf)
```
abhinav, line 853:
```suggestion
			require.NoError(t, binary.Default.Encode(tt.input, &buf))
```

gen/struct_test.go
```diff
@@ -793,17 +830,44 @@ func TestUnionFromWireInconsistencies(t *testing.T) {
...
+
+			sr := protocol.BinaryStreamer.Reader(bytes.NewReader(buf.Bytes()))
```
abhinav, line 855:
```suggestion
			sr := binary.Default.Reader(bytes.NewReader(buf.Bytes()))
			defer sr.Close()
```

gen/struct_test.go
```diff
@@ -793,17 +830,44 @@ func TestUnionFromWireInconsistencies(t *testing.T) {
...
+			err := o.Decode(sr)
+			if tt.success != nil {
+				if assert.NoError(t, err, tt.desc) {
+					assert.Equal(t, tt.success, &o, tt.desc)
+				}
```
abhinav, line 860:
```suggestion
				}
				return
```

gen/struct_test.go
```diff
@@ -729,20 +731,55 @@ func TestStructFromWireUnrecognizedField(t *testing.T) {
...
+			)
+
+			protocol.Binary.Encode(tt.give, &buf)
+
+			sr := protocol.BinaryStreamer.Reader(bytes.NewReader(buf.Bytes()))
+			err := o.Decode(sr)
+			if tt.wantError != "" {
+				if assert.Error(t, err, tt.desc) {
+					assert.Contains(t, err.Error(), tt.wantError)
+				}
```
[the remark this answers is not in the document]
witriew: Undid these returns, the only instance where the reads will return early is if there's an IO error, otherwise the whole struct should be consumed.

gen/field.go
```diff
@@ -617,8 +617,16 @@ func (f fieldGroupGenerator) Decode(g Generator) error {
 						<if .Required ->
 							<$isSet.Rotate (printf "%sIsSet" .Name)> = true
 						<- end>
+					} else {
+						if err := <$sr>.Skip(<$fh>.Type); err != nil {
+							return err
+						}
```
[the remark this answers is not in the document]
witriew: Yep, cleaned it up.

gen/struct_test.go
witriew: Done.

abhinav (approved):

PR #551, opened 2021-11-09
Add a flag to thriftbreak to output results as JSON.

shirchen:
Adding `--json` flag to thriftbreak to output results as JSON. This enables integration with tools like Phabricator.

Ref GO-938

### DIFF
--- a/cmd/thriftbreak/main.go
+++ b/cmd/thriftbreak/main.go
@@ -21,6 +21,7 @@
 package main
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
 	"log"
@@ -38,6 +39,7 @@ func main() {
 func run(args []string) error {
 	flag := flag.NewFlagSet("thriftbreak", flag.ContinueOnError)
 	gitRepo := flag.String("C", "", "location of git repository. Defaults to current directory.")
+	jsonOut := flag.Bool("json", false, "output as JSON")
 	if err := flag.Parse(args); err != nil {
 		return err
 	}
@@ -57,8 +59,18 @@ func run(args []string) error {
 	}
 
 	lints := pass.Lints()
+	var out string
 	for _, l := range lints {
-		fmt.Println(l.String())
+		if *jsonOut {
+			bytes, err := json.Marshal(l)
+			if err != nil {
+				return fmt.Errorf("failed to marshal error: %v", err)
+			}
+			out = string(bytes)
+		} else {
+			out = l.String()
+		}
+		fmt.Println(out)
 	}
 	if len(lints) > 0 {
 		return fmt.Errorf("found %d issues", len(lints))
--- a/cmd/thriftbreak/main.go
+++ b/cmd/thriftbreak/main.go
@@ -22,24 +22,53 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
+	"io"
 	"log"
 	"os"
 
+	"go.uber.org/thriftrw/internal/compare"
 	"go.uber.org/thriftrw/internal/git"
 )
 
 func main() {
-	if err := run(os.Args[1:]); err != nil && err != flag.ErrHelp {
+	if err := run(os.Args[1:]); err != nil && !errors.Is(err, flag.ErrHelp) {
 		log.Fatalf("%+v", err)
+	}
+}
+
+func readableOutput(w io.Writer) func(compare.Diagnostic) error {
+	return func(diagnostic compare.Diagnostic) error {
+		// fmt.Println(diagnostic.String())
+		if _, err := fmt.Fprintln(w, diagnostic.String()); err != nil {
+			return fmt.Errorf("failed to output a lint error: %v", err)
+		}
+
+		return nil
+	}
+}
+
+func jsonOutput(w io.Writer) func(compare.Diagnostic) error {
+	enc := json.NewEncoder(w)
+
+	return func(diagnostic compare.Diagnostic) error {
+		// Encode adds a trailing newline.
+		if err := enc.Encode(diagnostic); err != nil {
+			return fmt.Errorf("failed to encode: %v", err)
+		}
+
+		return nil
 	}
 }
 
 func run(args []string) error {
 	flag := flag.NewFlagSet("thriftbreak", flag.ContinueOnError)
-	gitRepo := flag.String("C", "", "location of git repository. Defaults to current directory.")
-	jsonOut := flag.Bool("json", false, "output as JSON")
+	gitRepo := flag.String("C", "",
+		"location of git repository. Defaults to current directory.")
+	jsonOut := flag.Bool("json", false,
+		"output as a list of newline-delimited JSON objects with the following fields: File and Message")
 	if err := flag.Parse(args); err != nil {
 		return err
 	}
@@ -58,20 +87,19 @@ func run(args []string) error {
 		return err
 	}
 
-	lints := pass.Lints()
-	var out string
-	for _, l := range lints {
-		if *jsonOut {
-			bytes, err := json.Marshal(l)
-			if err != nil {
-				return fmt.Errorf("failed to marshal error: %v", err)
-			}
-			out = string(bytes)
-		} else {
-			out = l.String()
-		}
-		fmt.Println(out)
+	var write func(compare.Diagnostic) error
+	if *jsonOut {
+		write = jsonOutput(os.Stdout)
+	} else {
+		write = readableOutput(os.Stdout)
 	}
+	lints := pass.Lints()
+	for _, l := range lints {
+		if err := write(l); err != nil {
+			return fmt.Errorf("failed to output error: %v", err)
+		}
+	}
+
 	if len(lints) > 0 {
 		return fmt.Errorf("found %d issues", len(lints))
 	}
--- a/cmd/thriftbreak/main.go
+++ b/cmd/thriftbreak/main.go
@@ -39,9 +39,9 @@ func main() {
 	}
 }
 
+// readableOutput prints every lint error on a separate line.
 func readableOutput(w io.Writer) func(compare.Diagnostic) error {
 	return func(diagnostic compare.Diagnostic) error {
-		// fmt.Println(diagnostic.String())
 		if _, err := fmt.Fprintln(w, diagnostic.String()); err != nil {
 			return fmt.Errorf("failed to output a lint error: %v", err)
 		}
@@ -50,6 +50,7 @@ func readableOutput(w io.Writer) func(compare.Diagnostic) error {
 	}
 }
 
+// jsonOutput prints out every lint error in JSON format.
 func jsonOutput(w io.Writer) func(compare.Diagnostic) error {
 	enc := json.NewEncoder(w)
 
--- a/cmd/thriftbreak/main_test.go
+++ b/cmd/thriftbreak/main_test.go
@@ -32,58 +32,86 @@ import (
 )
 
 func TestThriftBreakIntegration(t *testing.T) {
-	tmpDir := t.TempDir()
-	from := map[string]string{
-		"v1.thrift": "namespace rb v1\n" +
-			"struct AddedRequiredField {\n" +
-			"    1: optional string A\n" +
-			"    2: optional string B\n" +
-			"}\n" +
-			"\nservice Foo {\n    void methodA()\n}",
-		"test/v2.thrift": `service Bar {}`,
-		"test/c.thrift":  `service Baz {}`,
-		"test/d.thrift": `include "../v1.thrift"
-		service Qux {}`, // d.thrift will be deleted below.
-		"somefile.go": `service Quux{}`, // a .go file, not a .thrift.
+	type test struct {
+		desc      string
+		want      string
+		gitDirArg string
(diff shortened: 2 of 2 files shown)


cmd/thriftbreak/main.go
```diff
@@ -57,8 +59,18 @@ func run(args []string) error {
 	}
 
 	lints := pass.Lints()
+	var out string
```
abhinav, line 62: This variable declared outside the loop seems unnecessary.
It's only needed inside the loop.

cmd/thriftbreak/main.go
```diff
@@ -57,8 +59,18 @@ func run(args []string) error {
...
 	for _, l := range lints {
-		fmt.Println(l.String())
+		if *jsonOut {
+			bytes, err := json.Marshal(l)
+			if err != nil {
+				return fmt.Errorf("failed to marshal error: %v", err)
+			}
+			out = string(bytes)
+		} else {
```
abhinav, line 70: This if-else over here is a bit smelly.

Consider resolving the way output is processed once outside the loop (possibly into a `func(Lint) error`)
and calling that from within the loop.

So maybe something like the following would work:

```go
func readableOutput(w io.Writer) func(Lint) error
func jsonOutput(w io.Writer) func(Lint) error
```

Usage:

```go
var write func(Lint) error
if *json { write = jsonOutput(os.Stdout) }
else { write = readableOutput(os.Stdout) }

for _, l := range lints {
  if err := write(l); err != nil { .. }
}
```

---

For the JSON output, if you use [json.Encoder](https://pkg.go.dev/encoding/json#Encoder), the newline will be included for you:

```
  enc := json.NewEncoder(w)
  return func(l Lint) error {
    // Encode adds a trailing newline.
    if err := enc.Encode(l); err != nil {
        return fmt.Errorf(..)
    }
    return nil
  }
```

---

Lastly, I'm using `func(Lint) error` as the output consumer,
but I'll leave it up to you if that's okay,
or if you'd prefer something like a `Sink interface{ WriteLint(Lint) error }` type.

cmd/thriftbreak/main.go
```diff
@@ -38,6 +39,7 @@ func main() {
 func run(args []string) error {
 	flag := flag.NewFlagSet("thriftbreak", flag.ContinueOnError)
 	gitRepo := flag.String("C", "", "location of git repository. Defaults to current directory.")
+	jsonOut := flag.Bool("json", false, "output as JSON")
```
abhinav, line 42: minor: If we're adding a JSON flag, we should maybe mention what fields are included in the JSON object.

    "output as a list of newline-delimited JSON objects with the following fields: File and Message"

(reformat and reword for readability to your preference.)

cmd/thriftbreak/main_test.go
```diff
@@ -32,58 +32,86 @@ import (
...
-			"struct AddedRequiredField {\n" +
-			"    1: optional string A\n" +
-			"    2: optional string B\n" +
-			"    3: required string C\n}\n" +
-			"service Foo {}",
-		"test/v2.thrift": `service Foo {}`,
-		"test/c.thrift": `include "../v1.thrift"
-		service Bar {}`,
-		"somefile.go": `service Qux{}`,
+	tests := []test{
```
abhinav, line 41: Naming this type is unnecessary. For table tests, we typically don't name the test case struct unless we need to declare a separate `func` that accepts it.

```
tests := []struct {
  // fields
}{
  // test cases
}
```

cmd/thriftbreak/main_test.go
```diff
@@ -32,58 +32,86 @@ import (
...
+					"    1: optional string A\n" +
+					"    2: optional string B\n" +
+					"    3: required string C\n}\n" +
+					"service Foo {}",
+				"test/v2.thrift": `service Foo {}`,
+				"test/c.thrift": `include "../v1.thrift"
+		service Bar {}`,
+				"somefile.go": `service Qux{}`,
+			}
+			remove := []string{"test/d.thrift"}
```
abhinav, line 91: Seeing as these bits are constant, consider putting this outside the loop, above the test cases.

```
from := ...
to := ...
remove := ...

tests := []struct {
 ...
}{
  ..
}

for _, tt := range tests {
  t.Run(..., func(t *testing.T) {
    // ...
    breaktest.CreateRepoAndCommit(..)
  })
}
```

cmd/thriftbreak/main_test.go
```diff
@@ -32,58 +32,86 @@ import (
...
-		os.Stdout = oldStdout
-	}(os.Stdout)
-	os.Stdout = f
+			f, err := ioutil.TempFile(tmpDir, "stdout")
+			require.NoError(t, err, "create temporary file")
+			defer func(oldStdout *os.File) {
+				assert.NoError(t, f.Close())
+				os.Stdout = oldStdout
+			}(os.Stdout)
+			os.Stdout = f
```
abhinav, line 100: I don't remember if we previously discussed this.
stdout hijacking is undesirable.
It is preferable for `run` to parameterize stdout as an io.Writer parameter and write to that.

That will also significantly simplify this test case by obviating the need for the TempFile. We'll be able to pass a bytes.Buffer as the stdout io.Writer and verify output.

cmd/thriftbreak/main.go
```diff
@@ -57,8 +59,18 @@ func run(args []string) error {
 	}
 
 	lints := pass.Lints()
+	var out string
 	for _, l := range lints {
-		fmt.Println(l.String())
+		if *jsonOut {
```
sywhang, line 64: If we do this, aren't we appending many JSONs, separated by a newline? Sounds like that wouldn't be a valid JSON format if you piped the output - is that what we want or should we output one valid JSON-formatted string instead?

cmd/thriftbreak/main_test.go
shirchen: Sounds good. Yup, noted in GO-892

cmd/thriftbreak/main.go
abhinav: I think the typical behavior for something like this is newline delimited JSON objects. These can be fed directly into jq.

abhinav: (A single JSON string can also be fed into jq, but there's explicit support for ndjson in jq.)

sywhang: i see. nvm then :)

cmd/thriftbreak/main.go
```diff
@@ -21,23 +21,54 @@
...
+		return nil
+	}
+}
+
+func jsonOutput(w io.Writer) func(compare.Diagnostic) error {
+	enc := json.NewEncoder(w)
+
+	return func(diagnostic compare.Diagnostic) error {
+		// Encode adds a trailing newline.
+		if err := enc.Encode(diagnostic); err != nil {
```
sywhang, line 58: Do we care about failure cases being tested here?

cmd/thriftbreak/main_test.go
```diff
@@ -60,30 +85,33 @@ func TestThriftBreakIntegration(t *testing.T) {
...
-	}(os.Stdout)
-	os.Stdout = f
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			tmpDir := t.TempDir()
+			breaktest.CreateRepoAndCommit(t, tmpDir, from, to, remove)
+
+			f, err := ioutil.TempFile(tmpDir, "stdout")
+			require.NoError(t, err, "create temporary file")
+			defer func(oldStdout *os.File) {
```
sywhang, line 96: stdout hijacking is probably unnecessary now that you have jsonOutput and readableOutput but I assume you'll be doing that in a follow-up PR?

cmd/thriftbreak/main.go
shirchen: yeah, im not sure how to make serialization of a struct that we own fail. fields are strings too so i couldnt figure out how to make a test for catching an error.

cmd/thriftbreak/main.go
```diff
@@ -21,23 +21,54 @@
...
-	if err := run(os.Args[1:]); err != nil && err != flag.ErrHelp {
+	if err := run(os.Args[1:]); err != nil && !errors.Is(err, flag.ErrHelp) {
 		log.Fatalf("%+v", err)
 	}
 }
 
+func readableOutput(w io.Writer) func(compare.Diagnostic) error {
+	return func(diagnostic compare.Diagnostic) error {
+		// fmt.Println(diagnostic.String())
+		if _, err := fmt.Fprintln(w, diagnostic.String()); err != nil {
```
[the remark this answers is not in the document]
shirchen: I dont think it does, so leaving that here.

sywhang (approved):
:shipit:

cmd/thriftbreak/main_test.go
abhinav: I think the hijacking is still needed if while the code is still calling `run()` because run does not yet parameterize stdout. @shirchen plans to fix that in GO-892.

The unit tests for json/readableOutput don't need stdout hijacking, though, that's true.

cmd/thriftbreak/main_test.go
```diff
@@ -60,30 +88,67 @@ func TestThriftBreakIntegration(t *testing.T) {
...
+			}(os.Stdout)
+			os.Stdout = f
 
-	err = run([]string{fmt.Sprintf("-C=%s", tmpDir)})
-	require.Error(t, err, "expected no errors")
-	assert.EqualError(t, err, "found 5 issues")
+			// Pinning not to run into scoping errors.
+			gitDir := tt.gitDirArg
+			extraCmd := tt.extraCmd
+			err = run([]string{fmt.Sprintf(gitDir, tmpDir), extraCmd})
```
abhinav, line 108: So two things:

- the contents of gitDirArg don't vary
- the flag doesn't need to use the "-C=foo" form to be picked up

The suggestion from the first one would have been to just do:

```
run([]string{"-C="+tmpDir, extraCmd})
```

But also because you don't need to use `=`, tmpDir can just be the next arg, you can also do the following, which eliminates the string append too:

```suggestion
			err = run([]string{"-C", tmpDir, extraCmd})
```

cmd/thriftbreak/main_test.go
```diff
@@ -60,30 +88,67 @@ func TestThriftBreakIntegration(t *testing.T) {
...
+				os.Stdout = oldStdout
+			}(os.Stdout)
+			os.Stdout = f
 
-	err = run([]string{fmt.Sprintf("-C=%s", tmpDir)})
-	require.Error(t, err, "expected no errors")
-	assert.EqualError(t, err, "found 5 issues")
+			// Pinning not to run into scoping errors.
+			gitDir := tt.gitDirArg
+			extraCmd := tt.extraCmd
```
abhinav, line 107: It's unclear why these variables are being declared here.
What scoping error are you talking about?

cmd/thriftbreak/main_test.go
```diff
@@ -60,30 +88,67 @@ func TestThriftBreakIntegration(t *testing.T) {
...
+			err = run([]string{fmt.Sprintf(gitDir, tmpDir), extraCmd})
 
-	stderr, err := ioutil.ReadFile(f.Name())
-	require.NoError(t, err)
+			require.Error(t, err, "expected no errors")
```
abhinav, line 110: The message is incorrect. Did you mean "expected an error"?

cmd/thriftbreak/main_test.go
```diff
@@ -60,30 +88,67 @@ func TestThriftBreakIntegration(t *testing.T) {
...
-			`v1.thrift:removing method "methodA" in service "Foo"`+"\n"+
-			`v1.thrift:adding a required field "C" to "AddedRequiredField"`+"\n",
-		out)
+			out := string(stderr)
+			assert.Equal(t, tt.want, out)
+		})
+	}
+}
+
+func TestJsonPrinter(t *testing.T) {
```
abhinav, line 122: In Go naming, initialisms are usually all caps (unless they start an unexported variable).
```suggestion
func TestJSONPrinter(t *testing.T) {
```

ref: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

---

Although since this tests both printers, not just JSON, it should probably be named something else.

```suggestion
func TestDiagnosticPrinters(t *testing.T) {
```

cmd/thriftbreak/main_test.go
```diff
@@ -60,30 +88,67 @@ func TestThriftBreakIntegration(t *testing.T) {
...
+	t.Parallel()
+	tests := []struct {
+		desc   string
+		want   string
+		writer func(io.Writer) func(compare.Diagnostic) error
+	}{
+		{
+			desc:   "json writer",
+			want:   "{\"File\":\"foo.thrift\",\"Message\":\"error\"}\n",
```
abhinav, line 131: Can use backticks here to avoid escaping most of the JSON.

```suggestion
			want:   `{"File":"foo.thrift","Message":"error"}` + "\n",
```

abhinav (commented):
Minor comments; should be good to go after them.

cmd/thriftbreak/main_test.go
shirchen: https://github.com/kyoh86/scopelint but i'll remove them.

PR #553, opened 2021-12-03
Addressing missed feedback from #551 and returning FilePath

shirchen:
Addressing leftover comments and changing thriftbreak to return filepath relative to the passed in repository.

### DIFF
--- a/internal/compare/compare.go
+++ b/internal/compare/compare.go
@@ -139,7 +139,7 @@ func (p *Pass) service(from, to *compile.ServiceSpec) {
 // fallbacks to file name for cases when it was deleted.
 func (p *Pass) getRelativePath(filePath string) string {
 	if file, err := filepath.Rel(p.GitDir, filePath); err == nil {
-	    return file
+		return file
 	}
 	// If a file was deleted, then we will not be able to
 	// find a relative path to it.
--- a/internal/compare/compare.go
+++ b/internal/compare/compare.go
@@ -30,17 +30,18 @@ import (
 
 // Diagnostic is a message associated with an error and a file name.
 type Diagnostic struct {
-	File    string // File where error was discovered.
-	Message string // Message contains error message.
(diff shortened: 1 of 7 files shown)


internal/compare/compare.go
```diff
@@ -121,22 +123,37 @@ func (p *Pass) service(from, to *compile.ServiceSpec) {
...
 }
 
-func (p *Pass) function(to *compile.FunctionSpec, fn string, file string, service string) {
+// getRelativePath returns a relative path to a file or
+// fallbacks to file name for cases when it was deleted.
+func (p *Pass) getRelativePath(filePath string) string {
+	file, err := filepath.Rel(p.GitDir, filePath)
+	if err != nil {
+		// If a file was deleted, then we will not be able to
+		// find a full path to it.
```
sywhang, line 144: did you mean to say we will not be able to find a relative path to it?

sywhang (commented):
lgtm modulo a few minor fixes

internal/compare/compare_test.go
```diff
@@ -215,3 +215,33 @@ func TestServicesError(t *testing.T) {
...
+			want: "buz/buz.go",
+		},
+		{
+			desc: "fallback case where we just return file name",
+			path: "foo/buz.go",
+			want: "buz.go",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
```
abhinav, line 238: Need `tt := tt` here or we'll run into the closure scoping issue.
```suggestion
	for _, tt := range tests {
		tt := tt
		t.Run(tt.desc, func(t *testing.T) {
```

abhinav (approved):
LGTM minus minor sung's comments and the missing `tt := tt` in a test that uses `t.Parallel()`.

internal/compare/compare.go
shirchen: yup yup. thanks

PR #554, opened 2022-02-18
optimize String() on a typedef if the underlying type is a string

kishoresenji:
If the underlying type for a defined type is a string, then the Stringer interface implementation can be optimized by returning the string after type conversion. 

```
type UUID string
func (v UUID) String() string {
        x := (string)(v)
        return fmt.Sprint(x)
 }
```
to 

```
func (v UUID) String() string {
        x := (string)(v)
        return x
 }
 ```


abhinav (approved):
Thanks!

PR #555, opened 2022-03-01
Update copyright range to include 2022

jparise:
This is a proactive change to remove noise from future work that needs
`make generate`'s output.


sywhang (approved):
LGTM. Thank you for doing this.

PR #556, opened 2022-03-02
idl: Capture parsed comments

jparise:
Line-wise comment strings are available through the `Comment()` function
on the `idl.Config` struct.

---

### Open questions

1. Should we strip the comment prefix (`#` and `//`) or store the entire comment string unaltered?
2. Is a line-wise accessor the right interface?
3. How should we handle multiline comments (`/* ... */`)?
  a. If we strip prefixes in (1), then should be attempt to strip any leading `*` characters on intermediate lines?
  b. Should a line-wise accessor return the full multiline comment string for all lines in its range?

---

I also played with a fancier, AST-like system based on a rich struct:

```go
type CommentStyle int

const (
	UnixStyle      CommentStyle = iota + 1
    CPPStyle
    MultiLineStyle
)

type Comment struct {
    Style CommentStyle
    Text  string
    Start ast.Position
    End   ast.Position
}
```

... but I started to feel like that was overdoing it.

### DIFF
--- a/idl/config.go
+++ b/idl/config.go
@@ -37,6 +37,7 @@ func (c *Config) Parse(s []byte) (*ast.Program, error) {
 	result, errors := internal.Parse(s)
 	if c.Info != nil {
 		c.Info.nodePositions = result.NodePositions
+		c.Info.comments = result.Comments
 	}
 	return result.Program, newParseError(errors)
 }
--- a/idl/config_test.go
+++ b/idl/config_test.go
@@ -26,6 +26,7 @@ import (
 	"go.uber.org/thriftrw/ast"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestParse(t *testing.T) {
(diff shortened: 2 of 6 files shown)


abhinav:
Hello again! Thanks for the PR.
RE: your questions:

> 1\. Should we strip the comment prefix (`#` and `//`) or store the entire comment string unaltered?

> 3\. How should we handle multiline comments (`/* ... */`)?

As you pointed out, these two are related. I'm of mixed feelings on this. I like the idea of stripping the prefixes on line comments, but if we do that, then we need to also sanitize block comments. And that brings with it a whole slew of variations that we'd need to handle (leading `*` on intermediate lines or not, `/**` or `/*`, indented, partially indented, etc.).

Based on that, I'm leaning towards leaving comments unchanged—except maybe for block comments, we strip indentation up to the column where "/*" starts. Then we can let callers decide how to interpret the comments.

> 2\. Is a line-wise accessor the right interface?

I think maybe not—at least not long term.
To find all comments between two entities, a user will end up doing this:

```go
var comments []string
for i := start.Line; i < end.Line; i++ {
  if c, ok := info.Comment(i); ok {
    comments = append(comments, c)
  }
}
```

I haven't explored the complexity of this, but one idea is to track comments around AST nodes similar to [bazelbuild/buildtools/bulid](https://pkg.go.dev/github.com/bazelbuild/buildtools/build#Comments). For each AST node, it records a Comments object which holds:

- whole line comments that appear above the node
- suffix comments that appear on the line where the node ends (`struct Foo {} // bar`)
- whole line comments that appear after the node not attached to another node (I'm uncertain under which conditions this is used)

If we were to go this route, we could simplify a bit because we probably don't need the ability to reproduce the original Thrift file as-is. We could also omit suffix comments until needed.

That said, I'll also understand if that's a lot more complexity than you were planning for. In that case, we can also go for a simpler approach of just exposing a full list of all comments and the lines/columns they appear on (`LineComment{Line: 42, Value: ...}`), sorted by line number. Callers can do more everything with that that they can with a `map[int]string`.

jparise:
I haven't abandoned this, but it may be a little bit more time before I can share a revision that incorporates the ideas above.

jparise:
I'm afraid I'm not going to be able to pick this back up in a reasonable time frame after all, so I'm going to close this.

I'd be glad if someone with an interest were to reopen and take it over, though. 😉

PR #559, opened 2022-10-06
Add support for changed types in thriftbreak

moisesvega:
Changing field types is a breaking change, this could be prevented by raising an error in thriftbreak.

Example:
```
// some.thrift
struct MyStruct {
    1: optional bool status
}

changed to

// some.thrift
struct MyStruct {
    1: optional string status
}
```
is a breaking change because status went from being a bool to a string.

### DIFF
--- a/internal/compare/compare.go
+++ b/internal/compare/compare.go
@@ -99,6 +99,21 @@ func (p *Pass) requiredField(fromField, toField *compile.FieldSpec, to *compile.
 	}
 }
 
+func (p *Pass) changedTypes(fromField, toField *compile.FieldSpec, to *compile.StructSpec, file string) {
+	if fromField.Type == nil || toField.Type == nil {
+		return
+	}
+
+	if fromField.Type.ThriftName() != toField.Type.ThriftName() {
+		p.Report(Diagnostic{
+			FilePath: file,
+			Message: fmt.Sprintf(
+				"changing type field %q in %q to %q",
+				toField.Type.ThriftName(), to.ThriftName(), fromField.Type.ThriftName()),
+		})
+	}
+}
+
 // StructSpecs compares two structs defined in a Thrift file.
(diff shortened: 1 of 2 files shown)


manjari25:
> Changing filed types is a breaking change, this could be prevented by raising an error in thriftbreak.
> 
> Example:
> 
> ```
> // original.thrift
> struct MyStruct {
>     1: optional bool status
> }
> 
> // updated.thrift
> struct MyStruct {
>     1: optional string status
>     2: optional bool retry
> }
> ```
> 
> This diff add support for changed types in thriftbreak

Suggest including the following example in the description -  
Example:
 
```
// some.thrift
struct MyStruct {
    1: optional bool status
}
```

changed to 

```
// some.thrift
struct MyStruct {
    1: optional string status
}
```

is a breaking change because `status` went from being a `bool` to a `string`.

internal/compare/compare.go
```diff
@@ -99,6 +99,21 @@ func (p *Pass) requiredField(fromField, toField *compile.FieldSpec, to *compile.
...
+	if fromField.Type == nil || toField.Type == nil {
+		return
+	}
+
+	if fromField.Type.ThriftName() != toField.Type.ThriftName() {
+		p.Report(Diagnostic{
+			FilePath: file,
+			Message: fmt.Sprintf(
+				"changing type field %q in %q to %q",
+				toField.Type.ThriftName(), to.ThriftName(), fromField.Type.ThriftName()),
```
manjari25, line 112:
```suggestion
				"changing type of field %q in struct %q from %q to %q"
				toField.Name(), to.ThriftName(), fromField.Type.ThriftName(), toField.Type.ThriftName()),
```

manjari25 (approved):

r-hang (approved):

sywhang (approved):
LGTM

PR #560, opened 2023-02-04
idl: Fix parsing of hex integer values

KarboniteKream:
Parsing of integer values is done with `strconv.ParseInt()`, which currently fails when a value like `0x1234` is specified. This is because the method ignores the prefix when `base` is explicitly specified ([reference](https://pkg.go.dev/strconv#ParseInt)).

This PR adjusts the parsing code to skip the first 2 characters of the value, if it starts with `0x`.


idl/internal/lex.rl
```diff
@@ -283,6 +283,7 @@ func (lex *lexer) Lex(out *yySymType) int {
                 base := 10
                 if len(str) > 2 && str[0:2] == "0x" {
                     // Hex constant
+                    str = str[2:]
```
abhinav, line 286: Would you mind adding a comment here indicating why we're doing this? Because this is a bit subtle. Perhaps something like:

```
// Hex constant.
// strconv.ParseInt doesn't want the "0x" prefix if we explicitly set base to 16.
```

gen/enum_test.go
```diff
@@ -96,6 +96,37 @@ func TestEqualsEnumWithValues(t *testing.T) {
...
 }
 
+func TestValueOfEnumWithHexValues(t *testing.T) {
+	tests := []struct {
+		e te.EnumWithHexValues
+		i int32
+	}{
+		{te.EnumWithHexValuesX, 291},
+		{te.EnumWithHexValuesY, 1110},
+		{te.EnumWithHexValuesZ, 1929},
```
abhinav, line 106: Could we just use hex forms here? It'll make it obvious what this is matching against in the Thrift file.
```suggestion
		{te.EnumWithHexValuesX, 0x123},
		{te.EnumWithHexValuesY, 0x456},
		{te.EnumWithHexValuesZ, 0x789},
```

gen/constant_test.go
```diff
@@ -45,6 +45,36 @@ func TestConstants(t *testing.T) {
...
+		{
+			"hex32",
+			tk.Hex32,
+			int32(305419896),
+		},
+		{
+			"hex64",
+			tk.Hex64,
+			int64(1311768467294899695),
+		},
```
abhinav, line 77: Same: if we use hex forms here, this becomes slightly more obvious:

```suggestion
			tk.Hex16,
			int16(0x1234),
		},
		{
			"hex32",
			tk.Hex32,
			int32(0x12345678),
		},
		{
		
		
			"hex64",
			tk.Hex64,
			int64(0x1234567890abcdef),
		},
```

abhinav (approved):
Good catch! Thanks for the fix.

CC @sywhang @albertyw

KarboniteKream:
Updated, thank you.

abhinav (approved):

albertyw (approved):

PR #566, opened 2023-04-04
release preparation: make generate

r-hang:
The thriftrw release instructions call for running `make generate` and seeing no changes to verify the state of dev before creating a release commit PR on master. This change is the result of running `make generate`.


idl/internal/y.go
```diff
@@ -167,7 +167,7 @@ const yyEofCode = 1
 const yyErrCode = 2
 const yyInitialStackSize = 16
 
-var yyExca = [...]int{
+var yyExca = [...]int8{
```
r-hang, line 170: Flagging for reviewers that this file has non-comment changes.

sywhang (approved):

PR #568, opened 2023-04-05
string typedef: Fix fmt import and return type bugs

r-hang:
PR (#554) introduced an optimization for the String() method of typedefs.

There are two issues with this change:
- Since the "fmt" import was global, if fmt.Sprint was dropped from the typedef code generation the import would be unused.
- Using compiled.RootTypeSpec for isStringType results in situations where a non-string type is return from a String() function resulting in uncompilable code. ``` func(v StringReDef) String() string { x := (stringdef.StringDef)(v) return x } ``` This happens when a type is a typedef of another string typedef.

I've added generated code cases to demonstrate and verify the corrected behavior.


gen/type.go
```diff
@@ -76,7 +76,6 @@ func isPrimitiveType(spec compile.TypeSpec) bool {
 
 // isStringType returns true if the given type is a string type
 func isStringType(spec compile.TypeSpec) bool {
-	spec = compile.RootTypeSpec(spec)
```
sywhang, line 79: Rather than doing this, would it make more sense to change the template to return `(string)(s)`? That way we can still return the string as-is without having to call Sprint

r-hang: Ack, updated. I've added a 3 case check: direct string def, transitive string def, non-string def.

sywhang: I don't think you need to even separate the first and second case, it'll result in the same codegen.

r-hang: Wouldn't separating the 1st and 2nd case avoid a duplicate conversion if the type was already a direct typedef of a string?
`
func (v State) String() string {
	x := (string)(v)
	return (string)(x)
}
`
While preserving correctness here?
`
// String returns a readable string representation of StringReDef.
func (v StringReDef) String() string {
	x := (stringdef.StringDef)(v)
	return (string)(x)
}
`

sywhang: type conversion such as this should be no-op since we're not doing any wonky conversions like string to runes, etc.

That is, the below Go code should compile to the same binary:
```go
type s1 string

func (s s1) String() string {
	x := (string)(s)

	return (string)(x)
}

func (s s1) String2() string {
	x := (string)(s)

	return x
}
```

sywhang (approved):
I'll leave it up to you if you want to optimize for the generated code to look cleaner without the duplicate type conversion for direct string alias types, or if you wanna keep the implementation of codegen cleaner by eliminating the extra if-else clause. 

Approved!

gen/type.go
r-hang: Makes sense, yes i've examined the assembly you referenced and I also ran a benchmark to verify this. I've cut down the code to just an if/else statement like you suggested. Opting to simplify the code generational logic we maintain instead of the look of the generated code itself.

PR #572, opened 2023-05-22
optimise ReadIntXxx() funcs

saurabhagrawal-86:
Using [io.ByteReader](https://pkg.go.dev/io#ByteReader) to decode integers

benchstat results:

```
name         old time/op  new time/op  delta
ReadInt8     10.7ns ±28%   5.2ns ± 2%  -51.11%  (p=0.008 n=5+5)
ReadInt8-2   10.0ns ± 8%   5.2ns ± 2%  -48.22%  (p=0.008 n=5+5)
ReadInt8-4   9.87ns ± 5%  5.18ns ± 1%  -47.50%  (p=0.008 n=5+5)
ReadInt16    14.3ns ± 3%   7.2ns ± 5%  -49.49%  (p=0.008 n=5+5)
ReadInt16-2  14.0ns ± 3%   7.0ns ± 2%  -49.93%  (p=0.008 n=5+5)
ReadInt16-4  14.9ns ± 4%   7.2ns ± 3%  -51.98%  (p=0.008 n=5+5)
ReadInt32    19.5ns ± 3%  11.3ns ± 1%  -42.35%  (p=0.008 n=5+5)
ReadInt32-2  19.8ns ± 5%  11.3ns ± 3%  -43.11%  (p=0.008 n=5+5)
ReadInt32-4  20.3ns ± 5%  11.3ns ± 4%  -44.18%  (p=0.008 n=5+5)
ReadInt64    32.0ns ± 2%  18.6ns ± 2%  -41.76%  (p=0.008 n=5+5)
ReadInt64-2  32.4ns ±10%  18.4ns ± 2%  -43.27%  (p=0.008 n=5+5)
ReadInt64-4  34.9ns ± 6%  18.6ns ± 4%  -46.65%  (p=0.008 n=5+5)
```


saurabhagrawal-86:
I will remove the non-Go files before merging.

protocol/binary/stream_reader_test.go
```diff
@@ -0,0 +1,79 @@
...
+func (bg *byteGenerator) Read(p []byte) (n int, err error) {
+	p[0] = bg.val
+	return 1, nil
+}
+
+// ReadByte is the io.ByteReader implementation
+func (bg *byteGenerator) ReadByte() (byte, error) {
+	return bg.val, nil
+}
+
```
sywhang, line 40: I have a little doubt about this benchmark.

As written, this benchmark is skewing the performance difference using the bytereader.

`Read` and `ReadByte` has different performance characteristics, and is skewing the benchmark results to favor the ByteReader implementation. It's unlikely that a realistic server's profile would reflect the gains noted from the benchmarks here. 

if you took a profile of the benchmark, a big chunk of the claimed performance benefit would be due to the difference in time it spends in Read() vs ReadByte(). In other words, the performance gain you are getting is dependent upon the Read/ReadByte implementation of the underlying interface implementation, which makes this benchmark obsolete.

sywhang (commented):
Thanks for submitting this Saurabh. I left some comments about the benchmark being submitted.

sywhang (commented):
Thanks for submitting this Saurabh. I left some comments about the benchmark being submitted.

protocol/binary/stream_reader_test.go
saurabhagrawal-86: agreed. discarding this change. thanks!

PR #574, opened 2023-05-23
Optimise ReadString() and WriteString()

saurabhagrawal-86:
Optimise StreamReader's ReadString CPU by 35% and memory by 50% using the new `String` and `SliceData` functions introduced in the unsafe package in go 1.20

```
goos: linux
goarch: amd64
pkg: go.uber.org/thriftrw/protocol/binary
cpu: AMD EPYC 7B13
             │ bench_string_dev.out │        bench_string_opt.out         │
             │        sec/op        │   sec/op     vs base                │
ReadString              81.41n ± 2%   53.61n ± 3%  -34.14% (p=0.000 n=10)
ReadString-2            85.67n ± 6%   55.75n ± 6%  -34.92% (p=0.000 n=10)
ReadString-4            87.71n ± 5%   57.14n ± 5%  -34.86% (p=0.000 n=10)
geomean                 84.89n        55.48n       -34.64%

             │ bench_string_dev.out │        bench_string_opt.out        │
             │         B/op         │    B/op     vs base                │
ReadString               96.00 ± 0%   48.00 ± 0%  -50.00% (p=0.000 n=10)
ReadString-2             96.00 ± 0%   48.00 ± 0%  -50.00% (p=0.000 n=10)
ReadString-4             96.00 ± 0%   48.00 ± 0%  -50.00% (p=0.000 n=10)
geomean                  96.00        48.00       -50.00%

             │ bench_string_dev.out │        bench_string_opt.out        │
             │      allocs/op       │ allocs/op   vs base                │
ReadString               2.000 ± 0%   1.000 ± 0%  -50.00% (p=0.000 n=10)
ReadString-2             2.000 ± 0%   1.000 ± 0%  -50.00% (p=0.000 n=10)
ReadString-4             2.000 ± 0%   1.000 ± 0%  -50.00% (p=0.000 n=10)
geomean                  2.000        1.000       -50.00%
```

Similar changes to optimise StreamWriter's WriteString() method.
```
name           old time/op    new time/op    delta
WriteString      21.2ns ± 3%    13.3ns ± 2%  -37.48%  (p=0.000 n=10+9)
WriteString-2    20.8ns ± 2%    13.4ns ± 4%  -35.42%  (p=0.000 n=9+10)
WriteString-4    20.8ns ± 2%    13.4ns ± 8%  -35.86%  (p=0.000 n=9+10)
```


sywhang (approved):
This looks good to me. 

@r-hang PTAL.

protocol/binary/stream_reader_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	encoded := streamBuff.Bytes()
+
+	sr := new(StreamReader)
+	enc := bytes.NewReader(encoded)
+
+	b.ResetTimer()
+
+	for i:= 0; i<b.N; i++ {
+		resetStreamReader(sr, enc)
```
r-hang, line 45: what's the role of resetStreamReader? Would it also work to set the reader field once since the same encoder is being used?

Would managing the io.Seeker best be done here as well?

r-hang (commented):
Logic looks good, I left a question about the structure of the benchmark.

Since this change concerns a go1.20 feature, I think we should merge this after https://github.com/thriftrw/thriftrw-go/pull/575 goes to dev.

protocol/binary/stream_reader_test.go
saurabhagrawal-86: Yes, that would be cleaner. Thanks for the suggestion.

r-hang (approved):
lgtm! I think you'll need to rebase to pull in the contents of #575.

sywhang:
@r-hang i think you pulled in the changes incorrectly to this branch.

r-hang:
@sywhang, the author reached out after their rebase. I'm working with them now.

sywhang:
@r-hang mb, apologies for assuming :p saw your commits show up all of a sudden so I assumed you pushed to the branch. Thanks.

r-hang:
@saurabhagrawal-86 we've updated dev again to resolve some outstanding staticcheck issues. Could you rebase again and see if the remote go1.19 and go1.20 builds pass? Your change built for me locally when I last tested it out once rebased on dev.

protocol/binary/string_post_go120.go
```diff
@@ -0,0 +1,42 @@
...
+// +build go1.20
+
+package binary
+
+import "unsafe"
+
+// ReadString reads a Thrift encoded string.
+func (sr *StreamReader) ReadString() (string, error) {
+	bs, err := sr.ReadBinary()
+	return unsafe.String(unsafe.SliceData(bs), len(bs)), err
```
prashantv, line 31: As the name implies, any use of `unsafe` should be used very carefully. I'd strongly recommend adding a comment that indicates what assumptions are being made and why this is safe.

In this case, `bs` returned from `ReadBinary` has no mutable references (it creates a copy in that method), so it seems safe.

prashantv: An alternate implementation which avoids unsafe by using `strings.Builder` (which internally uses unsafe) would also work, and avoiding separate implementations by version
```go
var b strings.Builder
io.Copy(&b, sr.reader, length)
return b.String()
```

saurabhagrawal-86: Thanks for the suggestion. I've added a comment explaining why "unsafe" is safe to use here.

In your alternate implementation, I'm assuming you meant io.CopyN (not io.Copy). If yes, that ends up being even slower than the original implementation. 

When copying from a bytes.Buffer to a strings.Builder, I understand that io.Copy would avoid an allocation and a copy. However, io.CopyN is unable to make those optimisations and allocates a staging area for the copy operation.  That ends up outweighing the efficiency of the strings.Builder's String() function.

saurabhagrawal-86: benchmark results are here - https://github.com/saurabhagrawal-86/thriftrw-go/pull/1

r-hang: @prashantv, the benchmark results and evaluation of io.CopyN looks fine to me. We'll wait for your follow up!

protocol/binary/stream_reader_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func BenchmarkReadString(b *testing.B) {
+	var streamBuff bytes.Buffer
+
+	// Encode with Streaming protocol
+	w := NewStreamWriter(&streamBuff)
+	require.NoError(b, w.WriteString("the quick brown fox jumps over the lazy dog"), "WriteString")
+	require.NoError(b, w.Close(), "StreamWriter Close")
```
r-hang, line 37: nit: Since the content of the error strings are the names of the methods getting called I think we can drop this.

protocol/binary_test.go
```diff
@@ -560,6 +561,66 @@ func TestBinaryEOFFailure(t *testing.T) {
...
 }
 
+type limitWriter struct {
+	b []byte
+}
+
+var errWriteLimitReached = errors.New("write limit reached")
+
+func (lw *limitWriter) Write(p []byte) (int, error) {
+	fmt.Printf("[limitWriter] write %v bytes\n", len(p))
```
r-hang, line 571: Is there a need for the print statement here? The table test description strings can capture the number of bytes written.

protocol/binary_test.go
```diff
@@ -560,6 +561,66 @@ func TestBinaryEOFFailure(t *testing.T) {
...
+	n := copy(lw.b, p)
+	lw.b = lw.b[n:]
+	if n < len(p) {
+		return n, errWriteLimitReached
+	}
+	return n, nil
```
r-hang, line 577: Since it doesn't seem we're doing anything with the written bytes in the test case, could we just write the bytes to an [io.Discard](https://pkg.go.dev/io#Discard) and set a flag on on writer to direct it to error in the cases that you want it to? Is a limit writer a specific error case you're trying to exercise or will any error on the underlying Write do?

protocol/binary/stream_reader_test.go
saurabhagrawal-86: Sure, I can do that. But could you please help me understand this better?

I assumed that the optional message helps developers quickly understand the test failure without having to browse the test code. Is that correct? 

If we drop the message from the require.NoError() call, it will print the following on test failure:
```
--- FAIL: BenchmarkReadString
    stream_reader_test.go:37: 
                Error Trace:    stream_reader_test.go:37
                Error:          Received unexpected error: <>.
                Test:           BenchmarkReadString
```
It doesn't print the name of the method getting called. The only clue for the developer is the line number where the assertion failed.

protocol/binary_test.go
saurabhagrawal-86: WriteString internally calls the Write method twice sequentially:
1. for int32
2. for []byte
The second write will only be called if the first succeeds. We want to simulate a failure for both. 
Since there is a single `io.Writer`, I couldn't use a simpler failAlwaysWriter. 
A limit writer seemed fitting here as its errors are easy to reason about. It fails if we attempt to write more than what has been allotted. It could also be reused in other tests that require the Write call to fail. The advantage of doing an actual copy() instead of an `io.Discard` is that this implementation can also be used for testing non-error scenarios. I assumed its utility outweighs test performance.
Wdyt?

r-hang: Wanting to test both write cases makes sense, thanks for clarifying!

As for the copy() versus io.Discard, I do understand the potential usability. However, it took me some time to speculate on why live behavior like copy() was used when I didn't see the written bytes used anywhere in the PR -- I actively re-read the code multiple times in search for where the copy would be crucial. I think if I saw something like an io.Discard I would have interpreted the writer as simply testing error cases and would have not spent time on an additional search of the code.

protocol/binary/string_post_go120.go
```diff
@@ -0,0 +1,44 @@
...
+	// mutable references to bs.
+	return unsafe.String(unsafe.SliceData(bs), len(bs)), err
+}
+
+// WriteString encodes a string
+func (sw *StreamWriter) WriteString(s string) error {
+	if err := sw.WriteInt32(int32(len(s))); err != nil {
+		return err
+	}
+
```
r-hang, line 41: Could we also leave a last comment on why this use of "unsafe" is safe?

PR #576, opened 2023-05-30
ci: Test only against go1.20

r-hang:
If a change to ThriftRW utilizes an API only available in the latest version of Go, the CI validation for the previous version will fail. We should stick to using only the latest Go version for CI validation.


sywhang (changes requested):
Rather than do this, can we just use build tags to separate out the file that's relying on just go 1.20?

PR #582, opened 2023-06-09
Revert to strings.Title again for pascalCase

r-hang:
PR #578 replaced strings.Title with golang.org/x/text/cases due to staticcheck SA1019. There is at least one behavioral difference between these two Title casing approaches uncovered during pre-release testing.

A test case to flag this change and an exemption for staticcheck in the relevant file are also added.

Example:
```
chunk := "10x"
fmt.Println(strings.Title(strings.ToLower(chunk))) # 10x
fmt.Println(cases.Title(language.English).String(strings.ToLower(chunk))) # 10X
```


gen/string_test.go
```diff
@@ -109,6 +109,13 @@ func TestConstantName(t *testing.T) {
 			give: "API_VERSION",
 			want: "APIVersion",
 		},
+		{
+			// ensure code generation compatibility. strings.Title and
+			// golang.org/x/text/cases return different results in this
+			// case.
+			give: "10x",
+			want: "10x",
```
sywhang, line 117: might as well make it more fun
```suggestion
			give: "10xDeveloper",
			want: "10xDeveloper",
```

sywhang (approved):

PR #592, opened 2024-02-07
Add: the ability to redact values from Stringer and Error interface.

moisesvega:
Currently, there's no feature to exclude sensitive information from outputs generated by the Stringer and Error interface implementations. However, we use the nolog annotation to exclude specific properties from being logged by the zap.objectMarshal method. This is particularly useful for omitting sensitive data in zap logger outputs. Despite this, there are scenarios where sensitive information might still be exposed. For instance, consider the following example:

```
exception DoesNotExistException {
   /** Key that was missing. */
   1: required string key
   2: optional string Error (go.name="Error2")
   3: optional string userName (go.nolog)
}
```

The code snippet above leads to the following implementation for the MarshalLogObject method:
```golang
// MarshalLogObject allows DoesNotExistException to be logged quickly by zap.
func (v *DoesNotExistException) MarshalLogObject(enc zapcore.ObjectEncoder) error {
   if v == nil {
      return nil
   }
   enc.AddString("key", v.Key)
   if v.Error2 != nil {
      enc.AddString("Error", *v.Error2)
   }
   return nil
}
```
However, our implementation of the Stringer and Error interfaces inadvertently exposes properties marked with nolog:
```golang
func (v *DoesNotExistException) Error() string {
   return v.String()
}

// String provides a human-readable representation of DoesNotExistException.
func (v *DoesNotExistException) String() string {
   if v == nil {
      return "<nil>"
   }

   var fields [3]string
   i := 0
   fields[i] = fmt.Sprintf("Key: %v", v.Key); i++
   if v.Error2 != nil {
      fields[i] = fmt.Sprintf("Error2: %v", *v.Error2); i++
   }
   if v.UserName != nil {
      fields[i] = fmt.Sprintf("UserName: %v", *v.UserName); i++
   }

   return fmt.Sprintf("DoesNotExistException{%s}", strings.Join(fields[:i], ", "))
}
```

This behavior can lead to unintended exposure of information annotated as `nolog`. For example, when this exception is utilized as an Error, or when logged using `zap.Stringer("exception", e)` or `zap.Error(e)` , the nolog-annotated property is inadvertently revealed:

[playground link](https://go.dev/play/p/GSO-9Rco13V?v=)
```golang
func main() {
   name := "Jane Doe"
   e := &DoesNotExistException{
      Key:      "",
      Error2:   nil,
      UserName: &name,
   }
   logger, _ := zap.NewProduction()
   defer logger.Sync() // Ensures buffer is flushed
   logger.Info("hello",
      zap.Object("exception", e),
      zap.Error(e),
   )
     // output: {"level":"info","msg":"hello","exception":{"key":""},"error":"DoesNotExistException{Key: , UserName: Jane Doe"}
}
```
This PR introduces a new feature that enables the redaction of specified properties through a new Go annotation, go.redacted. When applied, this annotation ensures that the actual value of a property is replaced with <redacted> in the outputs generated by our Stringer and Error interface implementations. 

Output after this diff:
```
  // output: {"level":"info","msg":"hello","exception":{"key":""},"error":"DoesNotExistException{Key: , UserName: <redacted>}"}
```


moisesvega (commented):
on behalf of @r-hang

r-hang (commented):
Looks good on behalf of @tchung1118.

We can add the redacted feature designation to zap logs instead if skipping them outright.

CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- No changes yet.
+## Added
+- Redacted annotation provides a mechanism to redact certain struct fields from 
+  logs.
```
r-hang, line 10: errors messages and log objects.

gen/field.go
```diff
@@ -875,3 +900,23 @@ func verifyUniqueFieldLabels(fs compile.FieldGroup) error {
...
 }
+
+// RedactedLabel provides a mechanism to redact certain struct fields from
+// the outputs of String(), Error() and MarshalLogObject() methods.
+//
+//	struct Contact {
+//	    1: required string name
+//	    2: required string email (go.redacted)
+//	}
+const RedactedLabel = "go.redacted"
```
r-hang, line 911: could we change the label to `go.redact` this will be in the verb from like 'log' in the `go.nolog` annotation.

gen/field.go
```diff
@@ -779,18 +792,28 @@ func (f fieldGroupGenerator) Zap(g Generator) error {
 			if <$v> == nil {
 				return nil
 			}
+			< $redactedContent := redactedContent ->
 			<range .Fields>
+				<- $requiresRedaction := requiresRedaction . ->
```
r-hang, line 797: for brevity, I think we should make this variable `shouldRedact`

gen/zap_test.go
```diff
@@ -852,3 +853,26 @@ func TestLogNilStruct(t *testing.T) {
 	require.NoError(t, x.MarshalLogObject(enc))
 	assert.Empty(t, enc.Fields)
 }
+
+func TestZapOptOut(t *testing.T) {
```
r-hang, line 857: Is this a new test? Should we also be adding a the case here for a redacted field since we're adding this new feature here?

moisesvega: Yes, it is. I noticed that no tests have been conducted for ZapOut. Implementing a small test could help increase our code coverage. I updated the test and add a test case.

PR #613, opened 2025-11-18
Relaxed decode of Unions

amolmejari2:
Adds an option `union-decode-relaxed` which enables a flexible union decoding mode that permits unknown values and returns an empty union struct in such cases.

**Context**

Currently, thriftrw-go generated code deserialize unknown Enum values correctly by preserving them and passing them to the client for appropriate handling. However, for Unions, thriftrw-go generated code fails to deserialize unknown cases gracefully, leading to runtime errors and service outages when new fields are introduced on the server side.

This change introduces an opt-in mechanism in thriftrw-go generated code to allow unknown field in Union deserialization. It ensures backward compatibility in evolving schemas, complies with the Robustness Principle, and defers handling decisions to the caller.


anirudhraja (approved):

main.go
```diff
@@ -61,6 +61,8 @@ type genOptions struct {
 	NoZap                 bool   `long:"no-zap" description:"Do not generate code for Zap logging."`
 	OutputFile            string `long:"output-file" value-name:"FILENAME" description:"Generates a single .go file as an output. Specifying an OutputFile prevents code generation for included Thrift Files."`
 	EnumTextMarshalStrict bool   `long:"enum-text-marshal-strict" hidden:"true" description:"Generate code to throw error on trying to marshal unknown enum"`
+	UnionDecodeRelaxed    bool   `long:"union-decode-relaxed" hidden:"true" description:"Setting this flag to true enables a flexible union decoding mode that permits unknown values and returns an empty union struct in such cases."`
```
sywhang: Let's add some more meat to this documentation. 
```suggestion
	UnionDecodeRelaxed    bool   `long:"union-decode-relaxed" hidden:"true" description:"Setting this flag to true enables a flexible union decoding mode that permits unknown values and returns an empty union struct in such cases. The returned empty union cannot be re-encoded."`
```

gen/field.go
```diff
@@ -453,7 +454,7 @@ func (f fieldGroupGenerator) FromWire(g Generator) error {
 				<end>
 			<end>
 
-			<if and .IsUnion (len .Fields)>
+			<if and (not .UnionDecodeRelaxed) .IsUnion (len .Fields)>
```
sywhang, line 463: Let's modify the above template so that when UnionDecodeRelaxed is set, the generated `FromWire` method has a docstring that indicates that it will return an empty union struct when unknown values are encountered.

gen/internal/tests/thrift/union_decode_relaxed.thrift
```diff
@@ -0,0 +1,16 @@
+union Value {
```
sywhang, line 1: can we add a more complex case like multiple unknown values or nested unknown value?

main.go
amolmejari2: done

gen/field.go
amolmejari2: done

gen/internal/tests/thrift/union_decode_relaxed.thrift
amolmejari2: done

sywhang (approved):

sywhang (approved):

PR #619, opened 2026-06-24
Module to Thrift File Generator

Aryan-2509:
**Background**
The `Module` struct currently contains a `Raw` field that stores the original Thrift file contents as a `[]byte`. However, if the `Raw` field becomes inconsistent with the rest of the fields in the `Module` object, there is currently no mechanism to regenerate the Thrift file from the in-memory representation.

A similar issue was encountered by **Bliss-Hydrator** (a platform service at Uber), where the `Module` object was modified programmatically and a reliable way to reconstruct the corresponding Thrift file was required.

**New Functionality**
This change introduces support for generating a Thrift file directly from a `Module` object. The implementation traverses the `Module` representation and writes the corresponding Thrift definition to a specified target path, enabling the Thrift file to be regenerated from the object model itself.

**Additional Changes**
To support Thrift file regeneration, namespace information must be preserved within the `Module` object. As a result, the `Module` struct has been extended to store namespace declarations in addition to the existing metadata.

### DIFF
--- a/compile/generate.go
+++ b/compile/generate.go
@@ -1,0 +1,758 @@
+// Copyright (c) 2026 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package compile
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
+
+	"go.uber.org/thriftrw/ast"
+)
+
+// GenerateThriftFile writes the Thrift IDL representation of this Module to the
+// given path. The content is reconstructed entirely from the Module's compiled
+// structured fields (namespaces, includes, constants, types, services) and does
+// not rely on the raw IDL input.
+//
+// GenerateThriftFile is, in effect, the inverse of Compile: compiling the
+// generated file yields a Module structurally equivalent to the receiver.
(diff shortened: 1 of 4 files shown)


compile/generate.go
```diff
@@ -0,0 +1,758 @@
...
+}
+
+// writeIncludes writes the include statements for the module, reconstructing
+// each include path relative to the directory of the current thrift file.
+func writeIncludes(builder *strings.Builder, module *Module) error {
+	if len(module.Includes) == 0 {
+		return nil
+	}
+
+	moduleDir := filepath.Dir(module.ThriftPath)
```
Atygayev: This computes include paths relative to the original source module, but `GenerateThriftFile(path)` writes the IDL somewhere the caller chooses. If the caller writes to a different directory, the emitted include still points relative to the old directory and recompiling the generated file will look in the wrong place. 

To make a single generated file reliably recompile from any output directory, the include should be computed relative to the generated file’s directory, or the API should document/handle include file generation too.

compile/generate.go
```diff
@@ -0,0 +1,758 @@
...
+	case *StringSpec:
+		return "string", nil
+	case *BinarySpec:
+		return "binary", nil
+	case *ListSpec:
+		typeName, err := getType(t.ValueSpec, module)
+		if err != nil {
+			return "", fmt.Errorf("error while getting type string for list %s : %w", t.ValueSpec.ThriftName(), err)
+		}
+		return fmt.Sprintf("list<%s>", typeName), nil
```
Atygayev: `getType` renders only the bare type syntax, so annotations stored on primitive and container type specs are dropped in typedef targets, constants, service args/returns/exceptions, and container fields. 
For example `typedef list<string> (go.type = "slice") Tags`, with the current emitter, `getType(typedef.Target, module)` renders only the bare type: `typedef list<string> Tags`

Aryan-2509: right after **getType(typedef.Target, module)**; **getAnnotations(typedef.Annotations)** is called which extracts the annotations associated with the typedef and writes them in the same line; Hence this code generation will write entire `typedef list<string> (go.type = "slice") Tags`

compile/generate.go
Aryan-2509: Thanks. Updated the **writeIncludes** logic to calculate the relative paths between the target path and the dependency's path

compile/generate.go
```diff
@@ -0,0 +1,830 @@
...
+		if includedModule != nil && includedModule.Module != nil && typeFilePath == includedModule.Module.ThriftPath {
+			return fmt.Sprintf("%s.%s", includedModule.Name, typeName), nil
+		}
+	}
+
+	return "", fmt.Errorf("unable to resolve qualified type name for type: %s from file: %s", typeName, typeFilePath)
+}
+
+// escapeString escapes special characters in a Thrift string literal.
+func escapeString(s string) string {
```
Atygayev: replace with `strconv.Quote`?

compile/generate.go
```diff
@@ -0,0 +1,830 @@
...
+		if err != nil {
+			return fmt.Errorf("error while resolving absolute path for included module %q: %w", name, err)
+		}
+
+		relPath, err := filepathRel(absOutputDir, absIncluded)
+		if err != nil {
+			return fmt.Errorf("error while computing include path for %q relative to output %q: %w", name, outputPath, err)
+		}
+
+		builder.WriteString(fmt.Sprintf("include \"%s\"\n", relPath))
```
Atygayev: use `filepath.ToSlash(relPath)` to escape the path before writing it

compile/generate.go
Aryan-2509: Thanks for suggestion, replacing with strconv.Quote

compile/generate.go
Aryan-2509: Integrated filepath.ToSlash(relPath)
Thanks!

Atygayev (approved):

Aryan-2509:
Thanks for the review and approval! Could a maintainer please merge this PR ?

PR #620, opened 2026-08-05
Provide goCase as a built-in template function

mark200:
## Summary

Plugins that generate Go code referencing thriftrw-generated types need thriftrw's exact identifier-naming logic: a Thrift field `request_uuid` becomes `RequestUUID` (not `Request_uuid`), `http_request` becomes `HTTPRequest`, and so on. That logic (`goCase` and its helpers) has always been unexported inside the `gen` package, so plugin authors have
had to copy-paste it and hope it never drifts — any divergence produces generated code that fails to compile.

This change makes `goCase` available to all plugins as a built-in template function of `plugin.GoFileFromTemplate`, longside the existing `import` and `formatType` built-ins:

    func (t *<goCase $svc.Name>_<goCase $fn.Name>_Args) MyMethod() { ... }

## Changes

- **New `internal/gocase` package**: the casing implementation (`GoCase`, `PascalCase`, `IsAllCaps`, `CommonInitialisms`) moves out of `gen/string.go` into a leaf package that both `gen` and `plugin` can
  import. `plugin` deliberately does not import `gen` (that would pull the entire generation engine into every plugin binary), so the shared logic needs a home below both — the same structural role `plugin/api`plays for the protocol types.
- **`gen/string.go`**: `goCase`, `pascalCase`, `isAllCaps`, and `commonInitialisms` become thin delegates to `internal/gocase`.
- **`plugin/template.go`**: registers `"goCase": gocase.GoCase` in the
  default template FuncMap. Existing plugins that already register their own `goCase`via `TemplateFunc` are unaffected: user-supplied functions take precedence over built-ins.

## Testing

- Existing `gen` tests (`TestGoCase`, `TestPascalCase`, `TestConstantName`) pass unchanged, verifying the delegation
  preserves behavior.
- New `TestGoFileFromTemplate` case exercises the built-in end to end


laidbacklevi:
NIT: After reviewing manually, I ran the review through AI and it mentioned that there are some unused functions after this change and in certain scenarios manual copy-paste of the code is still needed. If these are actual concerns, please address them in follow-up diffs. I will stamp this diff.

